PR #44, opened 2017-05-06
translate es6 code to es5

therapon:
Translate es6 code to es5. 
Task is a leaf at the moment since I will need to update build logic for NPM and docs which is forthcoming


tgregg:
Looks good

PR #71, opened 2017-07-15
feat: Add 'raw_token' reader option

stanbailes:
To give access to the low level token stream

Ion:
 - Add 'raw_tokens' option to makeTextReader(), allowing the raw text parser
   to be used as a lexer 
 - Export the raw parser's get_ion_type() function to map the raw token type
   to the corrent Ion type.
IonParserTextRaw:
 - Fix bug: get_ion_type incorrectly returned "STRING" for single quoted symbols
 - For raw mode, add "_read_token" as a variant of "_read_value" which returns
   individual tokens.
 - Add accessors for start/end.
 - Add locator() and error_line() functions which wraps the corresponding
   IonSpan implementations.  Leverage them in error reporting.
 - Rationalize start/end handling -- start always points to the first
   character of the raw token, end always points one character after the
   last character of the raw token.
IonSpan:
 - fix spelling of CARRIAGE_RETURN
 - add locator() function to map a position in the span to line/column information
 - add error_line() which returns the line containing a given token followed
   by a line which points to the token
IonText:
 - Add is_single_operator_char() function so that e.g., "a::-1" tokenizes
   as 'a', '::', '-1' instead of 'a', '::-', '1'
IonTextReader:
 - add raw_parser() accessor
tests/unit:
 - add tokenizer test
tests/unit/iontests:
 - Replace backslashes in paths with forward slashes so these tests work on Windows
 - Add a way to run only specific files for debugging
 - Since makeSpan creates a binary span if you give it a Buffer, convert text
   files to strings so makeSpan will create a text span.
tests/unit/tokens:
 - new unit test for raw mode token handling

### DIFF
--- a/src/Ion.ts
+++ b/src/Ion.ts
@@ -39,12 +39,13 @@ const e = {
 export interface Options {
   catalog: Catalog;
   sourceType: string;
+  raw_tokens: boolean;
 }
 
 /**
- * Returns the `buf` type as binary or text. 
- * 
- * @param buffer we want to check its type 
+ * Returns the `buf` type as binary or text.
+ *
+ * @param buffer we want to check its type
  * @returns either `'binary'` or `'text'`
  */
 function get_buf_type(buf: Span) {
@@ -57,7 +58,7 @@ function makeBinaryReader(span: Span, options: Options) : BinaryReader {
 }
 
 function makeTextReader(span: Span, options: Options) : TextReader {
-  return new TextReader(span, options && options.catalog);
+  return new TextReader(span, options && options.catalog, options && options.raw_tokens);
 }
 
 /**
@@ -133,4 +134,4 @@ export { SharedSymbolTable } from "./IonSharedSymbolTable";
 export { Timestamp } from "./IonTimestamp";
 export { toBase64 } from "./IonText";
 export { TypeCodes } from "./IonBinary";
-
+export { get_ion_type } from "./IonParserTextRaw";
--- a/src/Ion.ts
+++ b/src/Ion.ts
@@ -15,7 +15,7 @@ import { BinaryReader } from "./IonBinaryReader";
 import { Catalog } from "./IonCatalog";
 import { IVM } from "./IonConstants";
 import { Reader } from "./IonReader";
-import { Span, makeSpan } from "./IonSpan";
+import { BinarySpan, StringSpan } from "./IonSpan";
 import { TextReader } from "./IonTextReader";
 import { InvalidArgumentError } from "./IonErrors";
 import { Writer } from "./IonWriter";
@@ -48,35 +48,19 @@ export interface Options {
  * @param buffer we want to check its type
  * @returns either `'binary'` or `'text'`
  */
-function get_buf_type(buf: Span) {
-  var firstByte = buf.valueAt(0);
+function get_buf_type(buf: any) {
+  let firstByte = typeof(buf) === 'string'
+    ? buf.charCodeAt(0)
+    : buf[0];
   return (firstByte === IVM.binary[0]) ? 'binary' : 'text';
 }
 
-function makeBinaryReader(span: Span, options: Options) : BinaryReader {
+function makeBinaryReader(span: BinarySpan, options: Options) : BinaryReader {
   return new BinaryReader(span, options && options.catalog);
 }
 
-function makeTextReader(span: Span, options: Options) : TextReader {
+function makeTextReader(span: StringSpan, options: Options) : TextReader {
   return new TextReader(span, options && options.catalog, options && options.raw_tokens);
-}
-
-/**
- * Wrapper method to capture and/or propagate exceptions occuring during span creation.
- *
- * @param buf value passed in that should represent Ion data, typically as a `string`
- * @returns {Span}
- */
-function asSpan(buf: any) : Span {
-  try {
-    return  makeSpan(buf)
-  } catch (e) {
-    if (e instanceof InvalidArgumentError) {
-      throw new Error("Invalid argument, expected \'string\' value found:  " + buf);
-    } else {
-      throw e;
-    }
-  }
 }
 
 /**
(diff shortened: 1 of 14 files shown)


src/Ion.ts
```diff
@@ -133,4 +134,4 @@ export { SharedSymbolTable } from "./IonSharedSymbolTable";
 export { Timestamp } from "./IonTimestamp";
 export { toBase64 } from "./IonText";
 export { TypeCodes } from "./IonBinary";
-
+export { get_ion_type } from "./IonParserTextRaw";
```
therapon: I know the current code base uses the C like naming convention but we would like to move to CamelCase. Can we rename the exported function to `getIonTypes` or something to that effect

src/Ion.ts
```diff
@@ -57,7 +58,7 @@ function makeBinaryReader(span: Span, options: Options) : BinaryReader {
 }
 
 function makeTextReader(span: Span, options: Options) : TextReader {
-  return new TextReader(span, options && options.catalog);
+  return new TextReader(span, options && options.catalog, options && options.raw_tokens);
```
therapon: We might want to consider factoring out a Lexer and Parser for IonJS and provide typescript interfaces for both. 

Then we can have 2 entry points to the lib 

1. `Ion.ts` for using (read/write) Ion data 
1. `Parser.ts` and `Lexer.ts` that can provide a richer API for tools that would like to reuse/rely on the Ion lexer/parser

src/IonSpan.ts
```diff
@@ -51,6 +51,14 @@ export abstract class Span {
 
   protected abstract clone(start: number, len: number): Span;
 
+  locator(pos: number) : number[] {
+    throw new Error("not implemented");
```
therapon: Did you mean for these methods to be `abstract` instead of throwing a not-implemented error?

src/IonSpan.ts
```diff
@@ -162,14 +170,53 @@ class StringSpan extends Span {
...
   }
 
   offset() : number {
     return this._pos - this._line_start;
   }
 
+  // Return the line and column number for a given character in the span
+  locator(pos: number): number[] {
+    var line = 1, column = 1;
+    for (var i = 0; i < pos; i++) {
```
therapon: use `let` rather than `var`

src/IonSpan.ts
```diff
@@ -162,14 +170,53 @@ class StringSpan extends Span {
...
+        column++;
+      }
+    }
+    return [line, column];
+  }
+
+  // Return a string which displays the line containing the given token followed
+  // by a line which places carats under the offending token
+  error_line(start: number, end: number): any[] {
+    var line_start = start;
```
therapon: for local vars use `let` instead

tests/unit/iontests.js
```diff
@@ -51,15 +50,31 @@ define([
       'good/utf32.ion',
     ];
 
+    // For debugging, put single files in this list to have the test run only
+    // that file.   (But don't forget to clean this up on checkin!)
+    var debugList = [
```
therapon: https://theintern.github.io/intern/#common-config

Have not really used it myself but given a naming convention for the tests would the `grep` option suffice here?

src/IonParserTextRaw.ts
```diff
@@ -263,6 +269,18 @@ export class ParserTextRaw {
     helpers[CH_MS] = this._read_value_helper_minus; // '-'
 
     this._read_value_helper_helpers = helpers;
+
+    helpers = {};
```
therapon: I would prefer not to reuse the var by re-initializing to `{}`. 
Is it true that for the case of tokens there is really one extra helper compared to the case for values? 

 I would refactor out a common object and add the variance for each set of helpers.

src/Ion.ts
stanbailes: Done.

src/IonSpan.ts
stanbailes: Well, I really just want these methods implemented on text spans, but for some reason the text parser constructor uses the abstract 'Span' class.  This seems wrong -- is there ever a reason to pass a binary span to a text reader?   If we can change the IonParserTextRaw constructor to take a TextSpan, I can clean this up easily enough....

src/IonSpan.ts
stanbailes: Done.

src/IonSpan.ts
stanbailes: Done

src/IonSpan.ts
therapon: The text reader should be dealing with a text span. That seems like a miss.

tests/unit/iontests.js
stanbailes: Thanks for the 'intern' pointer.  I tried using the grep option, but in this case it pollutes the output with hundreds of "SKIP:" lines.  I'm fine with removing this code if you like -- let me know...

therapon: Ah I see. I am fine with the code.

src/IonSpan.ts
stanbailes: So are you OK with my fixing makeReader so that it creates the correct Span type, changing the text reader and raw parser to accept StringSpan, and moving these methods out of the abstract Span class into StringSpan?

therapon: Yes.

src/IonParserTextRaw.ts
stanbailes: Ok, I've restructured and simplified this whole helper initialization -- it's now factored as you suggest, and there's no more "patching", and there are no more magic ASCII numbers in the code

src/Ion.ts
stanbailes: I  agree with this basic approach, but I'm not prepared to deal with this right now;  I think there will be some issues splitting it based on the way the existing parsing automaton works.

therapon: Agreed.

 As a means of keeping them  separate, lets add a new method `makeTokenReader` (or something like that) so that we can easily split clients calling the code.

The implementation, for now, can stay as is and we'll add an issue/todo for the parser/lexer split.

src/Ion.ts
```diff
@@ -134,4 +117,9 @@ export { SharedSymbolTable } from "./IonSharedSymbolTable";
 export { Timestamp } from "./IonTimestamp";
 export { toBase64 } from "./IonText";
 export { TypeCodes } from "./IonBinary";
-export { get_ion_type } from "./IonParserTextRaw";
+export { getIonType } from "./IonParserTextRaw";
+// Local Variables:
```
therapon: These look IDE related. Please remove

src/IonSpan.ts
```diff
@@ -306,46 +262,14 @@ class BinarySpan extends Span {
...
-  }
-  if (span === undefined) {
-    throw new InvalidArgumentError("invalid span source");
+  write(b: number) : never {
+    throw new Error("not implemented");
   }
-  return span;
 }
+
+// Local Variables:
```
therapon: ditto

therapon:
merged manually

PR #74, opened 2017-09-12
fix(source): Fix skipping past containers in IonTextReader

mtcalvin:
Fix an incorrect check for `undefined` rather than `EOF` to denote the end of a container when
skipping over containers via skip_past_container() (from next()).


mtcalvin:
Thanks!

PR #75, opened 2017-09-12
fix(source): Fix numberValue and booleanValue for TextReader

mtcalvin:
Calling `next()` on a text reader consumes `_value_type` into `_curr` and resets `_value_type` to
`ERROR`.  Changed `numberValue()` and `booleanValue()` to check `this._curr` instead.


mtcalvin:
Relevant parser code: https://github.com/amzn/ion-js/blob/4eac4687aec37eb5848cabeac065b38e8a6653b5/src/IonParserTextRaw.ts#L1349

PR #117, opened 2018-01-21
Add output target for CommonJS/ES5 (better support for module loaders)

### LINKED ISSUE #116: Loading ion-js breaks standard Webpack build
I'm trying to include ion-js in a project that uses a standard webpack build (create-react-app):

```javascript
import { makeReader } from 'ion-js';

console.log('Hello world!');
```

[See the full repo here](https://github.com/cimi/ion-js-webpack), this is [the commit that introduces ion-js](https://github.com/cimi/ion-js-webpack/commit/481f18ceb536fea155cdf4a1dd992bccd82dad46).

Trying to bundle the project fails with the following error:

```
➜  ion-js-webpack-demo git:(master) yarn build
yarn run v1.3.2
$ react-scripts build
Creating an optimized production build...
Failed to compile.

Failed to minify the code from this file:

    ./node_modules/ion-js/dist/commonjs/es6/IonUtilities.js:27

Read more here: http://bit.ly/2tRViJ9

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

This is because the loader does not support ES6 modules by default. While this can be probably worked around by ejecting the default configuration and customising the webpack build configuration, it would be much more convenient if ion-js produced an ES5 version that can be imported directly. The ES5 AMD bundle generated in the current build setup is meant to be included standalone in a browser and does not work with module loaders.

cimi:
It is possible to make the TypeScript compiler to produce ES5 output, even while using generators and other experimental features. [TypeScript 2.3+](ts-release) added the `downlevelIteration` option:

> With --downlevelIteration, the compiler uses new type check and emit behavior that attempts to call a [Symbol.iterator]() method on the iterated object if it is found, and creates a synthetic array iterator over the object if it is not.

I've had to upgrade [`grunt-ts` to the latest version](https://github.com/TypeStrong/grunt-ts/blob/master/CHANGELOG.md#v600-beta17) in order to be able to pass these options to the compiler.

Finally, `package.json` can be updated with a [`browser` field][browser-spec] instructing bundlers like webpack and browserify to use the ES5 code generated by the tyepescript compiler instead of requiring babel and polyfills.

Combining these three features, I've changed the build of the `ion-js` package to produce an ES5 version of its CommonJS output which can be used without any extra configuration. The change should be transparent to all consumers that are not module loaders.

After building ion-js with the proposed changes and linking the package locally the webpack build succeeds:

```
➜  ion-js-webpack-demo git:(master) yarn build
yarn run v1.3.2
$ react-scripts build
Creating an optimized production build...
Compiled with warnings.

./src/index.js
  Line 1:  'makeReader' is defined but never used  no-unused-vars

Search for the keywords to learn more about each warning.
To ignore, add // eslint-disable-next-line to the line before.

File sizes after gzip:

  28.75 KB (-14.11 KB)  build/static/js/main.55cc3a08.js
  [...]
```
Fixes #116 

[ts-release]: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-3.html
[compiler-options]: https://www.typescriptlang.org/docs/handbook/compiler-options.html
[grunt-ts-changelog]: https://github.com/TypeStrong/grunt-ts/blob/master/CHANGELOG.md#v600-beta17
[browser-spec]: https://github.com/defunctzombie/package-browser-field-spec
[cra-guidance]: https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/template/README.md#npm-run-build-fails-to-minify

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -75,7 +75,7 @@ module.exports = function(grunt) {
           sourceMap: true  // generate .map files for debugging
         }
       },
- 
+
       'amd-es6': {
         src: ['src/**/*.ts'],
         outDir: 'dist/amd/es6',
@@ -119,19 +119,19 @@ module.exports = function(grunt) {
         ]
       }
     },
-    babel: { 
-      options: { 
-        sourceMap: true, 
+    babel: {
+      options: {
+        sourceMap: true,
         presets: ['es2015']
       },
-      dist: { 
-        files: [{ 
-          'expand': true, 
+      dist: {
+        files: [{
+          'expand': true,
           cwd: 'dist/amd/es6',
-          'src': ['*.js'], 
-          'dest': 'dist/amd/es5/', 
+          'src': ['*.js'],
+          'dest': 'dist/amd/es5/',
           'ext': '.js',
-        }], 
+        }],
       },
     },
     /**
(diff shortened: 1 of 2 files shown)


cimi:
I'm not planning to update this PR so I'll close it - please feel free to reuse any of the code here.

PR #118, opened 2018-03-01
Comprehensive update to IonJS

wesboyt:
Prior to this update IonJS was failing the basic hello world example in the readme.
I rolled back until the state of the parser was able to run the hello world example and began implementing support for the cross implementation harness within these files:
- src/IonEvent.ts
- src/IonEventStream.ts
- tests/unit/iontests.js

As the testing revealed bugs throughout the codebase files were edited multiple times to fix different issues that were revealed. I will go over each file's overall diff and explain the changes within.
Changes:
 - src/IonTextReader.ts:
Ironed out the interface apis around outputting values to generalize based on the state of the reader.

 - src/IonParserTextRaw.ts:
Fixed improper state changes within the single quote helper function.
Fixed decimal detection.
Fixed infinity  state handling
Fixed float state handling
Rebuilt triple quote string parsing from scratch, still has a bug when in a list with multiple escaped characters but is mostly functional.
Separated the read behavior around detecting triple quotes from the state of the span.
Fixed (null .timestamps) from parsing as a corrupted null value by allowing null to be a value on its own.
Fixed random incorrect clob2 to clob3.
Rebuilt get_value_as_string almost entirely to correctly read through clobs, strings, triple quotes, etc.
Fixed escaped newline handling.
Separated reading through whitespaces from the span so that reading multi string values(triple quotes and clobs) doesn't corrupt the state of the parser.

 - src/IonSpan.ts:
Exported the stringspan class, not sure why it wasn't available before. accidentally keep including the random a in the file.

 - src/IonTextWriter.ts:
The writer used to lose it's state within nested containers like struct list sexp. I rebuilt the writer's context support from scratch.
Fixed Clob writing behavior around code points above 127.

 - src/IonText.ts:
The numeric terminator boolean array was missing '\v'.
CommonEscapes had an error using [] instead of ().

 - src/IonWriteable.ts:
Simplified the api around buffer allocation, its about 2-4 times slower on extremely large inputs but is error free (the prior implementation corrupted on roundtrip when a value was split between 4k chunks).

 - src/IonUnicode.ts:
Almost entirely spacing changes except for the difference in instantiating 'Writeables'.

 - src/IonTimestamp.ts:
Added argument to decimal parse function call, otherwise the same.

 - src/IonDecimal.ts:
Added optional argument to parse function to allow correct timestamp behavior around trimming zeroes.
Fixed 'shift' behavior around negative exponents

### DIFF
--- a/src/Ion.ts
+++ b/src/Ion.ts
@@ -23,6 +23,7 @@ import { TextWriter } from "./IonTextWriter";
 import { Writeable } from "./IonWriteable";
 import { BinaryWriter } from "./IonBinaryWriter";
 import { LocalSymbolTable, defaultLocalSymbolTable } from "./IonLocalSymbolTable";
+import { IonEventStream } from "./IonEventStream";
 
 const e = {
   name: "IonError",
@@ -133,4 +134,5 @@ export { SharedSymbolTable } from "./IonSharedSymbolTable";
 export { Timestamp } from "./IonTimestamp";
 export { toBase64 } from "./IonText";
 export { TypeCodes } from "./IonBinary";
+export { IonEventStream } from "./IonEventStream";
 
--- a/src/IonBinaryReader.ts
+++ b/src/IonBinaryReader.ts
@@ -154,7 +154,7 @@ export class BinaryReader implements Reader {
     return this._parser.hasAnnotations();
   }
 
-  annotations() : string[] {
+  annotations() : string[] {//TODO binary support
       return ["test"];
   }
 
--- a/src/IonBinaryReader.ts
+++ b/src/IonBinaryReader.ts
@@ -154,6 +154,10 @@ export class BinaryReader implements Reader {
     return this._parser.hasAnnotations();
   }
 
+  annotations() : string[] {
+      return ["test"];
+  }
+
   getAnnotation(index: number) : string {
     let t: BinaryReader = this;
     var id, n;
--- a/src/IonDecimal.ts
+++ b/src/IonDecimal.ts
@@ -89,29 +89,31 @@ export class Decimal {
             return "null.decimal";
         }
 
-        let shift: number = this._exponent;
+        let exponent: number = this._exponent;
         let image: string = this._value.digits();
 
-        if (shift <  0) {
+        if (exponent <  0) {
             // negative shift - prefix decimal point this may require leading zero's
-            if (image.length < shift + 1) {
-                for (let i : number = shift + 1 - image.length; i > 0; i--) {
+            if (image.length < exponent + 1) {
+                for (let i : number = exponent + 1 - image.length; i > 0; i--) {
                     image = "0" + image;
                 }
             }
-            let decimal_location: number = image.length + shift;
+            let decimal_location: number = image.length + exponent;
             if (decimal_location <= 0) {
                 image = '0.' + image;
             } else {
                 image = image.substr(0, decimal_location) + "." + image.substr(decimal_location);
             }
-        } else if (shift > 0) {
+        } else if (exponent > 0) {
             // positive shift,
             if (image.length > 1) {
-                shift = shift + image.length - 1;
+                exponent = exponent + image.length - 1;
                 image = image.substr(0, 1) + "." + image.substr(1);
             }
-            image = image + "d" + shift.toString();
+            image = image + "d" + exponent.toString();
+        } else if(exponent === 0) {
+            image = image + ".";
         }
 
 
--- a/src/IonDecimal.ts
+++ b/src/IonDecimal.ts
@@ -132,7 +132,7 @@ export class Decimal {
       return this.getExponent() === expected.getExponent() && this.isNegative() === expected.isNegative() && this.getDigits().numberValue() === expected.getDigits().numberValue();
   }
 
-  static parse(str: string, dontStrip? : boolean) : Decimal {
+  static parse(str: string, stripLeadingZeroes : boolean = true) : Decimal {
     let index: number = 0;
     let exponent: number = 0;
     let c: number;
@@ -152,7 +152,7 @@ export class Decimal {
 
     let digits: string = Decimal.readDigits(str, index);
     index += digits.length;
-    if(!dontStrip) digits = Decimal.stripLeadingZeroes(digits);
+    if(stripLeadingZeroes) digits = Decimal.stripLeadingZeroes(digits);
 
     if (index === str.length) {
       let trimmedDigits: string = Decimal.stripTrailingZeroes(digits);
--- a/src/IonDecimal.ts
+++ b/src/IonDecimal.ts
@@ -92,7 +92,7 @@ export class Decimal {
         let shift: number = this._exponent;
         let image: string = this._value.digits();
 
-        if (shift < 0) {
+        if (shift <  0) {
             // negative shift - prefix decimal point this may require leading zero's
             if (image.length < shift + 1) {
                 for (let i : number = shift + 1 - image.length; i > 0; i--) {
@@ -132,6 +132,10 @@ export class Decimal {
 
   getExponent() : number {
     return this._exponent;
+  }
+
+  equals(expected : Decimal) : boolean {
+      return this.getExponent() === expected.getExponent() && this.isNegative() === expected.isNegative() && this.getDigits().numberValue() === expected.getDigits().numberValue();
   }
 
   static parse(str: string, dontStrip? : boolean) : Decimal {
--- a/src/IonDecimal.ts
+++ b/src/IonDecimal.ts
@@ -90,37 +90,29 @@ export class Decimal {
         }
 
         let exponent: number = this._exponent;
-        let image: string = this._value.digits();
-
-        if (exponent <  0) {
+        let coefficient: string = this._value.digits();
+        let significantDigits : number = coefficient.length;
+        let result : string = '';
+        //digits returns an integer coefficient with an exponent that shifts it back to its original state
+        if (exponent < 0) {
             // negative shift - prefix decimal point this may require leading zero's
-            if (image.length < exponent + 1) {
-                for (let i : number = exponent + 1 - image.length; i > 0; i--) {
-                    image = "0" + image;
-                }
-            }
-            let decimal_location: number = image.length + exponent;
-            if (decimal_location <= 0) {
-                image = '0.' + image;
-            } else {
-                image = image.substr(0, decimal_location) + "." + image.substr(decimal_location);
+            let adjustedExponent : number = significantDigits - 1 + exponent;
+            if(adjustedExponent >= 0){
+                let decimalIndex : number = significantDigits + exponent;
+                result = coefficient.slice(0, decimalIndex) + '.' + coefficient.slice(decimalIndex);
+            }else if(adjustedExponent >= -6){//adapted from http://speleotrove.com/decimal/daconvs.html
+                result = '0.00000'.slice(0, (2 - exponent) - significantDigits) + coefficient;
+            }else{
+                result = coefficient + '.d' + exponent;
             }
         } else if (exponent > 0) {
-            // positive shift,
-            if (image.length > 1) {
-                exponent = exponent + image.length - 1;
-                image = image.substr(0, 1) + "." + image.substr(1);
-            }
-            image = image + "d" + exponent.toString();
+            // positive shift
+            result = coefficient + '.d' + exponent;
         } else if(exponent === 0) {
-            image = image + ".";
+            result = coefficient + ".";
         }
-
-
-        if (this.isNegative()) {
-            image = "-" + image;
-        }
-        return image;
+        if (this.isNegative()) result = '-' + result;
+        return result;
     }
 
   isNull() : boolean {
--- a/src/IonDecimal.ts
+++ b/src/IonDecimal.ts
@@ -84,42 +84,42 @@ export class Decimal {
     return this.stringValue();
   }
 
-  stringValue(): string {
-    if (this.isNull()) {
-      return "null.decimal";
-    }
-
-    let s: number = this._exponent;
-    let image: string = this._value.digits();
-
-    if (s < 0) {
-      // negative shift - prefix decimal point this may require leading zero's
-      if (image.length < s + 1) {
-        for (let i : number = s + 1 - image.length; i > 0; i--) {
-          image = "0" + image;
+    stringValue(): string {
+        if (this.isNull()) {
+            return "null.decimal";
         }
-      }
-      let decimal_location: number = image.length + s;
-      if (decimal_location === 0) {
-        image = '0.' + image;
-      } else {
-        image = image.substr(0, decimal_location) + "." + image.substr(decimal_location);
-      }
-    }
-    else if (s > 0) {
-      // positive shift, 
-      if (image.length > 1) {
-        s = s + image.length - 1;
-        image = image.substr(0, 1) + "." + image.substr(1);
-      }
-      image = image + "d" + s.toString();
-    }
 
-    if (this.isNegative()) {
-      image = "-" + image;
+        let shift: number = this._exponent;
+        let image: string = this._value.digits();
+
+        if (shift < 0) {
+            // negative shift - prefix decimal point this may require leading zero's
+            if (image.length < shift + 1) {
+                for (let i : number = shift + 1 - image.length; i > 0; i--) {
+                    image = "0" + image;
+                }
+            }
+            let decimal_location: number = image.length + shift;
+            if (decimal_location <= 0) {
+                image = '0.' + image;
+            } else {
+                image = image.substr(0, decimal_location) + "." + image.substr(decimal_location);
+            }
+        } else if (shift > 0) {
+            // positive shift,
+            if (image.length > 1) {
+                shift = shift + image.length - 1;
+                image = image.substr(0, 1) + "." + image.substr(1);
+            }
+            image = image + "d" + shift.toString();
+        }
+
+
+        if (this.isNegative()) {
+            image = "-" + image;
+        }
+        return image;
     }
-    return image;
-  }
 
   isNull() : boolean {
     var isnull = (this._value === undefined);
--- a/src/IonDecimal.ts
+++ b/src/IonDecimal.ts
@@ -134,7 +134,7 @@ export class Decimal {
     return this._exponent;
   }
 
-  static parse(str: string) : Decimal {
+  static parse(str: string, dontStrip? : boolean) : Decimal {
     let index: number = 0;
     let exponent: number = 0;
     let c: number;
@@ -154,7 +154,7 @@ export class Decimal {
 
     let digits: string = Decimal.readDigits(str, index);
     index += digits.length;
-    digits = Decimal.stripLeadingZeroes(digits);
+    if(!dontStrip) digits = Decimal.stripLeadingZeroes(digits);
 
     if (index === str.length) {
       let trimmedDigits: string = Decimal.stripTrailingZeroes(digits);
--- a/src/IonEvent.ts
+++ b/src/IonEvent.ts
@@ -67,7 +67,7 @@ export abstract class IonEvent implements IIonEvent {
         writer.writeFieldName('field_name');
         writer.writeString(this.fieldName);
         writer.writeFieldName('annotations');
-        this.writeAnnotations(writer, this.annotations());
+        this.writeAnnotations(writer);
         this.writeValues(writer);
         writer.writeFieldName('imports');
         this.writeImportDescriptor(writer);
@@ -76,11 +76,14 @@ export abstract class IonEvent implements IIonEvent {
         writer.endContainer();
     }
 
-    writeAnnotations(writer : Writer, annotations : string[]) {
-
+    writeAnnotations(writer : Writer) {
+        if(this.annotations === undefined){
+            writer.writeNull(TypeCodes.LIST);
+            return;
+        }
         writer.writeList();
-        for(var i = 0; i < annotations.length; i++){
-            writer.writeSymbol(annotations[i]);
+        for(var i = 0; i < this.annotations.length; i++){
+            writer.writeSymbol(this.annotations[i]);
         }
         writer.endContainer();
     }
@@ -144,7 +147,11 @@ export abstract class IonEvent implements IIonEvent {
         );
     }
 
-    annotationEquals(expectedAnnotations : string[]) : boolean {//TODO
+    annotationEquals(expectedAnnotations : string[]) : boolean {
+        if(this.annotations.length !== expectedAnnotations.length) return false;
+        for(let i = 0; i < this.annotations.length; i++){
+            if(this.annotations[i] !== expectedAnnotations[i]) return false;
+        }
         return true;
     }
 }
--- a/src/IonEvent.ts
+++ b/src/IonEvent.ts
@@ -354,7 +354,7 @@ class IonBlobEvent extends AbstractIonEvent {
     valueEquals(expected : IonBlobEvent) : boolean {
         return this.ionValue === expected.ionValue;
     }
-    writeIonValue(writer : Writer) : void {
+    writeIonValue(writer : Writer) : void {//TODO needs to be backed by number[] https://github.com/amzn/ion-js/issues/127
         let tempBuf = [];
         for (let i = 0; i < this.ionValue.length; i++) {
             tempBuf.push(this.ionValue.charCodeAt(i));
@@ -370,7 +370,7 @@ class IonClobEvent extends AbstractIonEvent {
     valueEquals(expected : IonClobEvent) : boolean {
         return this.ionValue === expected.ionValue;
     }
-    writeIonValue(writer : Writer) : void {
+    writeIonValue(writer : Writer) : void {//TODO needs to be backed by number[] https://github.com/amzn/ion-js/issues/127
         let tempBuf = [];
         for (let i = 0; i < this.ionValue.length; i++) {
             tempBuf.push(this.ionValue.charCodeAt(i));
--- a/src/IonEvent.ts
+++ b/src/IonEvent.ts
@@ -116,7 +116,7 @@ export abstract class IonEvent implements IIonEvent {
     writeTextValue(writer : Writer) : void {
         let tempTextWriter = new TextWriter(new Writeable());
         this.writeIonValue(tempTextWriter);
-        tempTextWriter.close();
+        //tempTextWriter.close();
         const numBuffer = tempTextWriter.getBytes();
         let stringValue : string = "";
         for(let i : number = 0; i < numBuffer.length; i++){
@@ -352,7 +352,11 @@ export class IonBlobEvent extends IonEvent {
         return this.ionValue === expected.ionValue;
     }
     writeIonValue(writer : Writer) : void {
-        writer.writeBlob(this.ionValue);
+        let tempBuf = [];
+        for(let i = 0; i < this.ionValue.length; i++){
+            tempBuf.push(this.ionValue.charCodeAt(i))
+        }
+        writer.writeBlob(tempBuf);
     }
 }
 export class IonClobEvent extends IonEvent {
@@ -363,8 +367,8 @@ export class IonClobEvent extends IonEvent {
         return this.ionValue === expected.ionValue;
     }
     writeIonValue(writer : Writer) : void {
-        var tempBuf = [];
-        for(var i = 0; i < this.ionValue.length; i++){
+        let tempBuf = [];
+        for(let i = 0; i < this.ionValue.length; i++){
             tempBuf.push(this.ionValue.charCodeAt(i))
         }
         writer.writeClob(tempBuf);
@@ -411,7 +415,7 @@ export class IonStructEvent extends IonContainerEvent {//no embed support as of 
         for(let i : number = 0; matchFound && i < actualEvents.length; i++) {
             matchFound = false;
             for(let j : number = 0; !matchFound && j < expectedEvents.length; j++) {
-                if(paired[j] !== true) {
+                if(!paired[j]) {
                     matchFound = actualEvents[i].equals(expectedEvents[j]);
                     paired[j] = matchFound;
                 }
--- a/src/IonEvent.ts
+++ b/src/IonEvent.ts
@@ -357,7 +357,11 @@ export class IonClobEvent extends IonEvent {
         return this.ionValue === expected.ionValue;
     }
     writeIonValue(writer : Writer) : void {
-        writer.writeClob(this.ionValue);
+        var tempBuf = [];
+        for(var i = 0; i < this.ionValue.length; i++){
+            tempBuf.push(this.ionValue.charCodeAt(i))
+        }
+        writer.writeClob(tempBuf);
     }
 }
 /*
--- a/src/IonEvent.ts
+++ b/src/IonEvent.ts
@@ -33,11 +33,11 @@ export enum IonEventType {
 }
 
 export interface IonEvent {
-    readonly eventType : IonEventType;
-    readonly ionType : IonType;
-    readonly fieldName : string;
-    readonly annotations : string[];
-    readonly depth : number;
+    eventType : IonEventType;
+    ionType : IonType;
+    fieldName : string;
+    annotations : string[];
+    depth : number;
     ionValue : any;
     write(writer : Writer) : void;
     equals(expected : IonEvent) : boolean;
@@ -45,11 +45,11 @@ export interface IonEvent {
 }
 
 abstract class AbsIonEvent implements IonEvent {
-    readonly eventType: IonEventType;
-    readonly ionType: IonType;
-    readonly fieldName: string;
-    readonly annotations: string[];
-    readonly depth: number;
+    eventType: IonEventType;
+    ionType: IonType;
+    fieldName: string;
+    annotations: string[];
+    depth: number;
     ionValue: any;
 
 
@@ -69,15 +69,27 @@ abstract class AbsIonEvent implements IonEvent {
         writer.writeStruct();
         writer.writeFieldName('event_type');
         writer.writeSymbol(IonEventType[this.eventType]);
-        writer.writeFieldName('ion_type');
-        writer.writeSymbol(this.ionType.name);
-        writer.writeFieldName('field_name');
-        writer.writeString(this.fieldName);
-        writer.writeFieldName('annotations');
-        this.writeAnnotations(writer);
-        this.writeValues(writer);
-        writer.writeFieldName('imports');
-        this.writeImportDescriptor(writer);
+        if(this.ionType !== null) {
+            writer.writeFieldName('ion_type');
+            writer.writeSymbol(this.ionType.name);
+        }
+        if(this.fieldName !== null && this.fieldName !== undefined) {
+            writer.writeFieldName('field_name');
+            writer.writeString(this.fieldName);
+        }
+        if(this.annotations !== null) {
+            writer.writeFieldName('annotations');
+            this.writeAnnotations(writer);
+        }
+        if(this.eventType === IonEventType.SCALAR){
+            this.writeValues(writer);
+        }
+        /*
+        if(this.imports !== null){
+            writer.writeFieldName('imports');
+            this.writeImportDescriptor(writer);
+        }
+        */
         writer.writeFieldName('depth');
         writer.writeInt(this.depth);
         writer.endContainer();
@@ -115,10 +127,13 @@ abstract class AbsIonEvent implements IonEvent {
     }
 
     writeValues(writer : Writer) : void {
-        writer.writeFieldName('value_text');
-        this.writeTextValue(writer);
-        writer.writeFieldName('value_binary');
-        this.writeBinaryValue(writer);
+        let containerCutoff : number = 11;
+        if(this.eventType === IonEventType.SCALAR) {
+            writer.writeFieldName('value_text');
+            this.writeTextValue(writer);
+            writer.writeFieldName('value_binary');
+            this.writeBinaryValue(writer);
+        }
     }
 
     writeTextValue(writer : Writer) : void {
@@ -155,6 +170,7 @@ abstract class AbsIonEvent implements IonEvent {
     }
 
     annotationEquals(expectedAnnotations : string[]) : boolean {
+        if(this.annotations === expectedAnnotations) return true;
         if(this.annotations.length !== expectedAnnotations.length) return false;
         for(let i = 0; i < this.annotations.length; i++){
             if(this.annotations[i] !== expectedAnnotations[i]) return false;
@@ -217,7 +233,7 @@ export class IonEventFactory {
                 throw new Error("symbol tables unsupported.");
             case IonEventType.CONTAINER_END :
             case IonEventType.STREAM_END :
-                return new IonEndEvent(eventType, ionType, fieldName, annotations, depth);
+                return new IonEndEvent(eventType, null, null, null, depth);
         }
     }
 }
--- a/src/IonEvent.ts
+++ b/src/IonEvent.ts
@@ -313,7 +313,7 @@ class IonSymbolEvent extends AbstractIonEvent {
         return this.ionValue.name === expected.ionValue.name;//will need to change when symboltokens are introduced.
     }
     writeIonValue(writer : Writer) : void{
-        writer.writeSymbol(this.ionValue.toString());//if symboltokens text is unknown we will need
+        writer.writeSymbol(this.ionValue.toString());//if symboltokens text is unknown we will need to write out symboltable
     }
 
 }
@@ -348,7 +348,7 @@ class IonTimestampEvent extends AbstractIonEvent {
 }
 
 class IonBlobEvent extends AbstractIonEvent {
-    constructor(eventType : IonEventType, ionType : IonType, fieldName : string, annotations : string[], depth : number, ionValue : Timestamp){
+    constructor(eventType : IonEventType, ionType : IonType, fieldName : string, annotations : string[], depth : number, ionValue : string){
         super(eventType , ionType, fieldName, annotations, depth, ionValue);
     }
     valueEquals(expected : IonBlobEvent) : boolean {
@@ -364,7 +364,7 @@ class IonBlobEvent extends AbstractIonEvent {
 }
 
 class IonClobEvent extends AbstractIonEvent {
-    constructor(eventType : IonEventType, ionType : IonType, fieldName : string, annotations : string[], depth : number, ionValue : Timestamp){
+    constructor(eventType : IonEventType, ionType : IonType, fieldName : string, annotations : string[], depth : number, ionValue : string){
         super(eventType , ionType, fieldName, annotations, depth, ionValue);
     }
     valueEquals(expected : IonClobEvent) : boolean {
--- a/src/IonEvent.ts
+++ b/src/IonEvent.ts
@@ -22,7 +22,7 @@ import { TextWriter } from "./IonTextWriter";
 import { Writeable } from "./IonWriteable";
 import { BinaryWriter } from "./IonBinaryWriter";
 import { Reader } from "./IonReader";
-import {defaultLocalSymbolTable} from "./IonLocalSymbolTable";
+import { defaultLocalSymbolTable } from "./IonLocalSymbolTable";
 
 export enum IonEventType {
     SCALAR = 0,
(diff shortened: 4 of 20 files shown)


src/IonEvent.ts
```diff
@@ -0,0 +1,473 @@
...
+
+            default : {
+                throw new Error("IonType " + ionType.name + " unexpected.");
+            }
+        }
+    }
+
+    makeContainerEvent(eventType : IonEventType, ionType : IonType, fieldName : string, annotations : string[], depth : number, ionValue : any) : IonContainerEvent {
+        switch(eventType) {
+            case IonEventType.CONTAINER_START : {
```
jonwilsdon: Is there a reason to have this outer `switch` if you are only checking to see if it is `IonEventType.CONTAINER_START`? Why not an if/else?

src/IonEvent.ts
```diff
@@ -0,0 +1,473 @@
...
+    }
+}
+
+export class IonSymbolEvent extends IonEvent {
+    constructor(eventType : IonEventType, ionType : IonType, fieldName : string, annotations : string[], depth : number, ionValue : string){
+        super(eventType , ionType, fieldName, annotations, depth, ionValue);
+    }
+    valueEquals(expected : IonSymbolEvent) : boolean {
+        if(expected.constructor.name !== IonSymbolEvent.name) return false;
+        return this.ionValue.name === expected.ionValue.name
```
jonwilsdon: You are missing a semi-colon at the end of the return.

src/IonEvent.ts
```diff
@@ -0,0 +1,473 @@
...
+    constructor(eventType : IonEventType, ionType : IonType, fieldName : string, annotations : string[], depth : number, ionValue : Timestamp){
+        super(eventType , ionType, fieldName, annotations, depth, ionValue);
+    }
+    valueEquals(expected : IonBlobEvent) : boolean {
+        return this.ionValue === expected.ionValue;
+    }
+    writeIonValue(writer : Writer) : void {
+        let tempBuf = [];
+        for(let i = 0; i < this.ionValue.length; i++){
+            tempBuf.push(this.ionValue.charCodeAt(i))
```
jonwilsdon: Missing semi-colon at the end of this line,

src/IonEvent.ts
```diff
@@ -0,0 +1,473 @@
...
+    constructor(eventType : IonEventType, ionType : IonType, fieldName : string, annotations : string[], depth : number, ionValue : Timestamp){
+        super(eventType , ionType, fieldName, annotations, depth, ionValue);
+    }
+    valueEquals(expected : IonClobEvent) : boolean {
+        return this.ionValue === expected.ionValue;
+    }
+    writeIonValue(writer : Writer) : void {
+        let tempBuf = [];
+        for(let i = 0; i < this.ionValue.length; i++){
+            tempBuf.push(this.ionValue.charCodeAt(i))
```
jonwilsdon: Missing semi-colon here as well.

src/IonEventStream.ts
```diff
@@ -0,0 +1,417 @@
...
+                        case IonTypes.DECIMAL :
+                            writer.writeDecimal(tempEvent.ionValue, tempEvent.annotations);
+                            break;
+                        case IonTypes.FLOAT :
+                            writer.writeFloat32(tempEvent.ionValue, tempEvent.annotations);
+                            break;
+                        case IonTypes.NULL :
+                            writer.writeNull(tempEvent.ionType.bid, tempEvent.annotations);
+                            break;
+                        case IonTypes.TIMESTAMP :
```
jonwilsdon: Nothing is written out for a timestamp?

src/IonEventStream.ts
```diff
@@ -0,0 +1,417 @@
...
+                }
+                case IonEventType.CONTAINER_START : {
+                    if(actualEvent.equals(expectedEvent)){
+                        actualIndex = actualIndex + (<IonContainerEvent>actualEvent).events.length;
+                        expectedIndex = expectedIndex + (<IonContainerEvent>expectedEvent).events.length;
+                    } else {
+                        return false;
+                    }
+                    break;
+                }
```
jonwilsdon: Should there be a `default:` here?

src/IonEventStream.ts
```diff
@@ -0,0 +1,417 @@
...
+            tid = this.reader.next();
+        }
+    }
+
+    private closeContainer(thisContainer : IonContainerEvent, thisContainerIndex : number) {
+        this.eventStream.push(this.eventFactory.makeEndEvent(IonEventType.CONTAINER_END, thisContainer.ionType, undefined, undefined, thisContainer.depth, undefined));
+        thisContainer.setEvents(this.eventStream.slice(thisContainerIndex, this.eventStream.length));
+    }
+
+//(event_type: EventType, ion_type: IonType, field_name: SymbolToken, annotations: list<SymbolToken>, value_text: string, value_binary: list<byte>, imports: list<ImportDescriptor>, depth: int)
```
jonwilsdon: What is this comment for? Some context would be nice if it is worth keeping.

src/IonEventStream.ts
```diff
@@ -0,0 +1,417 @@
...
+            case 'SCALAR' :
+                return this.eventFactory.makeScalarEvent(
+                    currentEvent.get('ion_type'),
+                    currentEvent.get('field_name'),
+                    currentEvent.get('annotations'),
+                    currentEvent.get('depth'),
+                    currentEvent.get('value_text')
+                );
+            case 'SYMBOL_TABLE' :
+                return undefined;
```
jonwilsdon: Should there be a `default` case here?

src/IonEventStream.ts
```diff
@@ -0,0 +1,417 @@
...
+                throw new Error('unexpected type: ' + input);
+            }
+
+        }
+    }
+
+    private parseAnnotations() : string[] {
+        if(this.reader.isNull()) {
+           return undefined;
+        } else {
```
jonwilsdon: You can remove the `else` around this block.

src/IonParserTextRaw.ts
```diff
@@ -520,20 +517,20 @@ export class ParserTextRaw {
...
-      op = this._read_string3;
-    }
-    else {
-      op = this._read_string1;
-    }
-    op.call(this);
-    if (this._test_string_as_annotation(op)) {
-      // this was an annoation, so we need to try again for the actual value
-      this._ops.unshift( calling_op ); 
+    private _read_value_helper_single( ch1, accept_operator_symbols: boolean, calling_op ) {
```
jonwilsdon: It looks like this definition could have these types made explicit fairly easily:
```
private _read_value_helper_single( ch1: number, accept_operator_symbols: boolean, calling_op: ReadValueHelper ) : void {
```

src/IonParserTextRaw.ts
```diff
@@ -757,41 +756,56 @@ export class ParserTextRaw {
...
+              if( ch === EOF) throw new Error('Closing triple quotes not found.');
+            // read single quoted strings until we see the triple quoted terminator
+            // if it's not a triple quote, it's just content
+          }
+          //mark the possible end of the series of triplequotes
+
+          // Set the end of the value,
+          // this._end will reset later if further triple quotes are found after indexing through whitespace,
+          this._end = this._in.position();
+          //Index past the triple quote.
```
jonwilsdon: It is unclear which code the comments in this `for` loop are supposed to apply to.

src/IonParserTextRaw.ts
```diff
@@ -839,6 +853,7 @@ export class ParserTextRaw {
         this._unread(ch);
         break;
       default:
+          console.log(ch);
```
jonwilsdon: Do we need to log this to the console? Or was it for debugging.

src/IonParserTextRaw.ts
```diff
@@ -1063,26 +1128,118 @@ export class ParserTextRaw {
...
+    return indice;
   }
 
+
+    private _skip_triple_quote_gap(entryIndex: number, end: number, acceptComments: boolean) : number {
+        let tempIndex : number = entryIndex + 3;
+        let ch: number = this._in.valueAt(tempIndex);
+        tempIndex = this.indexWhiteSpace(tempIndex, acceptComments);
+        if (tempIndex + 2 <= end && this.verifyTriple(tempIndex)) {
+            return tempIndex + 4;
```
jonwilsdon: Why don't you `tempIndex += 4;` here and then you can get rid if the `return` on this line and the `else` block after this? It would make the code cleaner to read.

src/IonParserTextRaw.ts
```diff
@@ -1063,26 +1128,118 @@ export class ParserTextRaw {
...
+                }
+                return IonText.ESCAPED_NEWLINE;
+            case ESC_x: // = CH_x, //  values['x'] = ESCAPE_HEX; //    any  \xHH  2-digit hexadecimal unicode character equivalent to \ u00HH
+                if (ii + 3 >= end) {
+                    this._error("invalid escape sequence");
+                    return;
+                }
+                ch = this._get_N_hexdigits(ii + 2, ii + 4);
+                this._esc_len = 3;
+                break;
```
jonwilsdon: Is a `default` case needed here?

src/IonTextWriter.ts
```diff
@@ -32,309 +32,347 @@ import { Writer } from "./IonWriter";
...
+            if (this.currentContainer.clean) {
+                this.currentContainer.clean = false;
+            } else {
+                switch (this.currentContainer.containerType) {
+                    case TypeCodes.LIST:
+                        this.writeable.writeByte(CharCodes.COMMA);
+                        break;
+                    case TypeCodes.SEXP:
+                        this.writeable.writeByte(CharCodes.SPACE);
+                        break;
```
jonwilsdon: `default` case needed here?

tests/unit/iontests.js
```diff
@@ -11,126 +11,376 @@
...
+        function goodExhaust(reader) {
+            var tries = 0;
+            for (;;) {
+                // Safety valve
+                tries++;
+                if (tries > 1000) {
+                    throw new Error("Reader spinning forever!");
+                }
+
+                var next = reader.next();
```
jonwilsdon: Should this be a `let`? `var` is going to hoist this to the top of the function.

tests/unit/iontests.js
```diff
@@ -11,126 +11,376 @@
...
+            var tries = 0;
+            try {
+                for (; ;) {
+                    // Safety valve
+                    tries++;
+                    if (tries > 1000) {
+                        throw new Error("Reader spinning forever!");
+                    }
+
+                    var next = reader.next();
```
jonwilsdon: Should this be a `let`? `var` is going to hoist this to the top of the function.

tests/unit/iontests.js
```diff
@@ -11,126 +11,376 @@
...
+            eventStreamSuite[file] = makeEventStreamTest(file);
+        }
+        for (var file of badUnskipped) {
+            badSuite[file] = makeBadTest(file);
+            badEventStreamSuite[file] = makeBadEventStreamTest(file);
+        }
+
+        function squashEscapes(inputString){
+            tempStr = '';
+            for(i = 0; i < inputString.length; i++){
```
jonwilsdon: If this is used, you don't actually declare `i` anywhere.

src/IonParserTextRaw.ts
```diff
@@ -22,7 +22,8 @@ import * as IonText from "./IonText";
 
 import { IonType } from "./IonType";
 import { IonTypes } from "./IonTypes";
-import { Span } from "./IonSpan";
+import { StringSpan, Span } from "./IonSpan";
+import {is_whitespace} from "./IonText";
 
 const EOF = -1;  // EOF is end of container, distinct from undefined which is value has been consumed
 const ERROR           = -2;
```
therapon: formatting looks off

src/IonParserTextRaw.ts
```diff
@@ -757,41 +756,56 @@ export class ParserTextRaw {
...
-      if (ch != CH_SQ) {
-        this._unread(ch);
-        break;
-      }
-      if (this._peek("\'\'") == ERROR) {
-        break;
+      let ch : number;
+      this._unread(this._peek(""));
+      // read sequence of triple quoted strings:
+        //not exiting correctly, need to put a correcting boolean to check for the end. just finds error probably due to the read calls or something.
```
therapon: comments seem misaligned

src/IonParserTextRaw.ts
```diff
@@ -757,41 +756,56 @@ export class ParserTextRaw {
...
+      for(this._start = this._in.position() + 3; this._peek("\'\'\'") !== ERROR; this._in.unread(this._read_after_whitespace(true))) {//need to check if this off by ones on print?
```
therapon: break each argument to `for` on a new line.

src/IonParserTextRaw.ts
```diff
@@ -757,41 +756,56 @@ export class ParserTextRaw {
...
+          this._read();
+          this._read();
+          this._read();
+          //in tripleQuotes
+          //index content of current triple quoted string,
+          //looking for more triple quotes
+          while(this._peek("\'\'\'") === ERROR) {
+              ch = this._read();
+              if( ch === EOF) throw new Error('Closing triple quotes not found.');
```
therapon: either always add a space after `(` and before `)` or not but not both.

src/IonParserTextRaw.ts
```diff
@@ -757,41 +756,56 @@ export class ParserTextRaw {
...
+            // read single quoted strings until we see the triple quoted terminator
+            // if it's not a triple quote, it's just content
+          }
+          //mark the possible end of the series of triplequotes
```
therapon: This looks like a dangling comment. Remove the empty line or join the comments.

src/IonParserTextRaw.ts
```diff
@@ -757,41 +756,56 @@ export class ParserTextRaw {
...
+
+          // Set the end of the value,
+          // this._end will reset later if further triple quotes are found after indexing through whitespace,
+          this._end = this._in.position();
+          //Index past the triple quote.
+
+          this._read();
```
therapon: second instance of 3 identical lines for reading 3 elements. Make a helper that takes the number of `next` 's to call and use that.

src/IonParserTextRaw.ts
```diff
@@ -757,41 +756,56 @@ export class ParserTextRaw {
...
       }
-      this._read(); // consume the 2 pending single quotes
-      this._read();
-    }
-    this._value_push( T_STRING3 );
+      this._value_push( T_STRING3 );
   }
 
+    private verifyTriple(entryIndex : number) : boolean {
+        return this._in.valueAt(entryIndex) === CH_SQ && this._in.valueAt(entryIndex + 1) === CH_SQ && this._in.valueAt(entryIndex + 2) === CH_SQ;
```
therapon: If the line is getting too long, break on the operator.

src/IonParserTextRaw.ts
```diff
@@ -757,41 +756,56 @@ export class ParserTextRaw {
...
+    }
+
+    private findTriple() : boolean {
```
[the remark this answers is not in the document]
therapon: Looks like leftover method since it returns `false`

src/IonParserTextRaw.ts
```diff
@@ -1004,7 +1023,7 @@ export class ParserTextRaw {
   }
 
   get_value_as_string(t: number) : string {
-    var ii, ch, s = "";
+    let indice : number, ch : number, escaped, acceptComments : boolean, s : string = "";
```
therapon: same comment on `indice`

also, more on a personal preference, if the variables declared are of *different* types then use separate declarations.

src/IonParserTextRaw.ts
```diff
@@ -1063,26 +1128,118 @@ export class ParserTextRaw {
...
-      let ch: number = this._in.valueAt(ii);
-      if (IonText.is_whitespace(ch)) {
-        // do nothing
-      }
-      else if (ch == CH_SQ) {
-        ii+= 3;
-        if (ii > end || this._in.valueAt(ii-2) != CH_SQ || this._in.valueAt(ii-1) != CH_SQ) {
-          return ii;
-        }
+  private indexWhiteSpace(indice : number, acceptComments : boolean) : number {
```
therapon: `indice` ... I'll stop now

src/IonParserTextRaw.ts
```diff
@@ -1063,26 +1128,118 @@ export class ParserTextRaw {
...
+                  }
+              }
+          }
       }
-      else {
-        this._error("unexpected character");
+      return indice;
+  }
+
+  private indexToNewLine(indice : number){
```
therapon: Sorry I could not help myself, the function name is `indexToNewLine` and it takes `indice` plural? :)

src/IonParserTextRaw.ts
```diff
@@ -1063,26 +1128,118 @@ export class ParserTextRaw {
...
+
+  private indexToCloseComment(indice : number) : number{
+        while(this._in.valueAt(indice) !== CH_AS && this._in.valueAt(indice + 1) !== CH_FORWARD_SLASH){
+            indice++;
+        }
+    return indice;
   }
 
+
+    private _skip_triple_quote_gap(entryIndex: number, end: number, acceptComments: boolean) : number {
```
therapon: naming convention changed here? Was this an existing method or new method?

src/IonParserTextRaw.ts
```diff
@@ -1139,11 +1296,11 @@ export class ParserTextRaw {
       default:
         this._error("unexpected character after escape slash");
     }
-    return;
+    return ch;
   }
 
   private _get_N_hexdigits(ii: number, end: number) : number {
-    var ch, v = 0;
+      var ch, v = 0;
```
therapon: indentation is off

src/IonParserTextRaw.ts
```diff
@@ -1166,6 +1323,7 @@ export class ParserTextRaw {
   }
 
   next(): number {
+      this._ann = [];
```
therapon: indentation is off

src/IonParserTextRaw.ts
```diff
@@ -1248,33 +1407,31 @@ export class ParserTextRaw {
...
-      ch = this._read();
-      this._unread(ch);
-      return ch;
+    if (expected === undefined || expected.length < 1) {
+      return this._in.valueAt(this._in.position());
     }
     while (ii<expected.length) { 
       ch = this._read();
-      if (ch != expected.charCodeAt(ii)) break;
+       if (ch != expected.charCodeAt(ii)) break;
```
therapon: indentation is off

src/IonUnicode.ts
```diff
@@ -27,82 +27,82 @@ const LOW_SURROGATE_OFFSET = 0xDC00;
...
+        if (pushback !== -1) {
+            codePoint = pushback;
+            pushback = -1;
+        } else {
+            let r: IteratorResult<number> = it.next();
+            if (r.done) {
+                return;
+            } else {
+                codePoint = r.value;
+            }
```
therapon: not sure because I cannot tell from the browser but are we adding tabs here or spaces? We should be consistent.

tests/unit/iontests.js
```diff
@@ -11,126 +11,376 @@
...
+                case '?':
+                    return '\?';
+                case '/':
+                    return '/';
+                default:
+                    return '\\' + char;
+            }
+        }
+
+        //registerSuite(goodSuite);
```
therapon: are these supposed to be commented out?

src/Ion.ts
```diff
@@ -23,6 +23,7 @@ import { TextWriter } from "./IonTextWriter";
 import { Writeable } from "./IonWriteable";
 import { BinaryWriter } from "./IonBinaryWriter";
 import { LocalSymbolTable, defaultLocalSymbolTable } from "./IonLocalSymbolTable";
+import { IonEventStream } from "./IonEventStream";
```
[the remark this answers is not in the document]
wesboyt: The eventstream is the only way afaik of connecting the reader to the writer up to this point, the library lacks datagrams or other related apis. I think its worth exploring adding the roundtripping functionality of the event stream by itself but it doesn't exist yet.

src/IonBinaryReader.ts
```diff
@@ -154,6 +154,10 @@ export class BinaryReader implements Reader {
     return this._parser.hasAnnotations();
   }
 
+  annotations() : string[] {
+      return ["test"];
+  }
```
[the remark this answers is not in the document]
wesboyt: its just a no op to stop the interface from puking on compile.

src/IonDecimal.ts
```diff
@@ -84,42 +84,42 @@ export class Decimal {
...
-      // negative shift - prefix decimal point this may require leading zero's
-      if (image.length < s + 1) {
-        for (let i : number = s + 1 - image.length; i > 0; i--) {
-          image = "0" + image;
+        let shift: number = this._exponent;
+        let image: string = this._value.digits();
+
+        if (shift < 0) {
+            // negative shift - prefix decimal point this may require leading zero's
+            if (image.length < shift + 1) {
```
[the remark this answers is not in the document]
wesboyt: I'm not sure refactoring is in the scope of the project at this point, I didn't make these functions.

src/IonEvent.ts
```diff
@@ -0,0 +1,473 @@
...
+    STREAM_END = 4
+}
+
+export interface IIonEvent {
+    write(writer : Writer) : void;
+    equals(expected : IonEvent) : boolean;
+    writeIonValue(writer : Writer) : void;
+}
+
+export abstract class IonEvent implements IIonEvent {
```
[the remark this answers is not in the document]
wesboyt: it allows general typing the ionevents within the eventstream logic even though its a mixture of types.

src/IonEvent.ts
```diff
@@ -0,0 +1,473 @@
...
+        super(eventType , ionType, fieldName, annotations, depth, ionValue);
+    }
+
+    abstract valueEquals(expected : IonContainerEvent);
+
+    setEvents(inputEvents : IonEvent[]) {
+        this.events = inputEvents;
+    }
+
+    writeIonValue(writer : Writer) : void {
```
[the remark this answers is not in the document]
wesboyt: its just part of the event stream spec.

src/IonEvent.ts
```diff
@@ -0,0 +1,473 @@
...
+        super(eventType , ionType, fieldName, annotations, depth, ionValue);
+
+    }
+    //two way equivalence between the structEvents, they must have the exact same number of equivalent elements.
+    valueEquals(expected : IonStructEvent) : boolean {
+        if(expected.constructor.name !== IonStructEvent.name) return false;
+        return this.structsEqual(this.events, expected.events) && this.structsEqual(expected.events, this.events);
+    }
+
+    //for each actual ionEvent, searches for an equivalent expected ionEvent,
```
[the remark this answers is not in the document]
wesboyt: basically structs equal tells you that for each element in my (not set) there is an equal element in the other (not set), you need this to be a two way street since one (not set) could have two equal elements and the other could have only one.

src/IonEvent.ts
```diff
@@ -0,0 +1,473 @@
...
+        for(let i : number = 0; i < this.events.length; i++){
+            if(!this.events[i].equals(expected.events[i])){
+                return false;
+            }
+        }
+        return true;
+    }
+}
+
+export class IonSexpEvent extends IonContainerEvent {
```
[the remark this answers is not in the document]
wesboyt: I messed up the type within value equals but basically they need to return false when comparing list to a sexp. So the code is bugged right now but a refactor to combine them would be incorrect.

src/IonEventStream.ts
```diff
@@ -0,0 +1,417 @@
...
+        writer.writeSymbol("ion_event_stream");
+        for(let i : number = 0; i < this.eventStream.length; i++) {
+            this.eventStream[i].write(writer);
+        }
+    }
+
+    writeIon(writer : Writer) {
+        let tempBuf : number[];
+        let tempEvent : IonEvent;
+        for(let indice : number = 0; indice < this.eventStream.length; indice++){
```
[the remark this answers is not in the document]
wesboyt: indice is not plural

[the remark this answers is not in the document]
wesboyt: Indice is equivalent to index in most dictionaries (english sucks). My first CS prof delineated between the position and the value at said position since index/indice ambiguously mean both.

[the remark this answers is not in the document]
wesboyt: I think its a math thing

src/IonEventStream.ts
```diff
@@ -0,0 +1,417 @@
...
+                default :
+                    throw new Error('Unexpected event field: ' + fieldName);
+            }
+        }
+
+        switch(currentEvent.get('event_type')) {
+            case 'CONTAINER_START' :
+                return this.eventFactory.makeContainerEvent(
+                    IonEventType.CONTAINER_START,
+                    currentEvent.get('ion_type'),
```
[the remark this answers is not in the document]
wesboyt: they actually dont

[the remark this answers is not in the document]
wesboyt: Nvm they totally do, looked at the wrong section.

src/IonTextWriter.ts
wesboyt: it should be a no op

src/IonUnicode.ts
wesboyt: I've been trying not to do a ton of indent manipulation to avoid polluting the pr.

src/IonEventStream.ts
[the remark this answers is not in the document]
tgregg: This keeps distracting people. Pragmatically, it might be advisable to switch to "index" if for no other reason than to not have to explain it during every review.

src/IonParserTextRaw.ts
wesboyt: index is also a verb

src/IonTextWriter.ts
```diff
@@ -32,309 +32,347 @@ import { Writer } from "./IonWriter";
...
-        switch (this.currentContainer) {
-          case TypeCodes.LIST:
-            this.writeable.writeByte(CharCodes.COMMA);
-            break;
-          case TypeCodes.SEXP:
-            this.writeable.writeByte(CharCodes.SPACE);
-            break;
+    private writeValue<T>(typeCode: TypeCodes, value: T, annotations: string[], serialize: Serializer<T>) {
+        if (this.currentContainer.state === State.STRUCT_FIELD) throw new Error("Expecting a struct field");
+        if (isNullOrUndefined(value)) {//should this still call write null for symbols? they should evaluate as undefined?
```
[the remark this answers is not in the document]
wesboyt: This whole library masks bugs by relying on undefined behavior/return values. Eventually the whole thing should fail early when it sees undefined at any point.

src/IonEventStream.ts
```diff
@@ -0,0 +1,410 @@
...
+    }
+
+    getEvents() : IonEvent[]{
+        return this.eventStream;
+    }
+
+    equals(expected : IonEventStream) {
+        let actualIndex : number = 0;
+        let expectedIndex : number = 0;
+        for(; actualIndex < this.eventStream.length && expectedIndex < expected.eventStream.length; actualIndex++, expectedIndex++) {
```
tgregg: What if the streams differ in length? Remember, streams of different lengths may still be equivalent (in the case when they have a different number of SYMBOL_TABLE events, since SYMBOL_TABLE events aren't considered for equality).

src/IonEventStream.ts
```diff
@@ -0,0 +1,410 @@
...
+                case 'annotations' : {
+                    currentEvent.set(fieldName, this.parseAnnotations());
+                    break;
+                }
+
+                case 'value_text' : {
+                    let tempString : string = this.reader.stringValue();
+                    let tempReader : Reader = makeReader(tempString, undefined);
+                    tempReader.next();
+                    currentEvent.set('isNull', tempReader.isNull());
```
tgregg: What's this?

src/IonEventStream.ts
```diff
@@ -0,0 +1,410 @@
...
+                throw new Error('Symbol tables unsupported');
+        }
+        return this.eventFactory.makeEvent(
+            eventType,
+            currentEvent.get('ion_type'),
+            currentEvent.get('field_name'),
+            currentEvent.get('depth'),
+            currentEvent.get('annotations'),
+            currentEvent.get('isNull'),
+            currentEvent.get('value_text')
```
tgregg: Need to verify that value_text and value_binary are data model equivalent. Otherwise, it's an invalid event and an error should be raised.

src/IonEventStream.ts
```diff
@@ -0,0 +1,410 @@
...
+    }
+
+    private parseAnnotations() : string[] {
+        let annotations : string[] = [];
+        if(this.reader.isNull()) {
+           return annotations;
+        } else {
+            this.reader.stepIn();
+            for(let tid = this.reader.next(); tid !== undefined; tid = this.reader.next()) {
+                annotations.push(this.reader.value());
```
tgregg: Needs to be converted to reading a list of SymbolTokens.

src/IonEvent.ts
```diff
@@ -0,0 +1,469 @@
...
+    CONTAINER_START = 1,
+    CONTAINER_END = 2,
+    SYMBOL_TABLE = 3,
+    STREAM_END = 4
+}
+
+export interface IonEvent {
+    eventType : IonEventType;
+    ionType : IonType;
+    fieldName : string;
```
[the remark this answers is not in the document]
wesboyt: They will be once everything within text works, right now the library has no concept of symboltokens, it was written when they were equivalent with strings.

src/IonEventStream.ts
```diff
@@ -0,0 +1,410 @@
...
+                        case IonTypes.SEXP :
+                            writer.writeSexp(tempEvent.annotations, false);
+                            break;
+                    }
+
+                case IonEventType.CONTAINER_END:
+                    writer.endContainer();
+                    break;
+                case IonEventType.STREAM_END:
+                    writer.close();
```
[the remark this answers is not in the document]
wesboyt: It seems like close doesn't necessarily do anything IO wise since its just writing to a buffer(no hooks into the OS)
It simply calls close container so its effectively a no op if the event stream input is valid ion.

    close() : void {
        while (!this.isTopLevel) {
            this.endContainer();
        }
    }

src/IonEventStream.ts
wesboyt: Binary was blowing everything up so i killed it until work starts on that side.

src/IonEventStream.ts
wesboyt: how the reader conveys if a value is null, you have to explicitly check for null and the type.

src/IonEventStream.ts
wesboyt: nvm gonna while loop

src/IonTimestamp.ts
```diff
@@ -348,6 +348,27 @@ export class Timestamp {
     return n;
   }
 
+  equals(expected : Timestamp) : boolean {
+    return this.getPrecision() === expected.getPrecision() && this.offset === expected.offset && this.instantEquals(expected);
+  }
+  instantEquals(expected : Timestamp) : boolean {
```
tgregg: This doesn't take into account timestamps with different local offsets that represent the same instant. For example,

`2007-01-02T01:00+01:00` and
`2007-01-01T23:00-01:00`

It also won't ignore decimal precision, e.g.

`2007-01-01T00:00:00.00-00:00 ` and
`2007-01-01T00:00:00.0000000000-00:00`

because it will use IonDecimal's (data model) equals implementation.

src/IonTimestamp.ts
```diff
@@ -356,8 +377,11 @@ export class Timestamp {
       default: throw { msg: "invalid value for timestamp precision", where: "IonValueSupport.timestamp.toString" };
       case Precision.NULL:
         return "null.timestamp";
-      case Precision.SECONDS:
+        case Precision.SECONDS:
```
tgregg: Superfluous whitespace.

tests/unit/iontests.js
```diff
@@ -11,126 +11,365 @@
...
-    };
-
-    function exhaust(reader) {
-      var tries = 0;
-      for (;;) {
-        // Safety valve
-        tries++;
-        if (tries > 1000) {
-          throw new Error("Reader spinning forever!");
+        let badUnskipped = [];
```
tgregg: Looks like a good opportunity to DRY this up. Trivially different than the `good` case above.

tests/unit/iontests.js
```diff
@@ -11,126 +11,365 @@
...
+        let badEventStreamSuite = {
+            name: 'Bad EventStream tests'
+        };
+
+        function goodExhaust(reader) {
+            let tries = 0;
+            for (;;) {
+                tries++;
+                if (tries > 1000) {
+                    throw new Error("Reader spinning forever!");
```
tgregg: How do you know it won't halt??!

In all seriousness, I'd remove this. If the reader gets caught in an infinite loop, that's a pretty urgent issue that should be addressed as it is found.

tests/unit/iontests.js
```diff
@@ -11,126 +11,365 @@
...
-            //console.log("Stepping out");
-            reader.stepOut();
-          } else {
-            // End of data
-            break;
-          }
-        } else if (next.container && !reader.isNull()) {
-          //console.log("Stepping in");
-          reader.stepIn();
+        function badExhaust(reader) {
```
tgregg: Another opportunity for a common method between good and bad.

tests/unit/iontests.js
```diff
@@ -11,126 +11,365 @@
...
+                    let input = fs.readFileSync(path, options);
+                    goodExhaust(ion.makeReader(input));
+                    resolve();
+                };
+
+                return new Promise(executor);
+            }
+        }
+
+        function makeBadTest(path) {
```
tgregg: And another. And for all the other `make*Test` functions.

tests/unit/iontests.js
```diff
@@ -11,126 +11,365 @@
...
-            } catch (e) {
-              reject(e);
-            }
-          });
-
-          stream.on('data', function(chunk) {
-            chunks.push(chunk);
-          });
-        };
+                let eventStream = new ion.IonEventStream(reader);
```
tgregg: All these files should fail on read, which happens here. There's nothing to re-write.

src/IonEventStream.ts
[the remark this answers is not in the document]
tgregg: I made a comment about this `close` implementation in my other review too. I'll also note here that (once you add support for symbol tables and the IVM), it will be expected that close/finish resets the local symbol table context. Something to keep in mind.

src/IonEventStream.ts
tgregg: Can that be conveyed by the absence of a value?

src/IonEventStream.ts
tgregg: Sure. Might consider adding a placeholder for the validation logic, because it will be important to add once there's binary support.

src/IonEvent.ts
```diff
@@ -0,0 +1,469 @@
...
+    constructor(eventType : IonEventType, ionType : IonType, fieldName : string, annotations : string[], depth : number, ionValue : Timestamp){
+        super(eventType , ionType, fieldName, annotations, depth, ionValue);
+    }
+    valueEquals(expected : IonBlobEvent) : boolean {
+        return this.ionValue === expected.ionValue;
+    }
+    writeIonValue(writer : Writer) : void {
+        let tempBuf = [];
+        for (let i = 0; i < this.ionValue.length; i++) {
+            tempBuf.push(this.ionValue.charCodeAt(i));
```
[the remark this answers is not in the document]
wesboyt: the apis basically take the input characters read them back to the reader as a string and pass it back up as stringv alue. unfortunately its really just an array of char codes.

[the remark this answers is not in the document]
tgregg: I understand that's how it works now, but is that broken? Can we pass any byte through a blob? e.g. 0x00?

src/IonEvent.ts
```diff
@@ -0,0 +1,469 @@
...
+}
+
+class IonSexpEvent extends AbsIonContainerEvent {
+    constructor(eventType : IonEventType, ionType : IonType, fieldName : string, annotations : string[], depth : number){
+        super(eventType , ionType, fieldName, annotations, depth);
+
+    }
+    valueEquals(expected : IonSexpEvent) : boolean {
+        for(let i : number = 0; i < this.ionValue.length; i++){
+            if(!this.ionValue[i].equals(expected.ionValue[i])){
```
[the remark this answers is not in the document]
wesboyt: seems like a good candidate for refactor

src/IonEventStream.ts
tgregg: Please give an example.

tests/unit/iontests.js
wesboyt: I'm not familiar with that term

src/IonEvent.ts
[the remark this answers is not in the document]
tgregg: An IonBlobEvent's `ionValue` should be a number array, right? (Event though it's currently erroneously marked as a Timestamp.) `IonWriter.writeBlob` accepts a number array. So why isn't this simply `writer.writeBlob(this.ionValue)`? What is the extra charCode conversion for? That's what I'm getting at.

src/IonEventStream.ts
wesboyt: read fails and it shows up as a null value

tests/unit/iontests.js
tgregg: Don't Repeat Yourself

src/IonEventStream.ts
tgregg: But if read fails, the test should already have failed before reaching this point.

wesboyt: yes, should have being the key word

tgregg: Let me rephrase that, then. The test *will* have already failed before reaching this point. If it doesn't do that, then the test harness isn't implemented correctly. The test harness doesn't test itself.

wesboyt: so reader.stringvalue returns undefined when parsing a misprinted null value but I print out the correct null value because I interpreted an undefined value as null.

tgregg: The reader should fail if it parses an invalid null value, like `null.foobar`. If the reader returning undefined is synonymous with failure, then fail as soon as a reader returns undefined. There's no opportunity to incorrectly re-write a valid null, because the test has already failed.

wesboyt: thats the goal

src/IonEvent.ts
```diff
@@ -348,7 +348,7 @@ class IonTimestampEvent extends AbstractIonEvent {
 }
 
 class IonBlobEvent extends AbstractIonEvent {
-    constructor(eventType : IonEventType, ionType : IonType, fieldName : string, annotations : string[], depth : number, ionValue : Timestamp){
+    constructor(eventType : IonEventType, ionType : IonType, fieldName : string, annotations : string[], depth : number, ionValue : string){
```
tgregg: Shouldn't the value be a `number[]`, not a text string?

src/IonEvent.ts
```diff
@@ -364,7 +364,7 @@ class IonBlobEvent extends AbstractIonEvent {
 }
 
 class IonClobEvent extends AbstractIonEvent {
-    constructor(eventType : IonEventType, ionType : IonType, fieldName : string, annotations : string[], depth : number, ionValue : Timestamp){
+    constructor(eventType : IonEventType, ionType : IonType, fieldName : string, annotations : string[], depth : number, ionValue : string){
```
tgregg: Also a `number[]`?

src/IonDecimal.ts
```diff
@@ -84,42 +84,36 @@ export class Decimal {
...
+        if (exponent < 0) {
+            // negative shift - prefix decimal point this may require leading zero's
+            let adjustedExponent : number = significantDigits - 1 + exponent;
+            if(adjustedExponent >= 0){
+                let decimalIndex : number = significantDigits + exponent;
+                result = coefficient.slice(0, decimalIndex) + '.' + coefficient.slice(decimalIndex);
+            }else if(adjustedExponent >= -6){//adapted from http://speleotrove.com/decimal/daconvs.html
+                result = '0.00000'.slice(0, (2 - exponent) - significantDigits) + coefficient;
+            }else{
+                result = coefficient + '.d' + exponent;
```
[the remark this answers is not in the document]
wesboyt: noted, breaks testing to change. Theres a bug around losing precision when the . is removed

src/IonDecimal.ts
```diff
@@ -84,42 +84,36 @@ export class Decimal {
...
+            }
+        } else if (exponent > 0) {
+            // positive shift
+            result = coefficient + '.d' + exponent;
```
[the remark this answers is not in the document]
wesboyt: ***

src/IonEvent.ts
```diff
@@ -0,0 +1,469 @@
...
+        writer.writeStruct();
+        writer.writeFieldName('event_type');
+        writer.writeSymbol(IonEventType[this.eventType]);
+        if(this.ionType !== null) {
+            writer.writeFieldName('ion_type');
+            writer.writeSymbol(this.ionType.name);
+        }
+        if(this.fieldName !== null && this.fieldName !== undefined) {
+            writer.writeFieldName('field_name');
+            writer.writeString(this.fieldName);
```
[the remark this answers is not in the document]
wesboyt: Valid, symboltoken support has been added to the issue list.

src/IonEvent.ts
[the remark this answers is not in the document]
wesboyt: noted

src/IonEventStream.ts
```diff
@@ -0,0 +1,410 @@
...
+
+    writeIon(writer : Writer) {
+        let tempBuf : number[];
+        let tempEvent : IonEvent;
+        for(let indice : number = 0; indice < this.eventStream.length; indice++){
+            tempEvent = this.eventStream[indice];
+            if(tempEvent.fieldName !== undefined) {
+                writer.writeFieldName(tempEvent.fieldName);
+            }
+            switch(tempEvent.eventType){
```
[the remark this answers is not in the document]
wesboyt: I agree seems that the logic should be tied to the event classes themselves and refactored out of the event stream.

src/IonEventStream.ts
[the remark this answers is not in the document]
wesboyt: noted https://github.com/amzn/ion-js/issues/129

src/IonEventStream.ts
wesboyt: https://github.com/amzn/ion-js/issues/131

src/IonEventStream.ts
wesboyt: noted

src/IonReader.ts
```diff
@@ -33,4 +33,5 @@ export interface Reader {
   stringValue: () => string;
   timestampValue: () => Timestamp;
   value: () => any;
+  annotations: () => string[];
```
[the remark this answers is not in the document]
wesboyt: noted

src/IonTextReader.ts
```diff
@@ -212,17 +212,48 @@ export class TextReader implements Reader {
...
+                return this.stringValue();
+            }
+            case IonTypes.TIMESTAMP : {
+                return this.timestampValue();
+            }
+            case IonTypes.CLOB : {
+                return this.stringValue();
+            }
+            case IonTypes.BLOB : {
+                return this.stringValue();
```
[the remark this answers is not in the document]
wesboyt: https://github.com/amzn/ion-js/issues/127

src/IonTextWriter.ts
```diff
@@ -32,309 +32,355 @@ import { Writer } from "./IonWriter";
...
-        let c : number = value[i];
-        if (c >= 128) {
-          throw new Error(`Illegal clob character ${c} at index ${i}`);
+    getBytes(): number[] {
+        return this.writeable.getBytes();
+    }
+    private containerContext : Context[] = [];
+
+    constructor(private readonly writeable: Writeable) {
+        this.containerContext.push(new Context(undefined));
```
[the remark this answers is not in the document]
wesboyt: it could be replaced by an if empty() check where clean state is part of the logic flow. Seemed simpler to just initialize it to clean.

src/IonTextWriter.ts
```diff
@@ -32,309 +32,355 @@ import { Writer } from "./IonWriter";
...
+
+        this.writeSymbolToken(fieldName);
+        this.writeable.writeByte(CharCodes.COLON);
+
+        this.currentContainer.state = State.VALUE;
     }
-    if (this.state !== State.STRUCT_FIELD) {
-      throw new Error("Expecting a struct value");
+
+    writeFloat32(value: number, annotations?: string[]) : void {
```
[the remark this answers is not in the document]
wesboyt: Within text they do the same thing, I believe its there because the way inheritance is implemented.

src/IonTimestamp.ts
wesboyt: https://github.com/amzn/ion-js/issues/132

src/IonTextWriter.ts
```diff
@@ -32,309 +32,355 @@ import { Writer } from "./IonWriter";
...
-    this.writeValue(TypeCodes.DECIMAL, value, annotations, (value: Decimal) => {
-      this.writeUtf8(value.toString());
-    });
-  }
-
-  writeFieldName(fieldName: string) : void {
-    if (this.isTopLevel || this.currentContainer !== TypeCodes.STRUCT) {
-      throw new Error("Cannot write field name outside of a struct");
+
+        if (!this.currentContainer.clean) {
```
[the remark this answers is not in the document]
wesboyt: noted

src/IonDecimal.ts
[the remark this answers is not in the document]
tgregg: Is there an issue for this?

src/IonTextWriter.ts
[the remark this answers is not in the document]
tgregg: If they do the same thing, they shouldn't be duplicated.

PR #151, opened 2018-06-04
fix(IonTextReader): Fix a bug which prevented skip_past_container() to function correctly

mikemelchi:
The skip_past_container() method was previously looping until an undefined value was found. This was
incorrectly causing the method to return undefined, which would break the reader. When checking for
EOF, it can correctly identify the end of a container.

### DIFF
--- a/src/IonTextReader.ts
+++ b/src/IonTextReader.ts
@@ -78,7 +78,7 @@ export class TextReader implements Reader {
         p = this._parser;
     while (d > 0) {
       type = p.next();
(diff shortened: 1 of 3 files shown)


mikemelchi:
I see that I committed the ion-tests file. I probably shouldn't have done that. I can update after somebody has taken a look.

mikemelchi:
Also, not too sure why the travis CI build failed...Is it a known issue or related to my change?

mikemelchi:
Discarding in favor of a new, working PR

PR #152, opened 2018-06-05
fix(IonTextReader): Fix a bug which prevented skip_past_container() t…

### LINKED ISSUE #140: TextReader.next() does not skip over containers.
With this input:
```javascript
{ hello: "Ion" }
{ ion: "Hello" }
```
calling `ionReader.next()` twice gives the following (in the Chrome console):
```
ionReader.next();
IonType {bid: 13, name: "struct", scalar: false, lob: false, num: false, …}
ionReader.next();
undefined
```
and it should take you to the second struct instead, skipping over the contents of the first struct:
```
ionReader.next();
IonType {bid: 13, name: "struct", scalar: false, lob: false, num: false, …}
ionReader.next();
IonType {bid: 13, name: "struct", scalar: false, lob: false, num: false, …}
```

mikemelchi:
…o function correctly

The skip_past_container() method was previously looping until an
undefined value was found. This was
incorrectly causing the method to return undefined, which would break
the reader. When checking for
EOF, it can correctly identify the end of a container.

fix #140

### DIFF
--- a/src/IonTextReader.ts
+++ b/src/IonTextReader.ts
@@ -73,16 +73,15 @@ export class TextReader implements Reader {
   }
 
   skip_past_container() {
-    var type, 
-        d = 1,  // we want to have read the EOC tha matches the container we just saw
-        p = this._parser;
-    while (d > 0) {
-      type = p.next();
+    let type;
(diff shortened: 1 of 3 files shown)


wesboyt:
looks good! nice job.

PR #153, opened 2018-06-20
Symbol table support

wesboyt:
After fixing IVM issues these commits handle some basic bugs around symbolID lookups and triple quote character escape sequences / illegal control characters. Also updated the readme to reflect the new levels of support.
Probably could be split into two PRs I can force push the head back 2 commits if you deem it necessary.

### DIFF
--- a/src/IonImport.ts
+++ b/src/IonImport.ts
@@ -28,58 +28,57 @@ import { SymbolIndex } from "./IonSymbolIndex";
  * @see http://amzn.github.io/ion-docs/symbols.html#imports
  */
 export class Import {
-  private readonly _offset: number;
-  private readonly _length: number;
-  private readonly index: SymbolIndex = {};
+    private readonly _offset: number;
+    private readonly _length: number;
+    private readonly index: SymbolIndex = {};
+    private readonly _parent : Import;
+    private readonly _symbolTable : SharedSymbolTable;
 
-  constructor(
-    private readonly _parent: Import,
-    private readonly _symbolTable: SharedSymbolTable,
-    length?: number
-  ) {
-    this._offset = (_parent && (_parent.offset + _parent.length)) || 1;
-    this._length = length || _symbolTable.symbols.length;
+    constructor(parent : Import, symbolTable : SharedSymbolTable, length?: number) {
+        this._parent = parent;
+        this._symbolTable = symbolTable;
+        this._offset = this.parent ? this.parent.offset + this.parent.length : 1;//this is wrong.
+        this._length = length || this.symbolTable.symbols.length;
 
-    let symbols: string[] = _symbolTable.symbols;
-    for (let i: number = 0; i < this._length; i++) {
-      this.index[symbols[i]] = this._offset + i;
-    }
-  }
-
-  getSymbol(symbolId: number) : string {
-    if (!isNullOrUndefined(this.parent)) {
-      let parentSymbol = this.parent.getSymbol(symbolId);
-      if (!isUndefined(parentSymbol)) {
-        return parentSymbol;
-      }
+        let symbols: string[] = this.symbolTable.symbols;
+        for (let i: number = 0; i < this.length; i++) {
+            this.index[symbols[i]] = this.offset + i;
+        }
     }
 
-    let index: number = symbolId - this.offset;
-    if (index < this.length) {
-      return this.symbolTable.symbols[index];
+    getSymbol(symbolId: number) : string {
+        if (!isNullOrUndefined(this.parent)) {
+            let parentSymbol = this.parent.getSymbol(symbolId);
+            if (!isUndefined(parentSymbol)) {
+                return parentSymbol;
+            }
+        }
+
+        let index: number = symbolId - this.offset;
+        if (index < this.length) {
+            return this.symbolTable.symbols[index];
+        }
+
+        return undefined;
     }
 
-    return undefined;
-  }
+    getSymbolId(symbol_: string) : number {
+        return (this.parent && this._parent.getSymbolId(symbol_)) || this.index[symbol_];
+    }
 
-  getSymbolId(symbol_: string) : number {
-    return (this.parent && this.parent.getSymbolId(symbol_))
-      || this.index[symbol_];
-  }
+    get parent() : Import {
+        return this._parent;
+    }
 
-  get parent() : Import {
-    return this._parent;
-  }
+    get offset() : number {
+        return this._offset;
+    }
 
-  get offset() : number {
-    return this._offset;
-  }
+    get length() : number {
+        return this._length;
+    }
 
-  get length() : number {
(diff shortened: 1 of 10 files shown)


src/IonImport.ts
```diff
@@ -28,58 +28,57 @@ import { SymbolIndex } from "./IonSymbolIndex";
...
-    private readonly _parent: Import,
-    private readonly _symbolTable: SharedSymbolTable,
-    length?: number
-  ) {
-    this._offset = (_parent && (_parent.offset + _parent.length)) || 1;
-    this._length = length || _symbolTable.symbols.length;
+    constructor(parent : Import, symbolTable : SharedSymbolTable, length?: number) {
+        this._parent = parent;
+        this._symbolTable = symbolTable;
+        this._offset = this.parent ? this.parent.offset + this.parent.length : 1;//this is wrong.
```
tgregg: Comments like "this is wrong" don't inspire confidence that this is not wrong...

src/IonLocalSymbolTable.ts
```diff
@@ -37,8 +37,7 @@ export class LocalSymbolTable  {
   }
 
   getSymbolId(symbol_: string) : number {
-    return this._import.getSymbolId(symbol_)
-      || this.index[symbol_];
+    return this._import.getSymbolId(symbol_) || this.index[symbol_];
```
tgregg: Can an LST have multiple imports?

src/IonParserBinaryRaw.ts
```diff
@@ -539,7 +539,7 @@ const MAX_BYTES_FOR_INT_IN_NUMBER = 6; // = floor(52 /* sig fig bits in number *
 const ZERO_POINT_ZERO = new Float64Array([0.0]);
 
 export class ParserBinaryRaw {
-  private buf = new ArrayBuffer(8);                   // TODO - what about threading?
+  private buf = new ArrayBuffer(8);                   // TODO - what about threading? no.
```
tgregg: If it's no longer a TODO, just remove the comment.

src/IonTextReader.ts
```diff
@@ -192,31 +192,29 @@ next() {
...
-        s = t._raw;   // TODO - this is a temp fix !!
-      }
-    }
-    else {
-      i = t.ionValue();
-      s = i.stringValue();
+    stringValue() : string {
+        this.load_raw();
+        if (this.isNull()) return (this._type === IonTypes.NULL) ? "null" : "null." + this._type.name;
+        if (this._type.scalar) {
```
tgregg: Should it be an error otherwise?

src/IonTextReader.ts
```diff
@@ -192,31 +192,29 @@ next() {
...
+            // BLOB is a scalar by you don't want to just use the string
+            // value otherwise all other scalars are fine as is
+            switch(this._type) {
+                case IonTypes.BLOB:
+                    throw new Error("Blobs currently unsupported.");
```
tgregg: Just through `stringValue`, right? Honestly, I'd be okay with only supporting `string` and `symbol` types through `stringValue`...

src/IonTextReader.ts
```diff
@@ -192,31 +192,29 @@ next() {
...
+                case IonTypes.SYMBOL:
+                    if(this._raw_type === T_IDENTIFIER && (this._raw.length > 1 && this._raw.charAt(0) === '$'.charAt(0))){
+                        let tempStr = this._raw.substr(1, this._raw.length);
+                        if (+tempStr === +tempStr) return this._symtab.getSymbol(Number(tempStr));//look up sid, +str === +str is a one line is integer hack
+                    } else if (this._raw_type === T_STRING1 && (this._raw.length > 1 && this._raw.charAt(0) === '$'.charAt(0))) {
+                        let tempStr = this._raw.substr(1, this._raw.length);
+                        if (+tempStr === +tempStr) return "'" + this._raw + "'";
```
tgregg: Adding single quotes here changes the data, which we shouldn't do (i.e. it would be roundtripped as '\'$99\'', when it began as '$99'). Logic should instead be added on the write side to identify text that looks like  a symbol identifier and write it as a quoted symbol when provided by the user. I suggest adding a test for this.

tgregg:
In the future, I'd request that you put whitespace changes in separate PRs marked as whitespace changes only. That makes those super simple to review and doesn't require reviewers to sift through a large CR to figure out what actually needs to be looked at. There's a risk of missing important stuff if the reviewer incorrectly assumes something is just a whitespace change.

src/IonImport.ts
wesboyt: sorry took me awhile to see how the exclusive/inclusive offset numbers matched up with maxID and multiple imports interacting with each other. It is correct I just wasn't convinced at first. I'll remove it.

src/IonLocalSymbolTable.ts
wesboyt: You have to chain their creation like: 
```
    var symbolTable = new ion.SharedSymbolTable('foo', 1, ['a', 'b', 'c']);
    var parent = new ion.Import(null, symbolTable, 1);
    var child = new ion.Import(parent, symbolTable);
```

wesboyt: theres a few tests in tests/unit that go over symbol tables and imports.

wesboyt: also that makes a lookup follow child to parent until it is able to resolve the sid

tgregg: This feels kind of icky. Do you think it's simpler than just having a list of Imports?

tgregg: Is the test coverage adequate? Given that ion-tests doesn't fully exercise shared symbol table cases (yet).

src/IonTextReader.ts
wesboyt: A number of the types have no representation outside of string/text/Number. Only the more complex ones like IonDecimal have real OO and functionality tied to them.

src/IonTextReader.ts
wesboyt: A number of the types have no representation outside of string/text/Number. Only the more complex ones like IonDecimal have real OO and functionality tied to them.

src/IonTextReader.ts
wesboyt: I dont think it does, I could be wrong though.

wesboyt: im wrong it does

src/IonTextReader.ts
wesboyt: yes

src/IonParserBinaryRaw.ts
wesboyt: done

src/IonTextReader.ts
wesboyt: fixed

PR #154, opened 2018-06-20
Ivm

wesboyt:
Fixed IVM logic that corrupted the state of the lexer.
Fixed logic around invalidIVM badfiles.
Fixed null indexing/roundtripping bug(for some reason encountering null would set indexes to -1,-1.
Fixed writing out raw ion from event stream when testing failed.

### DIFF
--- a/src/IonParserTextRaw.ts
+++ b/src/IonParserTextRaw.ts
@@ -563,7 +563,6 @@ export class ParserTextRaw {
                     symbol = this.get_value_as_string(T_IDENTIFIER);
                     kwt = get_type_from_name(symbol);
                 }
-                //this._start = tempNullStart;
                 this._start = -1;
                 this._end = -1;
             }
--- a/src/IonParserTextRaw.ts
+++ b/src/IonParserTextRaw.ts
@@ -563,8 +563,8 @@ export class ParserTextRaw {
                 }
                 this._start = -1;
                 this._end = -1;
-                this._value_push(kwt);
             }
+            this._value_push(kwt);
         } else {
             let ch = this._read_after_whitespace(true);
             if (ch == CH_CL && this._peek() == CH_CL) {
@@ -572,7 +572,7 @@ export class ParserTextRaw {
                 this._ann.push(symbol);
                 this._ops.unshift( calling_op );
             } else {
-                let kwt = is_keyword(symbol) ? get_keyword_type(symbol) : T_IDENTIFIER;
+                let kwt = T_IDENTIFIER;
                 this._unread(ch);
                 this._value_push(kwt); // put the value back on the stack
             }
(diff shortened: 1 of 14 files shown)


src/IonParserTextRaw.ts
```diff
@@ -561,6 +563,7 @@ export class ParserTextRaw {
                     symbol = this.get_value_as_string(T_IDENTIFIER);
                     kwt = get_type_from_name(symbol);
                 }
+                //this._start = tempNullStart;
```
tgregg: Delete

src/IonTextReader.ts
```diff
@@ -67,7 +69,7 @@ export class TextReader implements Reader {
   load_raw() {
     let t: TextReader = this;
     if (t._raw !== undefined) return;
-    if (t.isNull()) return;
+    //if (t.isNull()) return;
```
tgregg: Delete

src/IonTextReader.ts
```diff
@@ -87,41 +89,62 @@ export class TextReader implements Reader {
...
-        this._symtab = makeSymbolTable(this._cat, this);
-      }
-      else {
-        break;
-      }
+        this._raw_type = p.next();
+        if (this._raw_type === T_IDENTIFIER) {
+            if (this._depth > 0) break;
+            this.load_raw();
+            //if(this._raw !== "$ion_1_0") break;
```
tgregg: Delete

src/IonTextReader.ts
```diff
@@ -87,41 +89,62 @@ export class TextReader implements Reader {
...
+            if (!this.isIVM(this._raw)) break;
+            this._symtab = defaultLocalSymbolTable();
+            this._raw = undefined;
+            this._raw_type = undefined;
+        }else if (this._raw_type === T_STRING1) {
+            if (this._depth > 0) break;
+            this.load_raw();
+            if (this._raw !== "$ion_1_0") break;
```
tgregg: Is this needed?

wesboyt:
The eventstream/event changes are from files that were taken off the skip list and revealed additional bugs since the parser was able to get past the IVM at the top now. the switch was falling through the case and producing extra ContainerEnd events. Unrelated to IVM specifically I agree.

src/IonTextReader.ts
wesboyt: yes

tgregg: Oh because T_STRING1 mean's single-quoted symbol. Got it.

src/IonParserTextRaw.ts
wesboyt: done

src/IonTextReader.ts
wesboyt: done

src/IonTextReader.ts
wesboyt: done

PR #170, opened 2018-10-04
Blob support, utf8 support, amortized write cost support, binary writ…

wesboyt:
utf-8 support:
This change was first built to allow full integration testing, prior it was literally impossible to roundtrip between the text and binary since text could only read strings and binary could only write number[].
This required rewriting the entire library to support uint8array as its backing value, rewriting the utf8 encode/decode logic to use uint8array, and rebuilding the apis around getbytes/writeables.
These changes are visible in:
-src/IonLowLevelBinaryWriter.ts
-src/IonTextReader.ts
-src/IonTextWriter.ts
-src/IonUnicode.ts
-src/IonWriteable.ts

Blob support:
This change was built to allow blobs which previously had no api's/was backed by a string which made little sense. 
It is visible in:
-src/IonTextReader.ts
-src/IonText.ts
-src/IonTextWriter.ts

Amortized write costs:
This change was prompted by changing the backing buffer from number[] (really an object with number fields) to uint8array which is 10-12x faster and 20x faster with amortized costs. (also 4x smaller 8 bits vs 32)
It is visible in:
-src/IonWriteable.ts

Low level binary writer precalc lengths:
This change was due to bugs in the binary writer once integration testing was working. The binary writer would churn out garbage and randomly padded values based on a shoddy low level number writer implementation.
It is visible in:
-src/IonLowLevelBinaryWriter.ts

### DIFF
--- a/src/IonLowLevelBinaryWriter.ts
+++ b/src/IonLowLevelBinaryWriter.ts
@@ -11,8 +11,6 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
  * language governing permissions and limitations under the License.
  */
-import { isNumber } from "./IonUtilities";
-import { isUndefined } from "./IonUtilities";
 import { Writeable } from "./IonWriteable";
 
 /**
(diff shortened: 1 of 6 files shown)


tgregg:
Would you mind expanding the description a bit to describe each of the different parts of this review individually? I think that could make this fairly large review easier to digest. I'm really just looking for a brief description of what each change does and which files in the review it touches.

PR #177, opened 2019-04-13
Add typescript declarations to distribution artifacts

bglowney:
@wesboyt 

Here is a change to get around the issue we discussed preventing inclusion of typescript declarations. The change moves "BigInteger.js" out of the src, but keeps "BigInteger.d.ts" in the source. This way you don't have to set the tsc `allowJs` compiler flag. After typescript compilation there is a new `grunt copy:bigInt` task that copies "BigInteger.js" into the distribution artifacts along with its type declaration.

Tested by compiling the ion-js typescript and then depending on it locally in another typescript project. I was able to compile the other project successfully while importing on BigInteger.

Note, "BigInteger.js" doesn't need to live in the base directory; it could be moved somewhere else.
Note, the tests are not passing for me on this branch, but I understand that was already the case.


wesboyt:
SICK!!! thanks so much.

PR #178, opened 2019-04-17
Support printing blobs as base64

bglowney:
@wesboyt 

I think these are the changes we talked about to support printing blobs as base64 strings with ion-js text.

I did a little test with this locally printing a struct with a blob member.

These are just the patches copied out of your commit [here](https://github.com/amzn/ion-js/commit/58282cf1bb9228d24f646b3dbe60bcae8ac0103e)

Not sure what other testing would be useful, so please suggest.


wesboyt:
I built a new fully passing branch to fulfill this request here: https://github.com/amzn/ion-js/tree/masterblob

PR #198, opened 2019-06-03
Add millisecond precision for text Timestamps

bglowney:
@wesboyt I need this change to parse UTC ISO-8601 Ion text timestamps with second and millisecond precision. I've tested it locally by parsing string input like "2019-05-31T00:04:19.305Z". 

This package still doesn't build due to other existing issues.


wesboyt:
wont merge, breaks contracts.

Issue #202, opened 2019-06-11
Top-level value $ion_schema_1_0 results in an error


pbcornell:
Attempting to parse:

```
$ion_schema_1_0
```

produces `Error:  Only ion version 1.0 is supported.`

1. I believe this is a valid Ion value
1. The error message should spell "ion" as "Ion"

wesboyt:
merged

PR #203, opened 2019-06-17
isIVM fix for similar but non IVM symbols.

wesboyt:
#202 

Instead of throwing on all symbols with $ion_ we now throw on $ion_(0-9xxx...)_(0-9xxx...) when not exactly $ion_1_0.

### DIFF
--- a/src/IonTextReader.ts
+++ b/src/IonTextReader.ts
@@ -90,8 +90,10 @@ export class TextReader implements Reader {
 
     isIVM(input : string) : boolean {
         if (this.depth() > 0) return false;
-        if (input.length < 6 || this.annotations().length > 0) return false;
-        let prefix = "$ion_";
+        const ivm = "$ion_1_0";
+        const prefix = "$ion_";
+        if (input.length < ivm.length || this.annotations().length > 0) return false;
+
         let i = 0;
 
         while (i < prefix.length) {
@@ -111,7 +113,7 @@ export class TextReader implements Reader {
             if (ch < '0' || ch > '9') return false;
             i++;
         }
-        if (input !== "$ion_1_0") throw new Error("Only Ion version 1.0 is supported.");
+        if (input !== ivm) throw new Error("Only Ion version 1.0 is supported.");
         return true;
     }
 
--- a/src/IonTextReader.ts
+++ b/src/IonTextReader.ts
@@ -100,14 +100,15 @@ export class TextReader implements Reader {
         }
 
         while (i < input.length && input.charAt(i) != '_') {
-            let tempChar = input.charAt(i);
-            if (tempChar < '0' || tempChar > '9') return false;
+            let ch = input.charAt(i);
+            if (ch < '0' || ch > '9') return false;
             i++;
         }
         i++;
+
         while (i < input.length) {
-            let tempChar = input.charAt(i);
-            if (tempChar < '0' || tempChar > '9') return false;
+            let ch = input.charAt(i);
+            if (ch < '0' || ch > '9') return false;
             i++;
         }
         if (input !== "$ion_1_0") throw new Error("Only Ion version 1.0 is supported.");
--- a/src/IonTextReader.ts
+++ b/src/IonTextReader.ts
@@ -89,33 +89,33 @@ export class TextReader implements Reader {
   }
 
   isIVM(input : string, depth : number, annotations : string[]) : boolean {
-        if (depth > 0) return false;
-        const ivm = "$ion_1_0";
-        const prefix = "$ion_";
-        if (input.length < ivm.length || annotations.length > 0) return false;
+    if (depth > 0) return false;
+    const ivm = "$ion_1_0";
+    const prefix = "$ion_";
+    if (input.length < ivm.length || annotations.length > 0) return false;
 
-        let i = 0;
+    let i = 0;
 
-        while (i < prefix.length) {
-            if (prefix.charAt(i) !== input.charAt(i)) return false;
-            i++;
-        }
-
-        while (i < input.length && input.charAt(i) != '_') {
-            let ch = input.charAt(i);
-            if (ch < '0' || ch > '9') return false;
-            i++;
-        }
-        i++;
-
-        while (i < input.length) {
-            let ch = input.charAt(i);
-            if (ch < '0' || ch > '9') return false;
-            i++;
-        }
-        if (input !== ivm) throw new Error("Only Ion version 1.0 is supported.");
-        return true;
+    while (i < prefix.length) {
+      if (prefix.charAt(i) !== input.charAt(i)) return false;
+      i++;
     }
+
+    while (i < input.length && input.charAt(i) != '_') {
+      let ch = input.charAt(i);
+      if (ch < '0' || ch > '9') return false;
+      i++;
+    }
+    i++;
+
+    while (i < input.length) {
+      let ch = input.charAt(i);
+      if (ch < '0' || ch > '9') return false;
+      i++;
+    }
+    if (input !== ivm) throw new Error("Only Ion version 1.0 is supported.");
+    return true;
+  }
 
     isLikeIVM() : boolean {
       return false;
--- a/src/IonTextReader.ts
+++ b/src/IonTextReader.ts
@@ -98,19 +98,19 @@ export class TextReader implements Reader {
             if (prefix.charAt(i) !== input.charAt(i)) return false;
             i++;
         }
-        let tempChar;
-        while (input.charAt(i) != '_') {
-            tempChar = input.charAt(i);
+
+        while (i < input.length && input.charAt(i) != '_') {
+            let tempChar = input.charAt(i);
             if (tempChar < '0' || tempChar > '9') return false;
             i++;
         }
         i++;
-        while (i < input.length) {
-            tempChar = input.charAt(i);
+        while (i < input.length && i < input.length) {
+            let tempChar = input.charAt(i);
             if (tempChar < '0' || tempChar > '9') return false;
             i++;
         }
-        if (input !== "$ion_1_0") throw new Error("Only ion version 1.0 is supported.");
+        if (input !== "$ion_1_0") throw new Error("Only Ion version 1.0 is supported.");
         return true;
     }
 
--- a/src/IonTextReader.ts
+++ b/src/IonTextReader.ts
@@ -89,28 +89,28 @@ export class TextReader implements Reader {
   }
 
     isIVM(input : string) : boolean {
-        if(this.depth() > 0) return false;
-        if(input.length < 6 || this.annotations().length > 0) return false;
+        if (this.depth() > 0) return false;
+        if (input.length < 6 || this.annotations().length > 0) return false;
         let prefix = "$ion_";
         let i = 0;
 
-        while(i < prefix.length){
-            if(prefix.charAt(i) !== input.charAt(i)) return false;
+        while (i < prefix.length) {
+            if (prefix.charAt(i) !== input.charAt(i)) return false;
             i++;
         }
         let tempChar;
-        while(input.charAt(i) != '_'){
+        while (input.charAt(i) != '_') {
             tempChar = input.charAt(i);
-            if(tempChar < '0' || tempChar > '9') return false;
+            if (tempChar < '0' || tempChar > '9') return false;
             i++;
         }
         i++;
-        while(i < input.length){
+        while (i < input.length) {
             tempChar = input.charAt(i);
-            if(tempChar < '0' || tempChar > '9') return false;
+            if (tempChar < '0' || tempChar > '9') return false;
             i++;
         }
-        if(input !== "$ion_1_0") throw new Error("Only ion version 1.0 is supported.");
+        if (input !== "$ion_1_0") throw new Error("Only ion version 1.0 is supported.");
         return true;
     }
 
--- a/src/IonTextReader.ts
+++ b/src/IonTextReader.ts
@@ -94,7 +94,7 @@ export class TextReader implements Reader {
         let prefix = "$ion_";
         let i = 0;
 
-        while (i < input.length && i < prefix.length) {
+        while (i < prefix.length) {
             if (prefix.charAt(i) !== input.charAt(i)) return false;
             i++;
         }
--- a/src/IonTextReader.ts
+++ b/src/IonTextReader.ts
@@ -88,11 +88,11 @@ export class TextReader implements Reader {
     }
   }
 
-    isIVM(input : string) : boolean {
-        if (this.depth() > 0) return false;
+  isIVM(input : string, depth : number, annotations : string[]) : boolean {
+        if (depth > 0) return false;
         const ivm = "$ion_1_0";
         const prefix = "$ion_";
-        if (input.length < ivm.length || this.annotations().length > 0) return false;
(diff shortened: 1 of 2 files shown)


src/IonTextReader.ts
```diff
@@ -92,8 +92,23 @@ export class TextReader implements Reader {
...
+            if(tempChar < '0' || tempChar > '9') return false;
+            i++;
+        }
+        i++;
+        while(i < input.length){
+            tempChar = input.charAt(i);
+            if(tempChar < '0' || tempChar > '9') return false;
+            i++;
         }
         if(input !== "$ion_1_0") throw new Error("Only ion version 1.0 is supported.");
```
pbcornell: 'ion' should be spelled 'Ion'

src/IonTextReader.ts
```diff
@@ -92,8 +92,23 @@ export class TextReader implements Reader {
...
         if(input.length < 6 || this.annotations().length > 0) return false;
         let prefix = "$ion_";
-        for(let i = 0; i < prefix.length; i++){
+        let i = 0;
+
+        while(i < prefix.length){
             if(prefix.charAt(i) !== input.charAt(i)) return false;
+            i++;
+        }
+        let tempChar;
```
pbcornell: - define this var within the loop, as it has no value outside the loop
- consider renaming it to 'ch'

src/IonTextReader.ts
```diff
@@ -92,8 +92,23 @@ export class TextReader implements Reader {
...
+        while(input.charAt(i) != '_'){
```
pbcornell: potential index out of bounds issue here?

wesboyt: it returns undefined which triggers the false case.

src/IonTextReader.ts
wesboyt: done

src/IonTextReader.ts
wesboyt: done

src/IonTextReader.ts
```diff
@@ -89,13 +89,29 @@ export class TextReader implements Reader {
   }
 
     isIVM(input : string) : boolean {
-        if(this.depth() > 0) return false;
-        if(input.length < 6 || this.annotations().length > 0) return false;
+        if (this.depth() > 0) return false;
+        if (input.length < 6 || this.annotations().length > 0) return false;
```
pbcornell: i suspect we mean `input.length <= prefix.length`, yes?

tests/unit/IonTextReaderTest.js
```diff
@@ -125,6 +125,39 @@
       assert.equal(ionReader.fieldName(), "innerkey2");
     };
 
+      suite['Accept IVM like symbols and throw on IVMs except $ion_1_0'] = function() {
```
pbcornell: the nested try/catch structure is unusual, and there's quite a bit of repetition.  i'd suggest defining a helper function with parameters for the string to test and an indication of what the expected behavior is (false/true/error).

and let's add `$ion_1`, `$ion_1_1`, and `$ion_1_a` as additional test cases

src/IonTextReader.ts
```diff
@@ -89,13 +89,29 @@ export class TextReader implements Reader {
...
         let prefix = "$ion_";
-        for(let i = 0; i < prefix.length; i++){
-            if(prefix.charAt(i) !== input.charAt(i)) return false;
+        let i = 0;
+
+        while (i < prefix.length) {
```
pbcornell: potential simplification:  call `input.startsWith(prefix)` instead of this loop

ok if you'd prefer to leave as-is

src/IonTextReader.ts
```diff
@@ -89,13 +89,29 @@ export class TextReader implements Reader {
...
+            if (prefix.charAt(i) !== input.charAt(i)) return false;
+            i++;
+        }
+
+        while (i < input.length && input.charAt(i) != '_') {
```
pbcornell: we could potentially combine this while loop and the next, returning false if a non-digit and non underscore character is encountered, but your approach ensures there's a single underscore, which is nice, too.  fine as-is

src/IonTextReader.ts
pbcornell: looks like `"hello".charAt(1000)` returns an empty string (different than `undefined`), but the effect is the same.  ok

src/IonTextReader.ts
wesboyt: looks like indexOf() is about 100x faster than startswith. I'll alter all the char checks

wesboyt:
I believe all comments were addressed.

tests/unit/IonTextReaderTest.js
```diff
@@ -125,6 +125,29 @@
...
+          let shouldPass = ["$ion_1_0", "$ion_schema_1_0", "$ion_1", "$ion_1_a", "$ion_", "ion_1_"];
+          let shouldFail = ["$ion_2_0", "$ion_1_999", "$ion_999_0", "$ion_1_1", "$ion_1_00"];
+          for(let input in shouldPass) {
```
pbcornell: add space after `for`

tests/unit/IonTextReaderTest.js
```diff
@@ -125,6 +125,29 @@
...
+              let ionReader = ion.makeReader(input);
+              try {
+                  while (ionReader.next() !== undefined) {}
+              } catch(error) {
+                  throw new Error("Allowable IVM like symbol threw an error " + error);
+              }
+          }
+
+          for(let input in shouldFail) {
```
pbcornell: add space after `for`

tests/unit/IonTextReaderTest.js
```diff
@@ -125,6 +125,29 @@
...
+              let ionReader = ion.makeReader(input);
+              let caughtError = false;
+              try {
+                  while (ionReader.next() !== undefined) {}
+              } catch(error) {
+                caughtError = true;
+              }
+              if(!caughtError) throw new Error("Unsupported IVM symbol did not throw an error.");
```
pbcornell: add space after `if`

tests/unit/IonTextReaderTest.js
```diff
@@ -125,6 +125,29 @@
...
+              }
+          }
+
+          for(let input in shouldFail) {
+              let ionReader = ion.makeReader(input);
+              let caughtError = false;
+              try {
+                  while (ionReader.next() !== undefined) {}
+              } catch(error) {
+                caughtError = true;
```
pbcornell: fix indentation

tests/unit/IonTextReaderTest.js
pbcornell: if we move this inside the try block immediately after the while loop, the `caughtError` variable can be removed

tests/unit/IonTextReaderTest.js
```diff
@@ -125,6 +125,29 @@
       assert.equal(ionReader.fieldName(), "innerkey2");
     };
 
+      suite['Accept IVM like symbols and throw on IVMs except $ion_1_0'] = function() {
+          let shouldPass = ["$ion_1_0", "$ion_schema_1_0", "$ion_1", "$ion_1_a", "$ion_", "ion_1_"];
+          let shouldFail = ["$ion_2_0", "$ion_1_999", "$ion_999_0", "$ion_1_1", "$ion_1_00"];
+          for(let input in shouldPass) {
+              let ionReader = ion.makeReader(input);
+              try {
+                  while (ionReader.next() !== undefined) {}
```
pbcornell: this is a unit test, so (generally speaking) we should narrow the scope of what we're invoking.  invoking `ionReader.next()` seems fairly indirect.  is there any reason not to call `isIVM()` directly?

tests/unit/IonTextReaderTest.js
```diff
@@ -125,6 +125,29 @@
       assert.equal(ionReader.fieldName(), "innerkey2");
     };
 
+      suite['Accept IVM like symbols and throw on IVMs except $ion_1_0'] = function() {
+          let shouldPass = ["$ion_1_0", "$ion_schema_1_0", "$ion_1", "$ion_1_a", "$ion_", "ion_1_"];
```
pbcornell: shouldPass/shouldFail doesn't tell the whole story, does it?  shouldFail is really shouldThrow, and some of the values in shouldPass should produce `true`, while others should produce `false`.  let's assert the expected behavior for all test values

i'm thinking of something more along the following lines:
```javascript
suite['...'] = function() {
  test_isIVM("$ion_1_0", true);
  test_isIVM("$ion_schema_1_0", false);
  test_isIVM("$ion_2_0", "throw");
  ...
}

function test_isIVM(value, expect) {
  actual = isIVM(value);
  if (expect == "throw") {
    throw new Error("Expected " + value + " to throw an error.");
  }
  assert expect == actual;
}
```

thoughts?

tests/unit/IonTextReaderTest.js
wesboyt: nice

tests/unit/IonTextReaderTest.js
wesboyt: I think if the testing suites delineated between unit tests and integration tests I'd agree. I don't think .next() is a indirect in the context of an integration test. The intention is making the reader and parser is validate input correctly. I think a separate unit test could add value but this isn't intended to be a true unit test.

tests/unit/IonTextReaderTest.js
wesboyt: I think the description of the suite does the same thing?

pbcornell: not sure what you mean.  the description is just a string--it doesn't do anything to verify that `isIVM()` either returns the expected value or throws

tests/unit/IonTextReaderTest.js
pbcornell: i think a unit test makes sense here;  help me understand why you're thinking it should be an integration test?  (and if that's the approach, how does it verify that `isIVM()` returns true or false when it is expected to?)

wesboyt: Ah I see, if the isIVM function throwing was your original issue I can compare the error message! easy change.

tests/unit/IonTextReaderTest.js
wesboyt: same response as above! I'll get right on it.

wesboyt:
I believe I've responded to comments, let me know!

tests/unit/IonTextReaderTest.js
```diff
@@ -1,130 +1,171 @@
...
+                test_isIVM(textReader, unsupportedIVM[i], false, 0, ["taco"]);
+            }
+        };
+
+        function test_isIVM(reader, value, expected, depth, annotations) {
+            if (expected === "throw") {
+                try {
+                    reader.isIVM(value, depth, annotations);
+                } catch(error) {
+                    if (error.message !== "Only Ion version 1.0 is supported.") {
```
pbcornell: checking the error message is generally fairly brittle, i'd suggest removing it

tests/unit/IonTextReaderTest.js
```diff
@@ -1,130 +1,171 @@
...
+                        throw new Error(`input: ${value} threw an unexpected error: ${error.message}.`);
+                    }
+                    return;
+                }
+                throw new Error("Expected " + value + " to throw an error.");
```
pbcornell: move this immediately after the `reader.isIVM()` call

tests/unit/IonTextReaderTest.js
```diff
@@ -1,130 +1,171 @@
...
+
+        function test_isIVM(reader, value, expected, depth, annotations) {
+            if (expected === "throw") {
+                try {
+                    reader.isIVM(value, depth, annotations);
+                } catch(error) {
+                    if (error.message !== "Only Ion version 1.0 is supported.") {
+                        throw new Error(`input: ${value} threw an unexpected error: ${error.message}.`);
+                    }
+                    return;
```
pbcornell: not necessary if you move the `throw new Error()` line immediately after `reader.isIVM()`

tests/unit/IonTextReaderTest.js
```diff
@@ -1,130 +1,171 @@
...
+                }
+                throw new Error("Expected " + value + " to throw an error.");
+            } else {
+                let actual = reader.isIVM(value, depth, annotations);
+                assert.equal(expected, actual);
```
pbcornell: chai's [assert.equal](https://www.chaijs.com/api/assert/#method_equal) signature is `assert.equal(actual, expected)` -- where `actual` and `expected` are in alphabetical order (this is the opposite of JUnit's ordering, fyi).  please follow chai's signature wrt parameter ordering--otherwise, test failure messages are likely to cause confusion

tests/unit/IonTextReaderTest.js
```diff
@@ -1,130 +1,171 @@
...
-      assert.equal(ionReader.value(), "string2");
-    };
-
-    suite['Parse through struct can skip over nested containers'] = function() {
-      var ionToRead = "{ outerkey1 : { innerkey1 : {a1: \"a1\", b1: \"b1\"} }, outerkey2 : { innerkey2 : {a2: \"a2\", b2: \"b2\"} } }";
-      var ionReader = ion.makeReader(ionToRead);
-      ionReader.next();
-
-      assert.equal(ion.IonTypes.STRUCT, ionReader.valueType());
+* Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
pbcornell: i think other files have one space before each `*` -- let's continue following that

src/IonTextReader.ts
```diff
@@ -88,14 +88,32 @@ export class TextReader implements Reader {
...
   }
 
-    isIVM(input : string) : boolean {
-        if(this.depth() > 0) return false;
-        if(input.length < 6 || this.annotations().length > 0) return false;
-        let prefix = "$ion_";
-        for(let i = 0; i < prefix.length; i++){
-            if(prefix.charAt(i) !== input.charAt(i)) return false;
+  isIVM(input : string, depth : number, annotations : string[]) : boolean {
+        if (depth > 0) return false;
```
pbcornell: to many spaces for this indent (and lines below)

tests/unit/IonTextReaderTest.js
```diff
@@ -1,130 +1,171 @@
...
+            let isNotIVM = ["$ion_schema_1_0", "$ion_1", "$ion_1_a", "$ion_", "ion_1_"];
+            let unsupportedIVM = ["$ion_2_0", "$ion_1_999", "$ion_999_0", "$ion_1_1", "$ion_1_00"];
+
+            test_isIVM(textReader,"$ion_1_0", true, 0, []);
+            test_isIVM(textReader,"$ion_1_0", false, 1, []);
+            test_isIVM(textReader,"$ion_1_0", false, 0, ["taco"]);
+            for (let i = 0; i < isNotIVM.length; i++) {
+                test_isIVM(textReader, isNotIVM[i], false, 0, []);
+            }
+            for (let i = 0; i < unsupportedIVM.length; i++) {
```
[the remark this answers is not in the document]
wesboyt: lol good call

tests/unit/IonTextReaderTest.js
pbcornell: please change this line to:  `assert.equal(actual, expected)`

Issue #206, opened 2019-07-01
Text writer is not adding single quotes when writing annotations with booleans


raganhan:
```
writer.writeInt(1, ["false"]);
writer.close();
console.log(ion.decodeUtf8(writer.getBytes()));
```

outputs `false::1` when it should be `'false'::1` as `false` is an Ion boolean an annotations **must** be symbols

raganhan:
This works fine for numbers, but it's best to add tests for all data types 

```
writer.writeInt(1, ["1"]);
writer.close();
console.log(ion.decodeUtf8(writer.getBytes()));
```
outputs `'1'::1`

pbcornell:
Addressed in #263, closing.

```javascript
writer.writeInt(1, ["false"]);
writer.close();
console.log(ion.decodeUtf8(writer.getBytes()));
```
now correctly produces:
```
'false'::1
```

Issue #174, opened 2018-11-07
Keywords in SymbolTokens are currently not writable


sharanrokx:
currently writing symbol logic is below
```
    private writeSymbolToken(s: string) : void {
        if (isIdentifier(s)) {
            if(s.length > 1 && s.charAt(0) === '$'.charAt(0)){
                let tempStr = s.substr(1, s.length);
                if (+tempStr === +tempStr) {//+str === +str is a one line is integer hack
                    this.writeable.writeByte(CharCodes.SINGLE_QUOTE);
                    this.writeable.writeStream(escape(s, SymbolEscapes));
                    this.writeable.writeByte(CharCodes.SINGLE_QUOTE);
                } else {
                    this.writeUtf8(s);
                }
            } else {
                this.writeUtf8(s);
            }
        } else if ((!this.isTopLevel) && (this.currentContainer.containerType === TypeCodes.SEXP) && isOperator(s)) {
            this.writeUtf8(s);
        } else {
            this.writeable.writeByte(CharCodes.SINGLE_QUOTE);
            this.writeable.writeStream(escape(s, SymbolEscapes));
            this.writeable.writeByte(CharCodes.SINGLE_QUOTE);
        }
    }
```
We need to change its name to writeSymbolValue and add another function
```
    private writeSymbolToken(s: string) : void {
                if (s === "'false'" || s === "'true'" || s === "'null'") {
                    this.writeUtf8(s);
                } else {
                    this.writeSymbolValue(s);
                }
    }
```
writeSymbol() should use writeSymbolValue() & writeAnnotations(), writeFieldName() should use writeSymbolToken()

API's can call 
```writeAnnotations(["'false'"]) ```
for printing valid ion. This preserves bidirectional reader-writer functionality.

Here API's are using javascript functionality of equivalent representation of single & double quotes to send them in string without adding escape characters. They can also use literal i.e. backtick
```writeAnnotations([`'false'`])```

wesboyt:
We cant allow the boolean false in all places where symbol false gets called. We only guarantee data model equivalence roundtrip.

symbol tokens and import descriptors have issues tracking them.

raganhan:
As Wes mentioned there is a bug here when using `writeAnnotations` passing either `"false"` or `"true"`. 

`writeAnnotations` contract takes in a array of strings with the textual representation of symbols so:
* `writeAnnotations(["false"])` should represents annotating with the symbol `'false'` 
[...]

PR #208, opened 2019-07-01
Keyword text fixes

wesboyt:
Issues:
#206
#174
Description:
Adds single quotes to keywords.

### DIFF
--- a/src/IonParserTextRaw.ts
+++ b/src/IonParserTextRaw.ts
@@ -157,7 +157,7 @@ function get_type_from_name(str: string) : number {
   throw new Error("Unknown type: " + str + ".");
 }
 
-function is_keyword(str: string) : boolean {
+export function is_keyword(str: string) : boolean {
     return (str === "null") || (str === "true") || (str === "false") || (str === "nan") || (str === "+inf") || (str === "-inf");
 }
 
--- a/src/IonTextWriter.ts
+++ b/src/IonTextWriter.ts
@@ -18,6 +18,7 @@ import { Timestamp } from "./IonTimestamp";
 import { TypeCodes } from "./IonBinary";
 import { Writeable } from "./IonWriteable";
 import { Writer } from "./IonWriter";
+import { is_keyword } from "./IonParserTextRaw";
 
 type Serializer<T> = (value: T) => void;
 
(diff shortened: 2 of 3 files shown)


raganhan:
Travis is failing for this PR but it's passing for `master`. Did this PR cause the failure or is the build unstable?

raganhan:
Please change your commit message to something more descriptive. For example mention the associated issues, describe what is the bug and the fix

src/IonParserTextRaw.ts
```diff
@@ -157,7 +157,7 @@ function get_type_from_name(str: string) : number {
   throw new Error("Unknown type: " + str + ".");
 }
 
-function is_keyword(str: string) : boolean {
+export function is_keyword(str: string) : boolean {
```
raganhan: Think it's best to extract this function into a component that's used by both `IonParserTextRaw` and other places that need to figure out if a string is a keyword instead of exposing it here to avoid weird coupling between `IonParserTextRaw` and other classes

tests/unit/IonTextWriterTest.js
```diff
@@ -358,6 +358,43 @@
           writer => { writer.writeStruct(); writer.endContainer(); writer.writeStruct() },
           '{}\n{}');
 
+      //Annotations
+
+      writerTest('Escapes keyword null in an annotation',
```
raganhan: Tests look good, 

Add another one for text with spaces, e.g. `"A text with spaces"`

src/IonTextWriter.ts
```diff
@@ -370,7 +371,11 @@ export class TextWriter implements Writer {
                     this.writeUtf8(s);
                 }
             } else {
```
raganhan: use `else if {` instead of `else { if {`

Think `writeSymbolToken` can be simplified to something like: 
```
if( is_keyword(s) || (isIdentifier(s) && !isSid(s))) {
   // write s as is 
}
else {
   // escape s
}
```

`isSid(s)` function checks for text like `$1` which needs to be escaped to represent `'$1'` instead of the symbol with SID `1`

src/IonParserTextRaw.ts
wesboyt: its dependent on a buttload of constants defined in the parser, pulling all of them out felt like a worse option.

raganhan:
As we talked extensively offline the dependency between a reader and writer component that was introduced is problematic. 

Also travis is failing

pbcornell:
This effort was resumed and completed under #263, which has been merged into the development branch.

Issue #210, opened 2019-07-03
Reading binary UInt is broken for values requiring exactly 32 bits


zslayton:
Discovered while working on #209, part of #205.

Javascript's bitwise operators [convert their operands into 32-bit *signed* integers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_Operators#Signed_32-bit_integers). This means that we can only store unsigned values using up to 31 bits (inclusive) in each `number` value.

`ParserBinaryRaw::load_value` will call `readUnsignedLongInt()` if the `_len` value is > 4, which covers all numbers needing more than 32 bits. We need a special case for reading unsigned integer values where `_len` is 4, but the first byte (which we can `peek()` at before officially reading it) is >127, requiring the highest bit to be set. This will allow us to call `readUnsignedLongInt()` on values which require exactly 32 bits.

pbcornell:
We believe this is fixed, closing

zslayton:
Closed by #355.

Issue #209, opened 2019-07-03
Unit tests for reading binary UInt


zslayton:
Related to #205.

Add unit tests for reading [binary UInt](http://amzn.github.io/ion-docs/docs/binary.html#uint-and-int-fields). This will involve some refactoring to expose a this functionality in `ParserBinaryRaw`.

zslayton:
Addressed by PR #211.

PR #211, opened 2019-07-04
Unit tests: reading binary UInt

zslayton:
*Issue #, if available:* #209 

*Description of changes:*

Adds `tests/IonParserBinaryRawTest.js`, which contains unit tests for:
* `ParserBinaryRaw::readUnsignedInt`
* `ParserBinaryRaw::readUnsignedLongInt`

This testing surfaced issue #210. UInts which require exactly 32 bits to represent currently cause read failures.

### DIFF
--- a/src/IonParserBinaryRaw.ts
+++ b/src/IonParserBinaryRaw.ts
@@ -230,14 +230,16 @@ export class ParserBinaryRaw {
         }
 
         // We overflowed
-        if (value < 0) {
+        if (numberOfBytes > 4 || value < 0) {
             throw new Error("Attempted to read an unsigned int that was larger than 31 bits."
                 + " Use readUnsignedLongIntFrom instead. UInt size: " + numberOfBytes + ", value: " + value
             );
         }
         // Fewer bytes than the required `numberOfBytes` were available in the input
-        if (byte === EOF)
-            return undefined;
+        if (byte === EOF) {
+            throw new Error("Ran out of data while reading a " + numberOfBytes + "-byte unsigned int.");
+        }
+
         return value;
     }
 
--- a/src/IonParserBinaryRaw.ts
+++ b/src/IonParserBinaryRaw.ts
@@ -155,13 +155,13 @@ export class ParserBinaryRaw {
         }
     }
 
-    private readVarUnsignedInt() : number {
+    static readVarUnsignedFrom(input: BinarySpan) : number {
         let numberOfBits = 0;
         let byte;
         let magnitude = 0;
 
         while (true) {
-            byte = this._in.next();
+            byte = input.next();
             magnitude = (magnitude << 7) | (byte & 0x7F);
             numberOfBits += 7;
             if (byte & 0x80) break;
@@ -169,6 +169,10 @@ export class ParserBinaryRaw {
 
         if(numberOfBits > 32) throw new Error("Values larger than 32 bits need to be marshalled using LongInt")
         return magnitude;
+    }
+
+    private readVarUnsignedInt() : number {
+        return ParserBinaryRaw.readVarUnsignedFrom(this._in);
     }
 
     private readVarSignedInt() : number {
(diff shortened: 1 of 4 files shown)


src/IonParserBinaryRaw.ts
```diff
@@ -155,13 +155,13 @@ export class ParserBinaryRaw {
         }
     }
 
-    private readVarUnsignedInt() : number {
+    static readVarUnsignedFrom(input: BinarySpan) : number {
```
zslayton: The `readUnsignedInt` and `readVarUnsignedLongInt` methods both rely on internal state from the `ParserBinaryRaw` class. In order to make them testable, I've moved their core logic out into two new static methods: `readUnsignedIntFrom` and `readVarUnsignedLongIntFrom`. These methods allow you to bring your own state, making them easier to test.

src/IonParserBinaryRaw.ts
```diff
@@ -213,28 +217,40 @@ export class ParserBinaryRaw {
...
-    private readUnsignedInt() : number {
-        var v = 0, b;
+    static readUnsignedIntFrom(input: BinarySpan, numberOfBytes: number) {
+        let value = 0, bytesRead = 0, byte;
+        if (numberOfBytes < 1)
+            return 0;
 
-        // they have to tell us the length
-        if (this._len < 1) return 0;
+        while (bytesRead < numberOfBytes) {
```
zslayton: I expanded lots of single-character variable names into words for readability. The logic is the same, however.

src/IonParserBinaryRaw.ts
```diff
@@ -213,28 +217,40 @@ export class ParserBinaryRaw {
...
-            v = v | (b & 0xff);
+        // We overflowed
+        if (value < 0) {
+            throw new Error("Attempted to read an unsigned int that was larger than 31 bits."
+                + " Use readUnsignedLongIntFrom instead. UInt size: " + numberOfBytes + ", value: " + value
+            );
         }
+        // Fewer bytes than the required `numberOfBytes` were available in the input
+        if (byte === EOF)
+            return undefined;
```
zslayton: I believe that this should throw an `Exception` rather than returning `undefined`, but I'll leave that for another PR.

The original code accidentally omitted the `return` keyword, so it did nothing at all in this particular error case.

src/IonTests.ts
```diff
@@ -20,4 +20,6 @@ export { Import } from './IonImport';
 export { LocalSymbolTable } from './IonLocalSymbolTable';
 export { LowLevelBinaryWriter } from './IonLowLevelBinaryWriter';
 export { NullNode } from './IonBinaryWriter';
-export { BinaryWriter } from "./IonBinaryWriter";
\ No newline at end of file
+export { BinaryWriter } from "./IonBinaryWriter";
+export { ParserBinaryRaw } from "./IonParserBinaryRaw";
```
zslayton: This makes the `ParserBinaryRaw` and `BinarySpan` classes accessible from within the unit test framework.

src/IonParserBinaryRaw.ts
```diff
@@ -213,28 +217,40 @@ export class ParserBinaryRaw {
...
+        }
 
-        // shift in all but the last byte (we've already read the first)
-        while (this._len > 0) {
-            b = this._in.next();
-            this._len--;
-            v = v << 8;
-            v = v | (b & 0xff);
+        // We overflowed
+        if (value < 0) {
```
raganhan: is there a chance to overflow twice so it's back to positive?

src/IonParserBinaryRaw.ts
raganhan: Agreed on the Exception. Is there an issue tracking this so it's not lost?

tests/unit/IonParserBinaryRawTest.js
```diff
@@ -0,0 +1,81 @@
...
+    };
+
+    // Registers the provided test with the test suite.
+    let registerTest = function(testName, test, throwsException) {
+      suite[testName] = throwsException
+          ? function() { assert.throws(test, Error); }
+          : test;
+    }
+
+    let readUnsignedIntTest = function(bytes, expected, throwsException) {
```
raganhan: better to have two functions that call this one passing `true` or `false` for better test readability, e.g. `readUnsignedIntTest` and `readUnsignedIntTestReadMoreThanLength`

tests/unit/IonParserBinaryRawTest.js
```diff
@@ -0,0 +1,81 @@
...
+    readUnsignedIntTest([0x00], 0);
+    readUnsignedIntTest([0x01], 1);
+    readUnsignedIntTest([0x0F], 15);
+    readUnsignedIntTest([0x00, 0x00], 0);
+    readUnsignedIntTest([0xFF,], Math.pow(2, 8) - 1);
+    readUnsignedIntTest([0xFF, 0xFF], Math.pow(2, 16) - 1);
+    readUnsignedIntTest([0xFF, 0xFF, 0xFF], Math.pow(2, 24) - 1);
+    readUnsignedIntTest([0x7F, 0xFF, 0xFF, 0xFF], Math.pow(2, 31) - 1);
+
+    // Expected to fail
```
raganhan: Add tests for overflow

src/IonParserBinaryRaw.ts
zslayton: I've gone ahead and turned it into an thrown `Error`.

src/IonParserBinaryRaw.ts
zslayton: In extreme cases, yes. I've modified the check to also fail if the number of bytes read was >4, accounting for all overflow cases beyond 32 bits.

tests/unit/IonParserBinaryRawTest.js
zslayton: I borrowed this from the other unit tests, but I like your approach much better. Done.

tests/unit/IonParserBinaryRawTest.js
zslayton: The tests immediately following check for overflow.

Issue #212, opened 2019-07-05
Unit tests for reading binary Int


zslayton:
Related to #205.

Add unit tests for reading [signed binary Int](http://amzn.github.io/ion-docs/docs/binary.html#uint-and-int-fields).

zslayton:
Closed by #222.

Issue #205, opened 2019-06-25
Unit test meta-issue


zslayton:
While the event stream tests are pretty exhaustive, `ion-js`'s unit test coverage is patchy at best. When event stream test failures occur, we don't always have an obvious set of tests exercising the low-level functionality involved.

**Legend**
* ❌ No unit tests
* ⚠️ Unit tests present but obviously insufficient OR unit tests not present, but the functionality is indirectly tested elsewhere
* ✅ Unit tests offering coverage of at least basic functionality

|Type| Binary Deps | Text read | Text Write | Binary Read | Binary Write |
|---|:---:|:---:|:---:|:---:|:---:|
|`UInt`|-|-|-|❌|✅|
|`Int`|-|-|-|❌|✅|
|`VarUInt`|-|-|-|❌|✅|
|`VarInt`|-|-|-|❌|✅|
|`null`|-|✅|✅|❌|✅|
|`NOP pad`|`VarUInt`|-|-|❌|❌|
|`boolean`|-|⚠️|✅|❌|✅|
|`integer`|`VarUInt`<br/>`UInt`|❌|✅|❌|✅|
|32-bit `float`|-|❌|✅|❌|✅|
|64-bit `float`|-|❌|❌|❌|✅|
|`decimal`|`VarUInt`<br/>`VarInt`<br/>`Int`|✅|✅|❌|✅|
|`timestamp`|`VarUInt`<br/>`VarInt`<br/>`Int`|⚠️|✅|❌|✅|
|`symbol`|`VarUInt`<br/>`UInt`|❌|⚠️|❌|✅|
|`string`|`VarUInt`|⚠️|⚠️|❌|✅|
|`clob`|`VarUInt`|❌|✅|❌|✅|
|`blob`|`VarUInt`|❌|✅|❌|✅|
|`list`|`VarUInt`|❌|✅|❌|✅|
|`s-expression`|`VarUInt`|❌|✅|❌|⚠️|
|`struct`|`VarUInt`|✅|✅|❌|✅|
|`annotation`|`VarUInt`|❌|⚠️|❌|✅|

We'll create more focused sub-issues for adding or fixing tests in each functional area.

PR #222, opened 2019-07-09
Unit tests: reading binary Int

### LINKED ISSUE #213: ParserBinaryRaw::readLongInt reads wrong bit as sign
`ParserBinaryRaw` uses a bitwise `&` to extract the sign bit from the beginning of signed integers. However, [it's using `0x10`](https://github.com/amzn/ion-js/blob/development/src/IonParserBinaryRaw.ts#L211) instead of `0x80`, so it's pulling the 4th bit instead of the first.

Discovered as part of work on #205 and #212.

### LINKED ISSUE #214: ParserBinaryRaw::readLongInt creates view of incorrect byte sequence
When `ParserBinaryRaw` narrows the `BinarySpan` to a `Uint8Array` view of the bytes needed to read a signed integer, [it requests `position` bytes](https://github.com/amzn/ion-js/blob/f631889ab555306731e572fe7cbbb024cb7f4477/src/IonParserBinaryRaw.ts#L214) rather than `len` bytes, leading to unpredictable behavior.

Discovered as part of work on #205 and #212.

### LINKED ISSUE #217: Reading a signed integer modifies the input bytes being read
When reading a signed integer, `ParserBinaryRaw::readLongInt` [overwrites a byte in the input stream](https://github.com/amzn/ion-js/blob/development/src/IonParserBinaryRaw.ts#L216). This is done to simplify the process of calling `LongInt.fromIntBytes`, however:
* It's destructive--the same Ion bytes cannot be read twice with an `IonReader`.
* `LongInt.fromIntBytes` expects an `Array` of `Number`s, so our `Uint8Array` [has to be cloned anyway](https://github.com/amzn/ion-js/blob/development/src/IonParserBinaryRaw.ts#L217). We might as well modify the first `Number` in the new version of the data.

Related to #205 and #212.

zslayton:
*Issue #, if available:* #205, #212, #213, #214, #217 

*Description of changes:*

* Adds unit tests for reading binary Ion signed `Int` values. (#212)
* Fixes #213. 
* Fixes #214.
* Fixes #217.
* Includes some additional minor refactoring described in review comments.

### DIFF
--- a/src/IonParserBinaryRaw.ts
+++ b/src/IonParserBinaryRaw.ts
@@ -155,7 +155,7 @@ export class ParserBinaryRaw {
         }
     }
 
-    static readVarUnsignedFrom(input: BinarySpan) : number {
+    private static readVarUnsignedFrom(input: BinarySpan) : number {
         let numberOfBits = 0;
         let byte;
         let magnitude = 0;
@@ -209,7 +209,7 @@ export class ParserBinaryRaw {
         return LongInt.fromVarIntBytes(bytes, isNegative);
     }
 
-    static readSignedIntFrom(input: BinarySpan, numberOfBytes: number) : LongInt {
+    private static readSignedIntFrom(input: BinarySpan, numberOfBytes: number) : LongInt {
         if (numberOfBytes == 0) {
             return new LongInt(0);
         }
@@ -224,7 +224,7 @@ export class ParserBinaryRaw {
         return ParserBinaryRaw.readSignedIntFrom(this._in, this._len);
     }
 
-    static readUnsignedIntFrom(input: BinarySpan, numberOfBytes: number) : number {
+    private static readUnsignedIntFrom(input: BinarySpan, numberOfBytes: number) : number {
         let value = 0, bytesRead = 0, byte;
         if (numberOfBytes < 1)
             return 0;
@@ -254,7 +254,7 @@ export class ParserBinaryRaw {
         return ParserBinaryRaw.readUnsignedIntFrom(this._in, this._len);
     }
 
-    static readUnsignedLongIntFrom(input: BinarySpan, numberOfBytes: number) : LongInt {
+    private static readUnsignedLongIntFrom(input: BinarySpan, numberOfBytes: number) : LongInt {
         return LongInt.fromUIntBytes(Array.prototype.slice.call(input.view(numberOfBytes)));
     }
 
--- a/src/IonParserBinaryRaw.ts
+++ b/src/IonParserBinaryRaw.ts
@@ -209,15 +209,22 @@ export class ParserBinaryRaw {
         return LongInt.fromVarIntBytes(bytes, isNegative);
     }
 
-    private readLongInt() : LongInt {
-        if (this._len - this._in.position() < 1) return new LongInt(0);
-        let bytes : Uint8Array = this._in.view(this._in.position(), );
-        let sign = (bytes[0] & 0x10) != 0;
-        bytes[0] &= 0x7F;
-        return LongInt.fromIntBytes(Array.prototype.slice.call(bytes), sign);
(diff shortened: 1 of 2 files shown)


src/IonParserBinaryRaw.ts
```diff
@@ -209,15 +209,22 @@ export class ParserBinaryRaw {
         return LongInt.fromVarIntBytes(bytes, isNegative);
     }
 
-    private readLongInt() : LongInt {
-        if (this._len - this._in.position() < 1) return new LongInt(0);
-        let bytes : Uint8Array = this._in.view(this._in.position(), );
-        let sign = (bytes[0] & 0x10) != 0;
-        bytes[0] &= 0x7F;
-        return LongInt.fromIntBytes(Array.prototype.slice.call(bytes), sign);
+    static readSignedIntFrom(input: BinarySpan, numberOfBytes: number) : LongInt {
```
zslayton: I've renamed this method from `readLongInt` to `readSignedInt` because the reader currently does distinguish between `Int`s that could fit in a standard javascript `Number` value and those which require a `BigInt` (`LongInt` in this library). All `Int` values (which only occur in `Timestamp`s and `Decimal` values) are read into `LongInt`s. Also, the method's signature indicates that it always returns a `LongInt`. 

I've also extracted the meat of this method out into a `static` method that is more easily unit tested.

src/IonParserBinaryRaw.ts
```diff
@@ -209,15 +209,22 @@ export class ParserBinaryRaw {
...
+        if (numberOfBytes == 0) {
+            return new LongInt(0);
+        }
+        let bytes: Uint8Array = input.view(numberOfBytes);
+        let isNegative = (bytes[0] & 0x80) == 0x80;
```
zslayton: This line has been changed to address issue #213.

src/IonParserBinaryRaw.ts
```diff
@@ -209,15 +209,22 @@ export class ParserBinaryRaw {
...
-        if (this._len - this._in.position() < 1) return new LongInt(0);
-        let bytes : Uint8Array = this._in.view(this._in.position(), );
-        let sign = (bytes[0] & 0x10) != 0;
-        bytes[0] &= 0x7F;
-        return LongInt.fromIntBytes(Array.prototype.slice.call(bytes), sign);
+    static readSignedIntFrom(input: BinarySpan, numberOfBytes: number) : LongInt {
+        if (numberOfBytes == 0) {
+            return new LongInt(0);
+        }
+        let bytes: Uint8Array = input.view(numberOfBytes);
```
zslayton: This line has been changed to address issue #214.

src/IonParserBinaryRaw.ts
```diff
@@ -209,15 +209,22 @@ export class ParserBinaryRaw {
...
+        let isNegative = (bytes[0] & 0x80) == 0x80;
+        let numbers = Array.prototype.slice.call(bytes);
```
zslayton: This section has been modified to address #217.

src/IonParserBinaryRaw.ts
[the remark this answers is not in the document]
zslayton: I removed the `private` keyword because I assumed that would prevent that method from being used in the unit tests, but apparently it doesn't. I'll restore it.

wesboyt: Perfect, yes were only going support ints over 32 bits in decimals and timestamps. Eventually we could build a number tower but it would require altering the api contracts.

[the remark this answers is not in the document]
wesboyt: yea nothings really private =(

therapon:
I am assuming travis failures are due to bugs pending more PRs so ignoring them. If I am mistaken let me know.

Issue #231, opened 2019-07-10
Unit tests: reading VarUInt ant VarInt


zslayton:
Related to #205.

Add unit tests for reading [`VarUInt` and `VarInt` fields](http://amzn.github.io/ion-docs/docs/binary.html#varuint-and-varint-fields).

zslayton:
Addressed in #232.

PR #232, opened 2019-07-12
Unit tests: VarUint and VarInt.

zslayton:
*Issue #, if available:* #205 and #231 

*Description of changes:*
Adds unit tests for reading [VarInt and VarUInt](http://amzn.github.io/ion-docs/docs/binary.html#varuint-and-varint-fields).

### DIFF
--- a/tests/unit/IonParserBinaryRawTest.js
+++ b/tests/unit/IonParserBinaryRawTest.js
@@ -182,6 +182,120 @@ define([
     readSignedInt([0x7F, 0xFF, 0xFF, 0xFF, 0xFF], maxValueForBits(39));
     readSignedInt([0xFF, 0xFF, 0xFF, 0xFF, 0xFF], -1 * maxValueForBits(39));
 
+    /**
+     * VarUInt
+     *
+     * Spec: http://amzn.github.io/ion-docs/docs/binary.html#varuint-and-varint-fields
+     */
+
+    // The base test function called by the more specific flavors below.
+    // Creates a test which will attempt to read the `expected` VarUInt from the provided input bytes, handling
+    // any anticipated exceptions.
+    let readVarUnsignedIntTest = function(testName, bytes, expected, throwsException) {
+      let test = function() {
+        let binarySpan = new ion.BinarySpan(new Uint8Array(bytes));
+        let actual = ion.ParserBinaryRaw.readVarUnsignedIntFrom(binarySpan);
+        assert.equal(actual, expected)
+      }
+      registerTest(testName, test, throwsException);
+    }
+
+    // Should read the `expected` VarUInt value from the provided input bytes.
+    // Will fail if an exception is thrown or if the value that's read from the bytes is not equal to `expected`.
+    let readVarUnsignedInt = function(bytes, expected) {
+      let testName = 'Read var unsigned int ' + expected + ' from bytes: ' + bytes;
+      let testThrows = false;
+      readVarUnsignedIntTest(testName, bytes, expected, testThrows);
+    }
+
+    // Should detect that overflow has occurred while reading and throw an exception.
+    // Will fail if no exception is thrown.
+    let overflowWhileReadingVarUnsignedInt = function(bytes, expected) {
+      let testName = 'Overflow while attempting to read var unsigned int ' + expected + ' from the bytes: ' + bytes;
+      let testThrows = true;
+      readVarUnsignedIntTest(testName, bytes, expected, testThrows);
+    }
+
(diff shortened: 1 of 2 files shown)


tests/unit/IonParserBinaryRawTest.js
```diff
@@ -182,6 +182,120 @@ define([
...
+        readVarUnsignedInt([byte | 0x80], byte);
+    }
+
+    readVarUnsignedInt([0x80], 0);
+    readVarUnsignedInt([0x00, 0x81], 1);
+    readVarUnsignedInt([0x00, 0x00, 0x81], 1);
+    readVarUnsignedInt([0x00, 0x00, 0x00, 0x81], 1);
+
+    readVarUnsignedInt([0x0E, 0xEB], 1899);
+    readVarUnsignedInt([0x0E, 0xEC], 1900);
```
tgregg: Should we add one for `maxValueForBits(31)`?

tests/unit/IonParserBinaryRawTest.js
```diff
@@ -182,6 +182,120 @@ define([
...
+      let testThrows = false;
+      readVarUnsignedIntTest(testName, bytes, expected, testThrows);
+    }
+
+    // Should detect that overflow has occurred while reading and throw an exception.
+    // Will fail if no exception is thrown.
+    let overflowWhileReadingVarUnsignedInt = function(bytes, expected) {
+      let testName = 'Overflow while attempting to read var unsigned int ' + expected + ' from the bytes: ' + bytes;
+      let testThrows = true;
+      readVarUnsignedIntTest(testName, bytes, expected, testThrows);
```
tgregg: Is overflow the only reason that can cause the call to throw?

tests/unit/IonParserBinaryRawTest.js
```diff
@@ -182,6 +182,120 @@ define([
...
+      readVarSignedInt([byte | 0x80], byte);
+    }
+
+    // Test all negative values that can be encoded in a single byte
+    for (let byte = 0; byte <= 0x3F; byte++) {
+      readVarSignedInt([byte | 0xC0], -byte);
+    }
+
+    readVarSignedInt([0x80], 0);
+    readVarSignedInt([0xC0], -0); // Not forbidden by the spec
```
tgregg: It's not just not forbidden -- it's allowed.

tests/unit/IonParserBinaryRawTest.js
```diff
@@ -182,6 +182,120 @@ define([
...
+
+    readVarSignedInt([0x0E, 0xEB], 1899);
+    readVarSignedInt([0x4E, 0xEB], -1899);
+    readVarSignedInt([0x0E, 0xEC], 1900);
+    readVarSignedInt([0x4E, 0xEC], -1900);
+
+    readVarSignedInt([0x00, 0x00, 0x81], 1);
+    readVarSignedInt([0x00, 0x00, 0x00, 0x81], 1);
+    readVarSignedInt([0x3F, 0x7F, 0x7F, 0xFF], maxValueForBits(27));
+    readVarSignedInt([0x7F, 0x7F, 0x7F, 0xFF], -maxValueForBits(27));
```
tgregg: Should we add one for `maxValueForBits(32)`?

tests/unit/IonParserBinaryRawTest.js
zslayton: This would require more complex bit-counting logic.

`maxValueForBits(31)` requires 5 bytes to encode. At 7 usable bits per byte, it's counted as 35 bits even though the 3 most significant bits are 'unused'.

tests/unit/IonParserBinaryRawTest.js
zslayton: Same issue here RE: our approach to bit counting. We'd need to keep track of the number of 'unused' bits following the most significant `1` we've encountered, which seems like a lot of overhead.

tests/unit/IonParserBinaryRawTest.js
zslayton: ha, oops. This comment was a note-to-self I made before I found evidence that it was truly allowed. While -0 is forbidden for user-level integers, the spec didn't explicitly forbid or allow it for the `VarInt` primitive. As you pointed out, it's used later on in the spec in an example. I'll remove the comment.

tests/unit/IonParserBinaryRawTest.js
zslayton: As written, yes.

Issue #233, opened 2019-07-12
Unit tests: reading binary Decimal


zslayton:
Related to #205.

zslayton:
Addressed by #236.

PR #236, opened 2019-07-17
Unit tests: reading 10n decimal

### LINKED ISSUE #235: Reading 10n decimals doesn't use _len correctly
Found while working on #205 and #233.

Reading binary Ion decimal values is currently broken because `readSignedInt` expects the `_len` field to contain the byte length of the value being read. However, in [the case of the `read_decimal_value()` function](https://github.com/amzn/ion-js/blob/development/src/IonParserBinaryRaw.ts#L282), `readSignedInt` is called when the cursor has already been moved forward within a `Decimal` value's bytes, so only a few bytes (i.e. fewer than `_len` bytes) remains.

zslayton:
*Issue #, if available:* #205, #233, #235.

*Description of changes:*

* Fixes #235.
* Modifies the `Decimal` constructor to allow the `coefficient` to be provided as either a `LongInt` or an integral `number` value.
* Renames the private variable `_value` in the `Decimal` class to the more precise `_coefficient`.
* Adds unit tests for reading binary Ion values.

### DIFF
--- a/src/IonDecimal.ts
+++ b/src/IonDecimal.ts
@@ -43,18 +43,32 @@ import { LongInt } from "./IonLongInt";
 
 
 export class Decimal {
+
+    private _coefficient: LongInt;
+    private _exponent: number;
+
     public static readonly NULL: Decimal = new Decimal(undefined, undefined);
     public static readonly ZERO: Decimal = new Decimal(new LongInt(0), 0);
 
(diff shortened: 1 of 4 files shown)


zslayton:
What changes will it require? The user-facing APIs didn't change, just the implementation of `read_decimal_value()`.

PR #267, opened 2019-08-01
adds binary reader support for annotations

pbcornell:
Removes good/testfile9.ion from the eventstream skiplist.


src/IonBinaryReader.ts
```diff
@@ -138,16 +141,23 @@ export class BinaryReader implements Reader {
...
+      });
+    }
+  }
+
+  annotations() : string[] {
+    this._loadAnnotations();
+    return this._annotations;
   }
 
   getAnnotation(index: number) : string {
```
zslayton: I'm surprised this method exists since `annotations()` above provides direct access to the array itself.

PR #277, opened 2019-08-02
avoids use of bitwise operator in writeUnsignedInt, which treats operands as 32-bit values

### LINKED ISSUE #162: BinaryReader ints
78187493520 turns into 878082192

pbcornell:
Resolves #162, but isn't sufficient to remove good/equivs/ints.ion and good/non-equivs/ints.ion from the skiplist due to those files containing larger ints -- addressing those will require additional effort.

### DIFF
--- a/src/IonLowLevelBinaryWriter.ts
+++ b/src/IonLowLevelBinaryWriter.ts
@@ -54,8 +54,10 @@ export class LowLevelBinaryWriter {
     let i: number = tempBuf.length;
 
     while (value > 0) {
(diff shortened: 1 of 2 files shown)


pbcornell:
> Is it worth switching based on int size?

I'd say the additional complexity isn't worth it unless it improves performance in some material way;  in the absence of performance data, no, not worth switching based on int size.

pbcornell:
> Lets todo a perf comparison/correct bit shifting when size allows.

Created #280 to capture that idea.

Issue #289, opened 2019-08-12
ion-tests/iontestdata/good/decimal_zeros.ion (0d-0)


pbcornell:
ReaderCompare test fails on the seventh value in this file:  `0d-0`

PR #291, opened 2019-08-12
clarifies IonReader contract and adds ReaderCompare tests

pbcornell:
With these changes, IonReader implementations obey these rules:
1.  methods never return `undefined`
2.  *value() methods return `null` for the Ion value `null`
3.  *value() methods throw if a typed null doesn’t align with the type of the method (e.g., booleanValue() returns `null` for `null.bool`, but throws for `null.string`)
4.  *value() methods throw if a typed value doesn’t align with the type of the method (e.g., booleanValue() throws for `5` and `"hi"`)

Prior behavior of `stringValue()` has been renamed to `_stringRepresentation()`; prior behavior of `numberValue()` has been renamed to `_gitSid()`.

New skiplist entry (good/decimal_zeros.ion) is tracked in #289 


tests/unit/iontests.js
```diff
@@ -59,24 +64,130 @@ define([
...
+            allValueMethods.forEach(methodName => {
+                let typeName = type == null ? 'null' : type.name;
+                if (type != null && nonThrowingMethods[typeName][methodName]) {
+                    let v1 = r1[methodName]();
+                    let v2 = r2[methodName]();
+                    assert(v1 !== undefined, "unexpected 'undefined' response");
+                    assert(v2 !== undefined, "unexpected 'undefined' response");
+                    assert.deepEqual(v1, v2, methodName + '():  ' + v1 + ' != ' + v2);
+                } else {
+                    assert.throws(() => { r1[methodName]() }, '', '', 'Expected ' + methodName + '() to throw');
```
dlurton: How do you know this isn't throwing for some reason other than the one you intend?

pbcornell: We don't at this time; we could check for specific error messages, but that would be brittle and unwarranted, imo.  Based on a quick search, JavaScript support for custom errors appears to be minimal, and I haven't seen any code that does so.

Perhaps TypeScript includes capabilities that would be of help here.  I've opened #292 to explore this further.

src/IonTextReader.ts
```diff
@@ -246,79 +246,113 @@ next() {
...
+        throw new Error('Current value is not a decimal.')
     }
-  }
 
-  booleanValue() {
-    if (this._type !== IonTypes.BOOL) {
-      return undefined;
+    numberValue() {
+        if (this._type.num || this._type === IonTypes.NULL) {
+            return this._parser.numberValue();
```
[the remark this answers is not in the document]
pbcornell: Good catch--thank you.  Updated this `numberValue()` impl to follow the pattern of the other *value() methods.

Issue #226, opened 2019-07-09
TextReader.next:  "Error:  invalid character 14 in string"


pbcornell:
Encountered when parsing the symbol `'\x0b\x0c\x0e'`:

"Error: invalid character 14 in string"
    at ParserTextRaw._read_string_helper  </Volumes/Unix/dev/ion/ion-hash-js/ion-bundle.js:4519:76>
    at ParserTextRaw._read_string1  </Volumes/Unix/dev/ion/ion-hash-js/ion-bundle.js:5072:17>
    at ParserTextRaw._read_value_helper_single  </Volumes/Unix/dev/ion/ion-hash-js/ion-bundle.js:4803:19>
    at ParserTextRaw._read_value_helper  </Volumes/Unix/dev/ion/ion-hash-js/ion-bundle.js:4726:23>
    at ParserTextRaw._read_value  </Volumes/Unix/dev/ion/ion-hash-js/ion-bundle.js:4710:17>
    at ParserTextRaw._run  </Volumes/Unix/dev/ion/ion-hash-js/ion-bundle.js:5650:19>
    at ParserTextRaw.next  </Volumes/Unix/dev/ion/ion-hash-js/ion-bundle.js:5630:21>
    at TextReader.next  </Volumes/Unix/dev/ion/ion-hash-js/ion-bundle.js:7139:35>

Same behavior if the symbol is being parsed as an annotation or fieldname.

pbcornell:
The following code:
```javascript
let r = ion.makeReader("'\x0b\x0c\x0e'");
let t = r.next();
console.log(t.name);
console.log(r.value());
```
produces:
```
Error: invalid character 14 in string
    at ParserTextRaw._read_string_helper (dist/commonjs/es6/IonParserTextRaw.js:234:27)
    at ParserTextRaw._read_string1 (dist/commonjs/es6/IonParserTextRaw.js:757:14)
    at ParserTextRaw._read_value_helper_single (dist/commonjs/es6/IonParserTextRaw.js:494:16)
    at ParserTextRaw._read_value_helper (dist/commonjs/es6/IonParserTextRaw.js:424:20)
    at ParserTextRaw._read_value (dist/commonjs/es6/IonParserTextRaw.js:411:14)
    at ParserTextRaw._run (dist/commonjs/es6/IonParserTextRaw.js:1297:16)
    at ParserTextRaw.next (dist/commonjs/es6/IonParserTextRaw.js:1278:18)
[...]

PR #293, opened 2019-08-13
Allow escaped, non-displayable ascii characters

zslayton:
*Issue #, if available:* #226 

*Description of changes:*

Allows ASCII characters with a numeric value < 32 to be included in `string`, `symbol`, and `clob` values as long as they were escaped in the text source.

e.g. allow the escaped hex sequence `\x0b`, but not the literal byte `11`.


zslayton:
This changeset is not required. See #226 for details. Closing.

PR #296, opened 2019-08-14
Int fix

wesboyt:
*Issue #, if available:*

*Description of changes:*
Throws on ints over 4 bytes.

Needs another patch for 4-byte encodings that rely on the 32nd bit.

### DIFF
--- a/src/IonParserBinaryRaw.ts
+++ b/src/IonParserBinaryRaw.ts
@@ -464,7 +464,7 @@ export class ParserBinaryRaw {
                 } else if (this._len <= 4) {//32 bits of representation after shifts
                     this._curr = this.readUnsignedInt();
                 } else {
-                    this._curr = this.readUnsignedLongInt();
+                    throw new Error("Cant handle over 32 bits of int precision.")//this._curr = this.readUnsignedLongInt();
                 }
                 break;
             case IonBinary.TB_NEG_INT:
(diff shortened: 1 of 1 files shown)


src/IonParserBinaryRaw.ts
```diff
@@ -464,7 +464,7 @@ export class ParserBinaryRaw {
                 } else if (this._len <= 4) {//32 bits of representation after shifts
                     this._curr = this.readUnsignedInt();
                 } else {
-                    this._curr = this.readUnsignedLongInt();
+                    throw new Error("Cant handle over 32 bits of int precision.");
```
zslayton: This error message is a bit terse--it would require the (perhaps non-technical) user to read through the associated stack trace for more context. 

There's already a code path for detecting oversized ints in [`ParserBinaryRaw#readUnsignedInt()`](https://github.com/amzn/ion-js/blob/development/src/IonParserBinaryRaw.ts#L254-L259) -- should we just update the error message there and remove the `if` check here on `this._len`?

zslayton:
> Needs another patch for 5 bytes with under 3 bits of value on the 1st byte.

I'm not sure I follow. This code path is for reading `UInt`, not a `VarUInt`. 

There will still need to be a check for [4-byte encodings that rely on the 32nd bit](https://github.com/amzn/ion-js/issues/210), however.

src/IonParserBinaryRaw.ts
wesboyt: On this branch specifically were returning longint to apis expecting int so there might be other paths that reach this point. Supporting more bytes here will really require a number tower.

wesboyt:
asked zack to incorporate it into https://github.com/amzn/ion-js/pull/355 to avoid rebasing.

PR #301, opened 2019-08-15
Decimal negative zero coefficient/exponent fixes

### LINKED ISSUE #289: ion-tests/iontestdata/good/decimal_zeros.ion (0d-0)

pbcornell:
Addresses the following issues related to Decimal coefficients and exponents with value -0:
- third-party `BigInteger.sign` is supposed to be `true` for negative numbers, but sign is `false` for `BigInteger(-0)` and `BigInteger('-0')`;  constructing `BigInteger(0)` and calling `negate()` achieves the desired behavior
- `BinaryWriter.writeDecimal()` was producing an unnecessary byte when the coefficient is (+)0
- `BinaryWriter.writeDecimal()` special case for `0d0` was also (incorrectly) catching decimals with a 0 coefficient but _non-zero_ exponent 
- `Decimal.stringValue()` wasn't including `-` for negative coefficients and exponents

Resolves #289 

### DIFF
--- a/src/IonBinaryWriter.ts
+++ b/src/IonBinaryWriter.ts
@@ -117,11 +117,14 @@ export class BinaryWriter implements Writer {
       return;
     }
 
-    let coefficientBytes: Uint8Array = value.getDigits().intBytes();
+    let coefficient: LongInt = value.getDigits();
+    let writeCoefficient = !(coefficient.isZero() && coefficient.signum() === 1);  // no need to write a coefficient of 0
+    let coefficientBytes: Uint8Array | null = writeCoefficient ? coefficient.intBytes() : null;
 
-    let writer: LowLevelBinaryWriter = new LowLevelBinaryWriter(new Writeable(LowLevelBinaryWriter.getVariableLengthSignedIntSize(exponent) + coefficientBytes.length));
+    let bufLen = LowLevelBinaryWriter.getVariableLengthSignedIntSize(exponent) + (writeCoefficient ? coefficientBytes.length : 0);
+    let writer: LowLevelBinaryWriter = new LowLevelBinaryWriter(new Writeable(bufLen));
     writer.writeVariableLengthSignedInt(exponent);
-    if (!(coefficientBytes.length === 1 && coefficientBytes[0] === 0)) {   // no need to write the coefficient if it is 0
+    if (writeCoefficient) {
       writer.writeBytes(coefficientBytes);
     }
     this.addNode(new BytesNode(this.writer, this.getCurrentContainer(), TypeCodes.DECIMAL, this.encodeAnnotations(annotations), writer.getBytes()));
--- a/src/IonBinaryWriter.ts
+++ b/src/IonBinaryWriter.ts
@@ -109,18 +109,20 @@ export class BinaryWriter implements Writer {
     if (typeof value == 'string') value = Decimal.parse(value);
 
     let isPositiveZero: boolean = value.isZero() && !value.isNegative();
-    if (isPositiveZero) {
-      // Special case per the spec: http://amzn.github.io/ion-docs/binary.html#decimal
+    let exponent: number = value.getExponent();
+    if (isPositiveZero && exponent === 0 && !((1 / exponent) === -Infinity)) {
+      // Special case per the spec: http://amzn.github.io/ion-docs/docs/binary.html#5-decimal
       this.addNode(new BytesNode(this.writer, this.getCurrentContainer(), TypeCodes.DECIMAL, this.encodeAnnotations(annotations), new Uint8Array(0)));
       return;
     }
 
-    let exponent: number = value.getExponent();
-    let digits: Uint8Array = value.getDigits().intBytes();
+    let coefficientBytes: Uint8Array = value.getDigits().intBytes();
 
-    let writer: LowLevelBinaryWriter = new LowLevelBinaryWriter(new Writeable(LowLevelBinaryWriter.getVariableLengthSignedIntSize(exponent) + digits.length));
+    let writer: LowLevelBinaryWriter = new LowLevelBinaryWriter(new Writeable(LowLevelBinaryWriter.getVariableLengthSignedIntSize(exponent) + coefficientBytes.length));
     writer.writeVariableLengthSignedInt(exponent);
-    writer.writeBytes(digits);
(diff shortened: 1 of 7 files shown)


src/util.ts
```diff
@@ -0,0 +1,18 @@
...
+ * A copy of the License is located at:
+ *
+ *     http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ */
+
+export function _sign(x: number): number {
```
zslayton: * I'd like to use a more informative name for this module than `util`, in part to avoid it becoming a catch-all file that accumulates unrelated functions.
* My understanding of Typescript modules/exports/visibility is pretty weak. Can we make this package-visible by using `export` (as you've done here) and then simply not exposing it from the `Ion` module that end users will interact with? If so, can we avoid the underscore prefix?

src/util.ts
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
raganhan: In the new copyright policy you only need the first year, see https://github.com/amzn/ion-java/issues/230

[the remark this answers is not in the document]
pbcornell: opened #303 to address this

src/util.ts
pbcornell: * please suggest a more informative name;  numberutil?  mathutil?  or we can watch how util evolves and refactor if it becomes a problem;  copyright date is 2012 and this appears to be the first cross-module util function....
* at the end of the day, it all becomes JavaScript, which has no protections afaik, so _everything_ is available to the caller;  and there's nothing to stop someone from importing 'util';  in the absence of such protections, i think the '_'  prefix is actually the best option.  E.g., the following TypeScript:
```typescript
import { _sign } from "./util";
_sign(5);
```
transpiles to (commonjs):
```javascript
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
const util_1 = require("./util");
util_1._sign(5);
```
or (AMD):
```javascript
define(["require", "exports", "./util"], function (require, exports, util_1) {
    "use strict";
    Object.defineProperty(exports, "__esModule", { value: true });
    util_1._sign(5);
});
```

src/IonDecimal.ts
```diff
@@ -216,21 +223,17 @@ export class Decimal {
         let exponent = 0;
         if (str === 'null' || str === 'null.decimal') return null;
         let d = str.match('[d|D]');
-        let f  = str.match('\\.');
+        let f = str.match('\\.');
```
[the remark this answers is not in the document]
pbcornell: opened #304 to address these (and other) regex usage

src/IonBinaryWriter.ts
```diff
@@ -109,18 +110,23 @@ export class BinaryWriter implements Writer {
...
 
-    let exponent: number = value.getExponent();
-    let digits: Uint8Array = value.getDigits().intBytes();
+    let coefficient: LongInt = value.getDigits();
+    let writeCoefficient = !(coefficient.isZero() && coefficient.signum() === 1);  // no need to write a coefficient of 0
+    let coefficientBytes: Uint8Array | null = writeCoefficient ? coefficient.intBytes() : null;
 
-    let writer: LowLevelBinaryWriter = new LowLevelBinaryWriter(new Writeable(LowLevelBinaryWriter.getVariableLengthSignedIntSize(exponent) + digits.length));
+    let bufLen = LowLevelBinaryWriter.getVariableLengthSignedIntSize(exponent) + (writeCoefficient ? coefficientBytes.length : 0);
+    let writer: LowLevelBinaryWriter = new LowLevelBinaryWriter(new Writeable(bufLen));
```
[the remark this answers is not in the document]
pbcornell: As-written, LowLevelBinaryWriter isn't re-usable in this (and similar) instances, but it could be.  Opened #306 to capture this idea.

src/util.ts
zslayton: Javascript allows for extending the `Number` class by adding a function directly to it's prototype. I wonder if this method should live directly on `Number`s, allowing for:
```
2_048.sign()
```
I don't know whether that's considered poor style, however.

> or we can watch how util evolves and refactor if it becomes a problem

I'm fine with this.

zslayton: > at the end of the day, it all becomes JavaScript, which has no protections afaik, so everything is available to the caller; and there's nothing to stop someone from importing 'util'; in the absence of such protections, i think the '_' prefix is actually the best option.

Using a recent build of the `development` branch, I can run:
```
var ion = require('./dist/commonjs/es6/Ion.js')
  // or
var ion = require('./dist/browser/js/ion-bundle.js')
```
Either of which gives me access to only those functions/classes that were exported from `Ion.ts`. I can't get a handle to the contents of `IonSpan` using my `ion` variable declared above, for example.

As a developer I might be able to jump through some hoops to access that, but it's not ergonomically available. I would expect that we wouldn't export the contents of `util` or the module itself, and so am wondering if we really need the underscore to convince developers not to use it.

At any rate, this doesn't need to block your PR.

pbcornell: Right, I've since learned that the notion of a module does offer some protection.  I think the question becomes whether we care whether consumers do something like:
```
var util = require('./dist/commonjs/es6/util.js')
```

pbcornell: > Javascript allows for extending the `Number` class by adding a function directly to it's prototype.

Not sure it's warranted in this case, and it raises other questions:
* if another library also defines Number.sign(), which impl wins?
* feels like our API would be creeping outside it's boundary;  sounds appropriate for something like a `toIon()` method, but I'm not sure `sign()` should be any part of this library's API

jonwilsdon:
@zslayton It is considered bad style/bad practice to extend native types:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Inheritance_and_the_prototype_chain#Bad_practice_Extension_of_native_prototypes

@pbcornell The impl that wins with Number.sign() is the last one that is defined. Change the order of the definitions and you'll change the impl of Number.sign().

Issue #180, opened 2019-05-28
Timestamp bug


wesboyt:
- illegal state - /home/travis/build/amzn/ion-js/ion-tests/iontestdata/good/timestamp/timestamp2011-02-20T19_30_59_100-08_00.10n (2ms)
- EventStream tests - Given February 29th but year 1935 is not a leap year
  at Timestamp.checkValid - /home/travis/build/amzn/ion-js/ion-tests/iontestdata/good/timestamp/leapDay.ion (10ms)

pbcornell:
These files are no longer on any skiplists:
* good/timestamp/timestamp2011-02-20T19_30_59_100-08_00.10n
* good/timestamp/leapDay.ion

Closing.

PR #312, opened 2019-08-21
Timestamp updates

wesboyt:
*Issue #, if available:*
https://github.com/amzn/ion-js/issues/180
*Description of changes:*
Revamps some of the timestamp deserialization and fixes some differences between Date() and Timestamp around 0 and 1 being valid months.


src/IonTimestamp.ts
```diff
@@ -310,154 +281,173 @@ export class Timestamp {
...
-      // now the special case for feb 29th and leap year
-      if (this.month === 2 && this.day === 29) {
-        if (!is_leapyear(this.year)) {
-          throw new Error(`Given February 29th but year ${this.year} is not a leap year`);
+    dataModelEquals(expected : Timestamp) : boolean {
+        switch (this.precision) {
+            case Precision.NULL:
+                return expected.precision === Precision.NULL;
+            case Precision.FRACTION:
+                if(!this.fraction.equals(expected.fraction)) return false;
```
dlurton: Nit:  put return statements on the following line.

src/IonTimestamp.ts
```diff
@@ -310,154 +281,173 @@ export class Timestamp {
...
-          image = to_4_digits(t.year) + "-" + image;
+
+    stringValue() : string {
+        let image: string = "";
+        let t: Timestamp = this;
+
+        switch (t.precision) {
+            default: throw { msg: "invalid value for timestamp precision", where: "IonValueSupport.timestamp.toString" };
+            case Precision.NULL:
+                return "null.timestamp";
```
dlurton: As a user, I would find this annoying.  Perhaps return a zero length string instead?   Even throwing an exception would be better IMHO.

src/IonTimestamp.ts
```diff
@@ -310,154 +281,173 @@ export class Timestamp {
...
-          image = to_4_digits(t.year) + "-" + image + "T";
-        } else {
-          image = to_4_digits(t.year) + "-" + image;
+
+    stringValue() : string {
+        let image: string = "";
+        let t: Timestamp = this;
+
+        switch (t.precision) {
+            default: throw { msg: "invalid value for timestamp precision", where: "IonValueSupport.timestamp.toString" };
```
dlurton: This throw statement is inconsistent with the other `throw` statements I've seen in this file--should this be throwing an instance of `Error` instead?

Also, the `default` switch case should be last.

src/IonTimestamp.ts
```diff
@@ -310,154 +281,173 @@ export class Timestamp {
...
-        if (t.precision === Precision.YEAR) {
-          image = to_4_digits(t.year) + "T";
-        } else if (t.precision === Precision.MONTH) {
-          image = to_4_digits(t.year) + "-" + image + "T";
-        } else {
-          image = to_4_digits(t.year) + "-" + image;
+
+    stringValue() : string {
+        let image: string = "";
+        let t: Timestamp = this;
```
dlurton: This isn't needed--just refer to `this`.

src/IonTimestamp.ts
```diff
@@ -310,154 +281,173 @@ export class Timestamp {
...
+            case Precision.FRACTION:
+                let digits = t.fraction.getDigits().toString();
+                //exp always negative because we are a magnitude.
+                let tempExp = t.fraction.getExponent();
+                if(digits === '0') tempExp++;
+                let zeros = ".";
+                while(tempExp < 0) {
+                    zeros = zeros + '0';
+                    tempExp++;
+                }
```
dlurton: Does `Decimal` not have a `.toString()` function?  This feels like converting a `Decimal` to a string should be in its own function, at least.

Or... maybe this is another case where `String.padLeft()` should be used.

src/IonTimestamp.ts
```diff
@@ -310,154 +281,173 @@ export class Timestamp {
...
+
+        if (this.offset < MIN_OFFSET || this.offset > MAX_OFFSET) {
+            throw new Error(`Offset ${this.offset} must be between ${MIN_OFFSET} and ${MAX_OFFSET} inclusive`);
         }
-      case Precision.MONTH:
-        if (this.month < MIN_MONTH || this.month > MAX_MONTH) {
-          throw new Error(`Month ${this.month} must be between ${MIN_MONTH} and ${MAX_MONTH} inclusive`);
+
+        switch (this.precision) {
+            default:
```
dlurton: `default` should be the last switch case.

Another thought:  is having an offset valid for values that have a timestamp precision that is less than `HOUR_AND_MINUTE`?  Perhaps not.

src/IonTimestamp.ts
```diff
@@ -310,154 +281,173 @@ export class Timestamp {
...
-      case Precision.YEAR:
-        if (t.precision === Precision.YEAR) {
-          image = to_4_digits(t.year) + "T";
-        } else if (t.precision === Precision.MONTH) {
-          image = to_4_digits(t.year) + "-" + image + "T";
-        } else {
-          image = to_4_digits(t.year) + "-" + image;
+
+    stringValue() : string {
+        let image: string = "";
```
dlurton: `image` this name doesn't make sense to me.  How about `str` or `strValue` or somesuch?

src/IonTimestamp.ts
```diff
@@ -310,154 +281,173 @@ export class Timestamp {
...
-        image = _to_2_digits(t.month) + (image ? "-" + image : "");
-      case Precision.YEAR:
-        if (t.precision === Precision.YEAR) {
-          image = to_4_digits(t.year) + "T";
-        } else if (t.precision === Precision.MONTH) {
-          image = to_4_digits(t.year) + "-" + image + "T";
-        } else {
-          image = to_4_digits(t.year) + "-" + image;
+
+    stringValue() : string {
```
dlurton: What's the reason you always "prepend" i.e. `image = <new part> + image`?  That just seems weird to me. 

I would expect that the cases appear in the reverse and and new parts are appended--i..e.:  `image = image + <new part>`.  I suspect that this would be simpler as well...

dlurton: A cursory search doesn't turn up any tests for this method...  It's probably getting tested along with your text writer but I'd also add a dozen or two that test only this method.

src/IonLongInt.ts
```diff
@@ -17,6 +17,7 @@ import { BigInteger } from "./BigInteger";
 import { _sign } from "./util";
 
 export class LongInt {
+    public static readonly _zero : LongInt = new LongInt(0);
```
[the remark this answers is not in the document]
wesboyt: sure

src/IonTimestamp.ts
```diff
@@ -49,188 +48,167 @@ const MAX_YEAR: number = 9999;
 const MIN_OFFSET: number = (-23 * 60) - 59;
 const MAX_OFFSET: number = (23 * 60) + 59;
 const DAYS_PER_MONTH: number[] = [
-  -1,          // months start at 1, so we fill the 0 slot
-  31, 29, 31,  // jan, feb, mar
-  30, 31, 30,  // apr, may, june
-  31, 31, 30,  // jul, aug, sep
-  31, 30, 31,  // oct, nov, dec
+    -1,          // months start at 1, so we fill the 0 slot
```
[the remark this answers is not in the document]
wesboyt: Agreed I actually removed all the intellij formmating but it came back from the dead while I was rebasing multiple times in a row. My apologies.

src/IonLongInt.ts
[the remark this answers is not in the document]
jonwilsdon: Note that `ZERO` was just renamed to `_ZERO` here: https://github.com/amzn/ion-js/pull/313/files#diff-f70143119b3c81bc171068612933a03dR22 and has been merged to development.

src/IonTimestamp.ts
```diff
@@ -310,154 +281,173 @@ export class Timestamp {
...
-      case Precision.MONTH:
-        if (this.month < MIN_MONTH || this.month > MAX_MONTH) {
-          throw new Error(`Month ${this.month} must be between ${MIN_MONTH} and ${MAX_MONTH} inclusive`);
+
+        switch (this.precision) {
+            default:
+                throw new Error(`Unknown precision ${this.precision}`);
+            case Precision.FRACTION:
+                if (this.fraction === undefined || this.fraction === null) {
+                    throw new Error("Fractional Seconds and precision in illegal state.");
```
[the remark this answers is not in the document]
wesboyt: updated

src/IonTimestamp.ts
```diff
@@ -310,154 +281,173 @@ export class Timestamp {
...
+                }
+                if (this.fraction.geq(Decimal.ONE) || this.fraction.lessThan(Decimal.ZERO)) {
+                    throw new Error("Timestamp fractional seconds must a Decimal between 0.0 and less than 1.0")
+                }
+            case Precision.SECONDS:
+                if (this.seconds === undefined || this.seconds === null) {
```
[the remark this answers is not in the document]
wesboyt: i dont think thats accurate.

PR #314, opened 2019-08-22
aligns Decimal's compareTo() and equals() contracts with those of ion-java

### LINKED ISSUE #295: Decimal Compare
Currently Decimal compare is not complying with the equivalence function used in ion java.
for instance 2.1 and 2.10 evaluate as equal while java would evaluate them as unequal.
https://github.com/amzn/ion-js/blob/development/src/IonDecimal.ts#L139
https://github.com/amzn/ion-js/blob/development/tests/unit/IonDecimalTest.js#L67

spec:
http://speleotrove.com/decimal/daops.html#refnumco

http://speleotrove.com/decimal/damisc.html#refcotot

pbcornell:
Proposed changes update Decimal's compareTo() and equals() methods to follow the behavior found in ion-java's Decimal class.  Also adds some minimal API documentation.

Resolves #295 

### DIFF
--- a/src/IonDecimal.ts
+++ b/src/IonDecimal.ts
@@ -12,7 +12,8 @@
  * language governing permissions and limitations under the License.
  */
 
-import { LongInt } from "./IonLongInt";
+import {LongInt} from "./IonLongInt";
+import {_sign} from "./util";
 
 /**
  * This class provides the additional semantics necessary for decimal values.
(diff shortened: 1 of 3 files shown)


wesboyt:
I think the comments are a huge improvement. I dont think the entire rewrite shouldve been a priority.

pbcornell:
> I think the comments are a huge improvement. I dont think the entire rewrite shouldve been a priority.

I suspect you're referring to changes to the comareTo() method.  While I was focusing on #295 ("Decimal Compare"), I felt it was important to change variable names for clarity.  Also found opportunities to utilize JS features to simplify the implementation (specifically ternary expressions and the repeat() function).  The logic itself remains as it was.

PR #315, opened 2019-08-23
adds Reader.type() method

### LINKED ISSUE #252: No support for reader.type() in ion-js currently
There is no separate method for getting the reader type currently in the ion-js. The possible ways for getting the type right now are reader.next() and reader._type (which is a private field). 
Ion-js should also support a method that returns the Ion type of the reader calling the method.

pbcornell:
Resolves #252 


tests/unit/iontests.js
```diff
@@ -150,6 +150,10 @@ define([
             assert(r1.annotations() !== undefined, 'annotations() is udnefined');
```
tgregg: Not related, but 'udnefined' caught my eye.

pbcornell: **udnefined**, *adjective*:  not defined and scrambled.  Usage:  *"I prefer my eggs udnefined"*

fixed -- thank you

PR #318, opened 2019-08-23
adds Writer.writeValues() to transfer the contents of a reader to a writer

### LINKED ISSUE #284: provide a utility function to transfer contents of a reader to a writer
IonEventStream.writeIon helps facilitate this today (as shown below), but we need to decide whether this is the _right_ API.

```javascript
let reader = ion.makeReader("{taco : time, nacho : libre}");
let writer = ion.makePrettyWriter();
let stream = new ion.IonEventStream(reader);
stream.writeIon(writer);
console.log(ion.decodeUtf8(writer.getBytes()));
```

pbcornell:
This PR removes the `writeTo()` function from tests/unit/iontests.js in favor of a new `Writer.writeValues()` method.

Resolves #284 

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -12,7 +12,60 @@
  * language governing permissions and limitations under the License.
  */
 
+import {IonTypes} from "./IonTypes";
+import {IonType} from "./IonType";
+import {Reader as IonReader} from "./IonReader";
+import {Writer as IonWriter} from "./IonWriter";
+
+/**
+ * Returns -1 if x is negative (including -0); otherwise returns 1.
+ */
 export function _sign(x: number): number {
     return (x < 0 || (x === 0 && (1 / x) === -Infinity)) ? -1 : 1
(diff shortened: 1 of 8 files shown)


src/util.ts
```diff
@@ -12,7 +12,60 @@
...
+ * of the current container.  If there's no current value then this method
+ * calls Reader.next() to get started.
+ */
+export function _writeValues(reader: IonReader, writer: IonWriter, _depth = 0): void {
+    let type: IonType = reader.type();
+    if (type === null) {
+        type = reader.next();
+    }
+    while (type !== null) {
+        if (_depth > 0) {
```
dlurton: But there is one thing:  does this mean we will not write field names if the user has manually walked the reader into a struct?

pbcornell: that's correct;  see the "_writeValues(), start within container" test

PR #319, opened 2019-08-26
Added tests for annotations and corrected writeValues signature.

dblock:
This adds a rudimentary test for reading and writing annotations. However the add part can't be right, I'm adding it to a reader. 

```js
        var data = "a::b::123";
        var reader = ion.makeReader(data);
        reader.next();
        reader.annotations().push('c');
```

What's the right way to add an annotation? How do I get to `IonValue#addTypeAnnotation`?

---

### DIFF
--- a/src/IonBinaryWriter.ts
+++ b/src/IonBinaryWriter.ts
@@ -417,8 +417,8 @@ export class BinaryWriter implements Writer {
     this.stepOut();
   }
 
-  writeValues(reader: Reader, writer: Writer): void {
-    _writeValues(reader, this);
+  writeValues(reader: Reader, depth: number = 0): void {
+    _writeValues(reader, this, depth);
   }
 }
 
--- a/src/IonBinaryWriter.ts
+++ b/src/IonBinaryWriter.ts
@@ -417,8 +417,8 @@ export class BinaryWriter implements Writer {
     this.stepOut();
   }
 
-  writeValues(reader: Reader, depth: number = 0): void {
-    _writeValues(reader, this, depth);
+  writeValues(reader: Reader): void {
+    _writeValues(reader, this);
   }
 }
 
--- a/src/IonPrettyTextWriter.ts
+++ b/src/IonPrettyTextWriter.ts
@@ -165,7 +165,7 @@ export class PrettyTextWriter extends TextWriter {
         }
     }
 
-    writeValues(reader: Reader, writer: Writer): void {
-        _writeValues(reader, this);
+    writeValues(reader: Reader, depth: number = 0): void {
+        _writeValues(reader, this, depth);
     }
 }
--- a/src/IonPrettyTextWriter.ts
+++ b/src/IonPrettyTextWriter.ts
@@ -165,7 +165,7 @@ export class PrettyTextWriter extends TextWriter {
         }
     }
 
-    writeValues(reader: Reader, depth: number = 0): void {
-        _writeValues(reader, this, depth);
+    writeValues(reader: Reader): void {
+        _writeValues(reader, this);
     }
 }
--- a/src/IonTextWriter.ts
+++ b/src/IonTextWriter.ts
@@ -370,7 +370,9 @@ export class TextWriter implements Writer {
         }
     }
 
-    writeValues(reader: Reader, depth: number = 0): void {
+    writeValues(reader: Reader): void {
+        let depth = 0;
+        if (this.currentContainer.state == State.STRUCT_FIELD) depth++;
         _writeValues(reader, this, depth);
     }
 }
--- a/src/IonTextWriter.ts
+++ b/src/IonTextWriter.ts
@@ -370,7 +370,7 @@ export class TextWriter implements Writer {
         }
     }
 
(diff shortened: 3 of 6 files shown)


dblock:
I've updated the tests not to modify the reader via `annotations()` and to explicitly `writeInt` instead. I think this could be merged as it uses the public API and legitimately shows how to create/add an annotation.

This is not ideal since you need to know the type of the annotated item being written. I tried to roll out some kind of `writeValue` that just reads the current item off `reader`, but couldn't come up with anything worth seeing. Any suggestions on how to make this better?

dblock:
Also https://github.com/amzn/ion-js/issues/338 could make this test code simpler.

dblock:
Help me fix this one?

```js      
suite['Wrap annotation'] = function() {
        let data = "{ x: 1 }";
        let reader = ion.makeReader(data);
        reader.next();
        writer.stepIn(ion.IonTypes.STRUCT);
        writer.stepIn(ion.IonTypes.STRUCT, ['a', 'b']);
        writer.writeValues(reader);
        reader.stepOut();
        writer.stepOut();
        writer.stepOut();
        assert.equal(String.fromCharCode.apply(null, writer.getBytes()), '{a::b::{x:1}}');
      }
```

```
AssertionError: expected '{a::b::{{x:1}}}' to equal '{a::b::{x:1}}'
```

I just want to wrap my existing ion with an annotation. I tried advancing the reader into the struct, but couldn't make it work.

dblock:
I managed to make my code work with this:

```js
      suite['Wrap annotation'] = function() {
        let data = "{ x: 1 }";
        let reader = ion.makeReader(data);
        writer.stepIn(ion.IonTypes.STRUCT);
        writer.stepIn(ion.IonTypes.STRUCT, ['a', 'b']);

        reader.next();
        reader.stepIn();
        reader.next();

        util._writeValue(reader, writer, 1);

        reader.stepOut();

        writer.stepOut();
        writer.stepOut();
        assert.equal(String.fromCharCode.apply(null, writer.getBytes()), '{a::b::{x:1}}');
      }
```

Should `writeValues` take `depth` at top level?

dblock:
This is ready to re-review @pbcornell, thank you for your time!

pbcornell:
A few thoughts on the 'Wrap annotation' test:
* you've identified a way to produce invalid Ion (all values within a struct must have a fieldname); thank you, tracking this in #341 
* I'd like to avoid exposing the `depth` parameter;  I think a better approach will be to ask the writer if it is currently within a struct (i.e., add `Writer.isInStruct(): boolean`)
* I've been wondering whether specifying annotations should be separated from the write*() methods--I think you've pointed out a reason to do so
* we should probably go ahead and add `writeValue(reader)` to the Writer interface

I _think_ what you're trying to do will be properly written as either:
```javascript
let data = "{ x: 1 }";
let reader = ion.makeReader(data);
reader.next();

writer.setAnnotations(['a', 'b']);
writer.writeValue(reader);
writer.close();
// output:  a::b::{x: 1}
```
or:
```javascript
let data = "{ x: 1 }";
let reader = ion.makeReader(data);
reader.next();

writer.stepIn(ion.IonTypes.STRUCT);
writer.writeFieldName('wrapped_value');
writer.setAnnotations(['a', 'b']);
writer.writeValue(reader);
writer.stepOut();
writer.close();
// output:  {wrapped_value: a::b::{x: 1}}
```

Not sure what the right annotation behavior is if the reader's value has annotations and annotations are specified by `writer.setAnnotations()`.  I'd likely defer to ion-java's IonWriter contract, which defines `addTypeAnnotations()` and `setTypeAnnotations()` methods (and similar `writeValue()`/`writeValues()` methods).

dblock:
> * I'd like to avoid exposing the `depth` parameter;  I think a better approach will be to ask the writer if it is currently within a struct (i.e., add `Writer.isInStruct(): boolean`)

OK. Thanks for the detailed feedback. I do need to solve some of my more immediate requirements, aka being able to add an annotation, so would you please help me get to a minimal PR that you would merge before we address all the problems (eg. exposing `setAnnotations`)? I am just trying to make sure we don't consume some random unmerged branch rn.

Instead of adding a new `isInStruct` to the `Writer` interface I hacked it up in https://github.com/amzn/ion-js/pull/319/commits/940d9feaf5d3410ea96d88ae6d04c994b1e0ffc8. I am going to guess you won't like it. Please suggest a cleaner way of doing this? And whatever else is a must have to get this merged?

dblock:
So, this PR actually generates invalid ion.

Screen Shot 2019-09-04 at 4 38 42 PM

It's valid without the `{`.

Screen Shot 2019-09-04 at 4 39 21 PM

https://ion.amazon.com/

dblock:
@pbcornell I believe this should be merged. Please take a look?

- Corrects `Writer.writeValues` not to take an unused `writer` parameter. This is a bug.
- Adds a unit test that exercises current annotation APIs.

dblock:
So, `util._writeValues` isn't available form the generated bundle. Which means a consumer of ion-js still can't add an annotation in this scenario. Any suggestions of what we should do next for this?

Issue #338, opened 2019-08-29
add reader.toString()


dblock:
Should we add `Reader.toString()`?

Currently I am doing this:

```js
      function readerToBytes(reader) {
        var writer = ion.makeTextWriter();
        writer.writeValues(reader);
        writer.close();
        return writer.getBytes();        
      }

      function readerToString(reader) {
        return ion.decodeUtf8(readerToBytes(reader));
      }
```

Is there a better way?

pbcornell:
That's the way to do it today.  Reader is basically an iterator, so `Reader.toString()` would seem strange to me; perhaps a helper function would make sense.

dblock:
I PRed a proposal in #339. We can discuss it there?

PR #339, opened 2019-08-29
Reader raw

dblock:
*Issue #, if available:*

#338 

*Description of changes:*

Proposing an implementation to get raw data from a reader. The naming is the hardest part, I wanted `toString`, but that doesn't make sense for a binary reader.  Also will add more tests if we can agree on the name.

---

### DIFF
--- a/src/IonBinaryReader.ts
+++ b/src/IonBinaryReader.ts
@@ -34,6 +34,10 @@ import { Reader } from "./IonReader";
 import { SharedSymbolTable } from "./IonSharedSymbolTable";
 import { BinarySpan } from "./IonSpan";
 import { Timestamp } from "./IonTimestamp";
+import { BinaryWriter } from "./IonBinaryWriter";
+import { Writeable } from "./IonWriteable";
+import { makeBinaryWriter } from "./Ion";
+import { decodeUtf8 } from "./IonUnicode";
 
 const RAW_STRING = new IonType( -1, "raw_input", true,  false, false, false );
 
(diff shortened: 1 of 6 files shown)


dblock:
Arguments for this PR vs. doing a helper function in `ion`.

A helper function would do, but as other helper functions you would have to check what kind of reader you have (binary vs. text). This leads to code [like this](https://github.com/amzn/ion-js/blob/master/src/Ion.ts#L91).

pbcornell:
Current thoughts:
* as far as i know, ion-java's `IonReader` doesn't provide such a method.  that library has been around for a while;  not sure why ion-js would be special in this regard
* `Writer.writeValues()` seems to provide the desired functionality (and more, I'd say);  the proposal seems to be providing some convenience, but that doesn't suggest to me an addition to `Reader`
* the return value from `raw()` should be effectively the same as the input passed to `makeReader()` (modulo text formatting and perhaps subtle binary encoding differences, or if the use case is to get some subset of what is being read);  any reason not to simply hang on to the original input?

Perhaps I'm not fully understanding this use case, or perhaps it's a less common use case that doesn't suggest an api addition to me.

dblock:
My use case is that I have a large ion document and I am extracting an ion subdocument from it, then sending it to some other service.

almann:
Can you please rebase this PR against `master`, thanks!

dblock:
@almann This got closed? I understand this API as proposed is not what we want. Could you please  comment on how one can achieve my scenario?

zslayton:
@dblock We recently deleted the `development` branch (after merging it into `master`) which caused all open PRs targeting `development` to be closed as a side effect. Sorry about that. I don't seem to have the ability to re-open this PR--the only way to keep it going would be to open another one targeting `master` instead.

dblock:
I reopened this as https://github.com/amzn/ion-js/pull/435.

Issue #347, opened 2019-09-03
blobs are not handling newlines correctly


wesboyt:
good/equivs/blobs.ion fails the equivs testing due to newlines being incorporated into the base64 translation when they should be removed before the parser sets its raw string representation.

wesboyt:
#349 merged

PR #349, opened 2019-09-04
removes newlines and other accepted input before setting raw when par…

wesboyt:
…sing blobs.

*Issue #, if available:*
https://github.com/amzn/ion-js/issues/347
*Description of changes:*
skips non base64 chars in the parser.

### DIFF
--- a/src/IonParserTextRaw.ts
+++ b/src/IonParserTextRaw.ts
@@ -1032,9 +1032,16 @@ export class ParserTextRaw {
             case T_TIMESTAMP:
             case T_IDENTIFIER:
             case T_OPERATOR:
-            case T_BLOB:
                 for (index = this._start; index < this._end; index++) {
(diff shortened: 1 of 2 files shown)


src/IonParserTextRaw.ts
```diff
@@ -1032,11 +1032,18 @@ export class ParserTextRaw {
...
             case T_OPERATOR:
-            case T_BLOB:
                 for (index = this._start; index < this._end; index++) {
                     s += String.fromCharCode(this._in.valueAt(index));
                 }
                 break;
+            case T_BLOB:
+                for (index = this._start; index < this._end; index++) {
+                    ch = this._in.valueAt(index);
+                    if(IonText.is_base64_char(ch)){
```
pbcornell: please add spaces around the if condition, like this:
```javascript
if (IonText.is_base64_char(ch)) {
```

Issue #310, opened 2019-08-20
remove IonLongInt from public API


pbcornell:
This class is intended to be internal, but is currently exposed by Decimal.getValue() (and maybe other methods?)

pbcornell:
#313 renamed Decimal.getValue() to _getCoefficient() (and thus "internal").  Are there other places where IonLongInt is exposed via the public API?

pbcornell:
IonDecimal's constructor allows a LongInt to be passed; perhaps we can hide this?

zslayton:
Fixed by #355.

PR #355, opened 2019-09-05
Removed LongInt from public API.

zslayton:
*Issue #, if available:* #310, #210 

*Description of changes:*
* Created a new static factory method `Decimal#_fromLongIntCoefficient` that accepts a `LongInt` coefficient instead of a `Number`.
* Reading a binary integer that requires more than 31 bits to store will throw an `Error`.
* Added 10 tests (all related to reading large ints) to the skip lists.

### DIFF
--- a/src/IonBinaryWriter.ts
+++ b/src/IonBinaryWriter.ts
@@ -130,15 +130,15 @@ export class BinaryWriter implements Writer {
 
     if (typeof value == 'string') value = Decimal.parse(value);
 
-    let isPositiveZero: boolean = value.numberValue() === 0 && !value.isNegative();
     let exponent: number = value._getExponent();
+    let coefficient: LongInt = value._getCoefficient();
+    let isPositiveZero: boolean = coefficient.isZero() && !value.isNegative();
     if (isPositiveZero && exponent === 0 && _sign(exponent) === 1) {
       // Special case per the spec: http://amzn.github.io/ion-docs/docs/binary.html#5-decimal
       this.addNode(new BytesNode(this.writer, this.getCurrentContainer(), IonTypes.DECIMAL, this.encodeAnnotations(annotations), new Uint8Array(0)));
       return;
     }
 
-    let coefficient: LongInt = value._getCoefficient();
     let writeCoefficient = !(coefficient.isZero() && coefficient.signum() === 1);  // no need to write a coefficient of 0
     let coefficientBytes: Uint8Array | null = writeCoefficient ? coefficient.intBytes() : null;
 
--- a/src/IonDecimal.ts
+++ b/src/IonDecimal.ts
@@ -23,18 +23,31 @@ export class Decimal {
     private _coefficient: LongInt;
     private _exponent: number;
 
-    public static readonly ZERO : Decimal = new Decimal(new LongInt(0), 0);
+    public static readonly ZERO : Decimal = new Decimal(0, 0);
     public static ONE = new Decimal(1, 0);
 
-    constructor(coefficient: LongInt | number, exponent: number) {
-        if (typeof coefficient === "number") {
-            if (!Number.isInteger(coefficient)) {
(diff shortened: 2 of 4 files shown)


src/IonBinaryWriter.ts
```diff
@@ -130,15 +130,15 @@ export class BinaryWriter implements Writer {
 
     if (typeof value == 'string') value = Decimal.parse(value);
 
-    let isPositiveZero: boolean = value.numberValue() === 0 && !value.isNegative();
```
zslayton: This call to `numberValue()` is [somewhat expensive](https://github.com/amzn/ion-js/blob/master/src/IonDecimal.ts#L73-L77) considering we're just checking whether the coefficient is zero, so I've replaced it with a call to `value._getCoefficient().isZero()` below.

src/IonDecimal.ts
```diff
@@ -23,18 +23,31 @@ export class Decimal {
...
+    static _fromLongIntCoefficient(coefficient: LongInt, exponent: number) {
+        let value = Object.create(this.prototype);
+        value.initialize(coefficient, exponent);
+        return value;
+    }
+
+    /**
+     * Constructor helper shared by the public constructor and _fromLongIntCoefficient
+     */
+    private initialize(coefficient: LongInt, exponent: number) {
```
pbcornell: rename to `_initialize`

tests/unit/iontests.js
```diff
@@ -344,6 +344,16 @@ let goodSkipList = toSkipList([
     'ion-tests/iontestdata/good/symbolZero.ion',
     'ion-tests/iontestdata/good/utf16.ion',
     'ion-tests/iontestdata/good/utf32.ion',
+
+    // Support for reading LongInts was removed as part of: https://github.com/amzn/ion-js/issues/310
```
pbcornell: I've been running a script to regenerate these skip lists on a weekly basis, and that process will drop comments like this.  Maybe we need to rethink that approach, or find a different way to maintain comments such as this...?

src/IonParserBinaryRaw.ts
```diff
@@ -478,31 +478,33 @@ export class ParserBinaryRaw {
         t._a = arr;
     }
 
+    /**
+     * Positive integers and negative integers are both encoded as an unsigned integer magnitude.
+     * This function will read in the magnitude, leaving it to the caller to set the value's sign as appropriate.
+     */
+    private readIntegerMagnitude(): number {
```
pbcornell: rename to `_readIntegerMagnitude()`

tests/unit/iontests.js
zslayton: It's valuable to have a documented mapping from skiplist entries to github issues. It helps contributors to know when their patch has fully addressed the (known aspects of the) problem they're tackling.

I'm all for keeping that mapping in a github meta issue if you prefer. We could add a link to the issue to the top of the script-generated skiplist.

src/IonDecimal.ts
zslayton: Done. 👍

src/IonBinaryWriter.ts
```diff
@@ -130,15 +130,15 @@ export class BinaryWriter implements Writer {
...
     let exponent: number = value._getExponent();
+    let coefficient: LongInt = value._getCoefficient();
+    let isPositiveZero: boolean = value._getCoefficient().isZero() && !value.isNegative();
```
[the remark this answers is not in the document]
zslayton: /facepalm, after going to all that trouble. Good catch!

src/IonParserBinaryRaw.ts
zslayton: Done.

tests/unit/iontests.js
pbcornell: We'll figure out something; perhaps it's time to stop auto-generating the skiplist.

PR #362, opened 2019-09-06
Cleans up build and adds tsconfig for TypeScript compiler.

### LINKED ISSUE #360: Add tsconfig.json
Currently, we encode the Typescript compilation options directly in the Grunt build.  We should have proper `tsconfig.json` files and have the Grunt files refer to them for better tooling integration *especially* with IDEs.

almann:
* Adds a vanilla `tsconfig.json` that targets native ES6 modules.
* Adds extension `tsconfig*.json` for AMD and CommonJS targets.
* Reconfigures Grunt to use `tsconfig*.json` for `ts` targets.
* Fixes copy target in to put `BigInteger.d.ts` in the right place.
* Removes seemingly redundant `amd-debug` build target.
* Adds some README around the build distribution format.

Fixes #360

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -65,33 +65,14 @@ module.exports = function(grunt) {
       }
     },
     ts: {
-      'amd-es6-debug': {
-        src: ['src/**/*.ts'],
-        outDir: 'dist/amd/es6',
-        options: {
-          target: "es6",
-          module: "amd",
-          declaration: true
-        }
+      'es6-es6': {
+        tsconfig: 'tsconfig.json'
       },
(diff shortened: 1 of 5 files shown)


Gruntfile.js
```diff
@@ -65,33 +65,14 @@ module.exports = function(grunt) {
       }
     },
     ts: {
-      'amd-es6-debug': {
```
therapon: https://github.com/amzn/ion-js/blob/master/DEBUGGER.md

I think this is a left over, we did not use to generate maps for our standard targets.

Issue #342, opened 2019-08-29
Ion.ts exports BinaryWriter, should it?


pbcornell:
Neither TextWriter nor PrettyTextWriter are exported; why is BinaryWriter special?  I'd suggest we consider not exporting it.

zslayton:
`BinaryWriter` is no longer exported as of #364.

Issue #307, opened 2019-08-19
define public API and mark the rest as internal


pbcornell:
TypeScript/JavaScript don't provide much support for declaring public vs. internal code.  Module systems (i.e., AMD or CommonJS) provide the notion of "export" (and everything else is internal), but a class is either exported or not--all methods of an exported class are exported along with the class.

The scope of this issue includes:
* reviewing the current public API and identifying what does/doesn't belong in that API
* making a best-effort to indicate (and enforce when possible) that everything not in the public API is internal.  This includes functions, classes, methods, and variables.
  * at a minimum, this involves prefixing such code with '_' to indicate _by convention_ that something is internal (for non-module environments)
  * in addition, where possible, a module system's export facility should be used to enforce the distinction between public and internal

Related to #299

pbcornell:
The following are known API issues:
- remove the Writer.writeList/writeSexp/writeStruct isNull parameter (#254)
- Writer.writeNull(TypeCode) vs. Reader.next(): IonType (#309)
- IonLongInt is intended to be internal, but is exposed by Decimal.getDigits() (#310)
- clarify Decimal.getDigits() and LongInt.digits() (#298)
- provide some notion of internal vs. external (e.g., prefix internal classes/methods/properties/functions with '_') (currently also covered by this issue; perhaps should be split to another)
- IonEventStream.ts and IonEvent.ts don't belong in the src folder -- they are only for testing (#311)
- [fixed] reader.stringValue() previously returned a string representation for non-string/symbols (e.g., the string "null.<type>" or "5" for the int 5)
- [fixed] reader.numberValue() previously returned SIDs of symbols
- [fixed] various Reader \*value\*() methods previously returned `undefined`, but now returns `null`

The following are missing from the API:
- add Reader.type() (#252)
- utility function to transfer contents of a reader to a writer (#284)

almann:
We really should consider having some first-class error type like `IonError` and have it prototype extend `Error`.  Since we're targeting ES6, this is pretty straightforward.

E.g.

```
class IonError extends Error {
  constructor(...args) {
    super(...args);
  }
}
```

These errors should be explicit, and we should support forms that allow for properties like `lineNumber` and `columnNumber` or the like.

pbcornell:
Custom error classes is captured in #292.  Other items in this issue have been addressed, with the exception of '_' prefixing all internal artifacts;  public artifacts are in the generated documentation, all others are considered internal (as indicated by the v3.0.0 release notes).  Closing.

PR #364, opened 2019-09-06
Stop exporting BinaryWriter

zslayton:
*Issue #, if available:* #342, #307

*Description of changes:*
* `Ion.ts` currently exports the `BinaryWriter` class, but the prescribed way to create a writer is using `ion.makeBinaryWriter`. We can avoid making this class part of the public API until we have a good reason to.
* Removed an unused value, `const e = {...}`, that looks like it was intended to be the basis for an `IonError`-ish type but wasn't.
* Removed the `Options` interface, which was meant to be used to configure new `Catalog` instances but wasn't.
* Renamed a parameter on the `isBinary` method that didn't align with the associated typedoc (`buf` -> `buffer`).

### DIFF
--- a/src/Ion.ts
+++ b/src/Ion.ts
@@ -25,32 +25,18 @@ import { BinaryWriter } from "./IonBinaryWriter";
 import { LocalSymbolTable, defaultLocalSymbolTable } from "./IonLocalSymbolTable";
 import { decodeUtf8 } from "./IonUnicode";
 
-const e = {
-  name: "IonError",
-  where: undefined,
-  msg: "error",
-}
-
-
 /**
- * Options object to be passed during the creation of an Ion Reader.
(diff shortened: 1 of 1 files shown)


src/Ion.ts
```diff
@@ -25,32 +25,15 @@ import { BinaryWriter } from "./IonBinaryWriter";
...
-
 /**
  * Returns the `buf` type as true for binary, false for text.
  * 
- * @param buffer we want to check its type 
+ * @param buffer we want to check its type
  * @returns either `'binary'` or `'text'`
  */
-function isBinary(buf: Uint8Array) {
+function isBinary(buffer: Uint8Array) {
```
pbcornell: let's specify the return type:  `: boolean`

zslayton: Done. I've also clarified the typedoc text, which I just noticed was pretty outdated.

PR #368, opened 2019-09-10
forces local offset to -0 for year/month/day-precision timestamps; fixes binary reader for timestamps with year < 100

### LINKED ISSUE #164: Timestamp 0001-01-01T00:00:00Z roundtrip fails
Equivalent values returned by text/binary sides do not compare equal:
0001-01-01T00:00:00Z does not equal 0001-01-01T00:00:00Z
Probably need to have timestamp, decimal etc. all refactored across the binary/text side so comparisons work correctly around type.

pbcornell:
* forces local offset to `-0` for year/month/day-precision timestamps, and updates binary writer tests to write  `c0` (-0) for such offsets to match ion-java behavior
* adds `_adjustMsSinceEpochIfNeeded` method to compensate for timestamps with year < 100 and applies it in the binary reader

Resolves #164 

### DIFF
--- a/src/IonParserBinaryRaw.ts
+++ b/src/IonParserBinaryRaw.ts
@@ -357,7 +357,9 @@ export class ParserBinaryRaw {
             fractionalSeconds = Decimal.parse(secondInt + '.' + fractionStr);
         }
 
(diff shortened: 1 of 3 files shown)


pbcornell:
Reopened as https://github.com/amzn/ion-js/pull/383 after development branch was deleted.

PR #370, opened 2019-09-10
Adds TypeDoc Configuration

### LINKED ISSUE #369: Add Doc Build Configuration to Prune Documentation to Public API
As per #365, we should also tighten the `typedoc` build to eliminate non-exported and internal APIs.

almann:
* Moves most of the doc configuration to `typedoc.json`.
* Adds various options to only publish exported/public APIs.
* Adds an exclude list for internal modules based on cursory examination
  of what can be reached from `Ion` module.

Resolves #369.

### DIFF
--- a/typedoc.json
+++ b/typedoc.json
@@ -1,0 +1,31 @@
+{
+  "name": "Ion-JS Library",
+
+  "excludeNotExported": true,
+  "excludePrivate": true,
+  "excludeProtected": true,
+
+  "exclude": [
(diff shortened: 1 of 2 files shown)


typedoc.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "Ion JS Library",
```
zslayton: Nit: should this just be `ion-js` or `Ion JS`?

almann: Done.

Issue #376, opened 2019-09-10
Clean up documentation for Decimal


zslayton:
Related to #365. Some non-public functions need to be `@hidden`, class-level docs could use some examples.

zslayton:
Largely addressed by #377.

Issue #365, opened 2019-09-09
v3.0.0 Public API documentation meta-issue


zslayton:
### Legend
* ❌ More or less undocumented
* ⚠️ Comments are present, but may be incomplete, outdated, or expose non-public information.
* ✅ Fully documented, with typedoc at the class level and on member functions
* ❎ Excluded from typedoc

### `src` Documentation
file name|documentation|issue
-|:-:|:-:
Ion.ts |✅|#373
IonCatalog.ts |✅|#375
IonDecimal.ts |✅|#376, #377 
IonLocalSymbolTable.ts |❎|#382
IonSharedSymbolTable.ts |❎|#382
IonSystemSymbolTable.ts |❎|#382
IonSymbols.ts |❎|#382
IonReader.ts |✅|#389
IonTimestamp.ts |✅|#377
IonType.ts |✅|#395
IonTypes.ts |✅|#395
IonWriter.ts |✅|#391
IonBinary.ts |❎|n/a
IonBinaryReader.ts |❎|n/a
IonBinaryWriter.ts |❎|n/a
IonConstants.ts |❎|n/a
IonEvent.ts |❎|n/a
IonEventStream.ts |❎|n/a
IonImport.ts |❎|n/a
IonLongInt.ts |❎|n/a
IonLowLevelBinaryWriter.ts |❎|n/a
IonParserBinaryRaw.ts |❎|n/a
IonParserTextRaw.ts |❎|n/a
IonPrecision.ts |❎|n/a
IonPrettyTextWriter.ts |❎|n/a
IonSpan.ts |❎|n/a
IonSubstituteSymbolTable.ts |❎|n/a
IonSymbol.ts |❎|n/a
IonSymbolIndex.ts |❎|n/a
IonTests.ts |❎|n/a
IonText.ts |❎|n/a
IonTextReader.ts |❎|n/a
IonTextWriter.ts |❎|n/a
IonUnicode.ts |❎|n/a
IonValue.ts |❎|n/a
IonWriteable.ts |❎|n/a
util.ts |❎|n/a

### Contributing

Open a new issue with the following information in the description:
* A reference to this, the meta-issue, so the two are linked.
* A list of the class files you intend to document as part of that work.

For example, to claim the `Reader` interface and `TextReader`/`BinaryReader` classes, you'd make a new issue called 'Reader Documentation' that lists the associated files.

[...]

PR #378, opened 2019-09-10
Added to docs for Decimal class/constructor.

zslayton:
*Issue #, if available:* #365, #376

*Description of changes:*

* Adds a definition of the terms `exponent` and `coefficient` to `Decimal`'s class-level docs, along with examples.
* Adds TSDoc for `Decimal`'s public constructor.

### DIFF
--- a/src/IonDecimal.ts
+++ b/src/IonDecimal.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * Copyright 2012 Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License").
  * You may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


src/IonDecimal.ts
```diff
@@ -17,7 +17,22 @@ import {_sign} from "./util";
 
 /**
  * This class provides the additional semantics necessary for decimal values.
- * Note that range of an exponent is limited to +/- 15 digits.
+ *
+ * Unlike IEEE-754 floating point representation, Decimal values can represent numeric values without loss of precision.
+ * They combine an arbitrarily large integer value (the "coefficient") with a second integer representing its scale
```
pbcornell: Most definitions of "scale" are "the number of digits to the right of the decimal point."  It's a bit different than exponent.

src/IonDecimal.ts
```diff
@@ -17,7 +17,22 @@ import {_sign} from "./util";
...
+ * (the "exponent") using the following formula:
+ *
+ *     value = coefficient * (10 ^ exponent)
+ *
+ * Here are some examples:
+ *
+ *      coefficient     exponent        value
+ *             0           0                0
+ *           314          -2                3.14
+ *           -76           1              -76
```
pbcornell: something's off in this example:  `-76d1` would be `-760` (vs. `-76d0`, which would be `-76`)

src/IonDecimal.ts
```diff
@@ -17,7 +17,22 @@ import {_sign} from "./util";
...
  * This class provides the additional semantics necessary for decimal values.
- * Note that range of an exponent is limited to +/- 15 digits.
+ *
+ * Unlike IEEE-754 floating point representation, Decimal values can represent numeric values without loss of precision.
+ * They combine an arbitrarily large integer value (the "coefficient") with a second integer representing its scale
+ * (the "exponent") using the following formula:
+ *
+ *     value = coefficient * (10 ^ exponent)
+ *
+ * Here are some examples:
```
pbcornell: maybe it would be helpful to have an example such as `0.00037` ?

src/IonDecimal.ts
zslayton: I wasn't aware that 'scale' had an official meaning in this space, but that makes sense. I'll rephrase it.

src/IonDecimal.ts
zslayton: Whoops, good catch!

src/IonDecimal.ts
zslayton: Yep, added.

zslayton:
Rebased from `master` and moved to #386

PR #381, opened 2019-09-10
Cleans up documentation for IonCatalog.

### LINKED ISSUE #375: Cleans up API documentation for IonCatalog
Child of #365

almann:
Specifically annotates returns as nullable where appropriate.

Fixes #375.

### DIFF
--- a/src/IonCatalog.ts
+++ b/src/IonCatalog.ts
@@ -21,17 +21,19 @@ import { SharedSymbolTable } from "./IonSharedSymbolTable";
     }
 
     /**
-    * A catalog holds available shared symbol tables and always includes the system symbol table.
-    * @see http://amzn.github.io/ion-docs/symbols.html#the-catalog
-    */
+     * A catalog holds available shared symbol tables and always includes the system symbol table.
+     * @see https://amzn.github.io/ion-docs/docs/symbols.html#the-catalog
+     */
     export class Catalog {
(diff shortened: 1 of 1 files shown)


src/IonCatalog.ts
```diff
@@ -21,33 +21,45 @@ import { SharedSymbolTable } from "./IonSharedSymbolTable";
     }
 
     /**
-    * A catalog holds available shared symbol tables and always includes the system symbol table.
-    * @see http://amzn.github.io/ion-docs/symbols.html#the-catalog
-    */
+     * A catalog holds available shared symbol tables and always includes the system symbol table.
+     * @see http://amzn.github.io/ion-docs/symbols.html#the-catalog
```
pbcornell: This URL redirects (and drops the anchor reference).  Please use http://amzn.github.io/ion-docs/docs/symbols.html#the-catalog instead.

almann: Umm, just to make sure I read this right, should it be http***s***?

pbcornell: either http or https works, the difference is whether the path is `ion-docs` vs `ion-docs/docs`

almann: Ah, got it, thanks--I couldn't see it, but I fixed this by copy/pasting the URL over what was there.

PR #390, opened 2019-09-10
Adds API documentation for Reader.

### LINKED ISSUE #389: Cleanup API Documentation for Reader
Child of #365.

almann:
* Makes the API contracts explicit with respect to `null`.
* Changes interface definition to use method syntax versus function
  types.

Resolves #389

### DIFF
--- a/src/IonReader.ts
+++ b/src/IonReader.ts
@@ -16,23 +16,139 @@ import { Decimal } from "./IonDecimal";
 import { IonType } from "./IonType";
 import { Timestamp } from "./IonTimestamp";
 
+/** Represents the possible return values of [[Reader.value]]. */
+export type ReaderScalarValue = null | boolean | number | Decimal | Timestamp | string | Uint8Array;
+
 /**
- * Reads a sequence of Ion values.
+ * A pull parser interface over Ion data.
+ *
+ * Generally, users will use [[next]] and [[stepIn]] and [[stepOut]]
+ * to traverse the structure of the data. The **Value** suffixed methods are
+ * used to extract the scalar data out of the current position of the [[Reader]].
+ *
+ * [[Reader]] instances start *before* a value, thus you have to
+ * invoke [[next]] to position a newly created reader on the first value.
  */
 export interface Reader {
-  booleanValue: () => boolean;
-  byteValue: () => Uint8Array;
-  decimalValue: () => Decimal;
-  depth: () => number;
-  fieldName: () => string;
-  isNull: () => boolean;
-  next: () => IonType;
-  numberValue: () => number;
-  stepIn: () => void;
-  stepOut: () => void;
-  stringValue: () => string;
-  timestampValue: () => Timestamp;
-  type: () => IonType;
-  value: () => any;
-  annotations: () => string[];//TODO implement symboltokens to replace string[] https://github.com/amzn/ion-js/issues/121
+    /**
+     * Advances the reader to the next value in the stream at the current depth.
+     *
+     * @return The corresponding [[IonType]] of the value the reader moves to, or `null`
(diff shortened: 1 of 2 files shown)


src/IonReader.ts
```diff
@@ -16,23 +16,139 @@ import { Decimal } from "./IonDecimal";
...
+     * Returns the annotations for the current value.
+     *
+     * @return the empty array if the there are no annotations or the reader is not positioned on a value.
+     */
+    annotations(): string[];
+
+    /**
+     * Returns the field name of the current value.
+     *
+     * @return `null` if the reader is not positioned on a value or is in a value that has no field name.
```
pbcornell: "or is in a value" --> "or is on a value"

src/IonReader.ts
```diff
@@ -16,23 +16,139 @@ import { Decimal } from "./IonDecimal";
...
+     */
+    fieldName(): string | null;
+
+    /** Returns true if and only if the reader is positioned on a `null` value of any type. */
```
pbcornell: so when not positioned on a value, i should expect `false`?  (just checking)

src/IonReader.ts
```diff
@@ -16,23 +16,139 @@ import { Decimal } from "./IonDecimal";
...
+     */
+    booleanValue(): boolean | null;
+
+    /**
+     * Returns the current value as a `Uint8Array`.  This is only valid if `type() == IonTypes.BLOB`
+     * or `type() == IonTypes.CLOB`.
+     *
+     * @return `null` if the current Ion value [[isNull]].
+     *
+     * @throw Error when the reader is not positioned on a `clob` or `blob` typed value.
```
pbcornell: nit, consistent ordering:  "on a `blob` or `clob` typed value"

src/IonReader.ts
```diff
@@ -16,23 +16,139 @@ import { Decimal } from "./IonDecimal";
...
+     * @return `null` if the current Ion value [[isNull]]. For `int` values that are outside of
+     *  the range specified by `Number.MIN_SAFE_INTEGER` and `Number.MAX_SAFE_INTEGER`, this method
+     *  will truncate the result.
+     *
+     * @throw Error when the reader is not positioned on a `int` or `number` typed value.
+     */
+    numberValue(): number | null;
+
+    /**
+     * Returns the current value as a [[Timestamp]].  This is only valid if `type() == IonTypes.TIMESTAMP`.
```
pbcornell: this is for `stringValue()`, but appears to be a copy of the timestamp typedoc

src/IonReader.ts
```diff
@@ -16,23 +16,139 @@ import { Decimal } from "./IonDecimal";
...
+     *
+     * @return `null` if the current Ion value [[isNull]].
+     *
+     * @throw Error when the reader is not positioned on a `timestamp` typed value.
+     */
+    timestampValue(): Timestamp | null;
+
+    /**
+     * Returns the current value as a "native" JavaScript value.  This is only valid when `type().scalar == true`.
```
pbcornell: Decimal and Timestamp probably don't qualify as `"native" JavaScript value`s.  Perhaps just "Returns the current value. ..." ?

src/IonReader.ts
almann: Yes

src/IonReader.ts
almann: Whoops--yep

src/IonReader.ts
almann: Done.

src/IonReader.ts
almann: Done.

src/IonReader.ts
almann: Fixed.

PR #394, opened 2019-09-11
Cleans up documentation and API for Writer interface.

### LINKED ISSUE #391: Cleanup API Documentation for Writer
Child of #365.

almann:
The relationship between `close()` and `getBytes()` is poor
coupling of API and should be reconsidered in the future.
To wit, since an internal buffer is implied by `Writer`, perhaps
we can model this as `finish(): Uint8Array` instead of these
two methods.

Resolves #391

### DIFF
--- a/src/IonWriter.ts
+++ b/src/IonWriter.ts
@@ -17,43 +17,138 @@ import {Reader} from "./IonReader";
 import {Timestamp} from "./IonTimestamp";
 
 /**
- * Writes values in the Ion text or binary formats.
+ * Serializes data in Ion text or binary format to an implementation defined buffer.
  */
 export interface Writer {
-  writeBlob(value: Uint8Array) : void;
-  writeBoolean(value: boolean) : void;
-  writeClob(value: Uint8Array) : void;
-  writeDecimal(value: Decimal) : void;
-  writeFieldName(fieldName: string) : void;
-  writeFloat32(value: number) : void;
-  writeFloat64(value: number) : void;
-  writeInt(value: number) : void;
-  writeNull(type: IonType) : void;
-  writeString(value: string) : void;
-  writeSymbol(value: string) : void;
-  writeTimestamp(value: Timestamp) : void;
-  stepIn(type: IonType) : void;
-  stepOut() : void;
-  getBytes(): Uint8Array;
+    /** Writes a null value of the given [[IonType]]. */
+    writeNull(type: IonType) : void;
 
-  /**
-   * Adds an annotation to the list of annotations to be used when
-   * writing the next value.
-   */
-  addAnnotation(annotation: string) : void;
+    /**
+     * Writes a `boolean` as an Ion `bool` value.
+     *
+     * @param value The `boolean` to write, which may be `null` to write a `null.bool`.
+     */
+    writeBoolean(value: boolean | null) : void;
 
-  /**
-   * Specifies the list of annotations to be used when writing
-   * the next value.
-   */
-  setAnnotations(annotations: string[]) : void;
+    /**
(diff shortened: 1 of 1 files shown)


src/IonWriter.ts
```diff
@@ -17,43 +17,138 @@ import {Reader} from "./IonReader";
...
+    /**
+     * Writes the field name for a member of a `struct`.
+     *
+     * @throws Error if the [[Writer]] is not within a `struct` or this method was already
+     *  called before a value was written.
+     */
+    writeFieldName(fieldName: string) : void;
+
+    /**
+     * Starts a container and positions the writer within that container.
```
zslayton: For `stepIn` and `stepOut`, it might be nice to mention that you still have to call `next()` before you can read the following value at whichever depth you now find yourself. We could also address this by having good examples in the `README.md`. Or both!

src/IonWriter.ts
```diff
@@ -17,43 +17,138 @@ import {Reader} from "./IonReader";
...
+     * Flushes data to the internal buffer and finalizes the Ion stream.
+     *
+     * @throws Error if the [[Writer]] is already closed or the writer is not
+     *  at the top-level (i.e. inside of a container).
+     */
+    close() : void;
+
+    /**
+     * Retrieves the serialized buffer as an array of octets.  Generally this will
+     * be either a UTF-8 encoded buffer for Ion text or Ion binary.  The buffer is
```
zslayton: > Generally this will be either a UTF-8 encoded buffer for Ion text or Ion binary.

* You say "generally" -- are there other options?
* This was a bit of a garden path sentence for me. I would rearrange it a bit for easier parsing:
   > Generally, this will be either Ion binary or a UTF-8 encoded buffer for Ion text.

   Alternatively, since we already know the return type to be `Uint8Array`:
   > [The buffer] will contain either Ion binary data or UTF-8 encoded Ion text.

almann: Good point, I did think it wasnkind of hedgy as I wrote it, let me reword.

The use case I was thinking wss an Ion writer for things that don't serialize (e.g. to a DOM) or using the interface to leverage the logical Ion types but emit something like CBOR.

The API is not amenable to the former, and the later is somewhat a stretch.

src/IonWriter.ts
```diff
@@ -17,43 +17,138 @@ import {Reader} from "./IonReader";
...
+    writeFloat64(value: number | null) : void;
+
+    /**
+     * Writes a [[Decimal]] value as an Ion `decimal` value.
+     *
+     * @param value The [[Decimal]] to write, which may be `null` to write a `null.decimal`.
+     */
+    writeDecimal(value: Decimal | null) : void;
+
+    writeTimestamp(value: Timestamp) : void;
```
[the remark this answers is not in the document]
almann: Whoops, fixed.

src/IonWriter.ts
zslayton: Whoops, sure did. Sorry about that.

src/IonWriter.ts
almann: Fixed

src/IonWriter.ts
```diff
@@ -17,43 +17,143 @@ import {Reader} from "./IonReader";
...
+     *
+     * @throws Error if the [[Writer]] is not within a `struct` or this method was already
+     *  called before a value was written.
+     */
+    writeFieldName(fieldName: string) : void;
+
+    /**
+     * Starts a container and positions the writer within that container.
+     *
+     * @throws Error if `type.container` is not one of `IonTypes.LIST`, `IonTypes.SEXP`,
```
pbcornell: i think we want `type` instead of `type.container` here

PR #397, opened 2019-09-11
Clarified names of IonType member fields.

zslayton:
*Description of changes:*

The names of the member fields in the `IonType` class were terse and did not reflect their data types. These fields constitute part of `ion-js`'s public API, and so should be as self-explanatory as possible.

This PR expands shortened names and adds an `is` prefix to `boolean` fields. I recommend skipping down to `src/IonType.ts` to see the renamings before reading through the usage changes.


src/IonType.ts
```diff
@@ -15,29 +15,29 @@
 /** Ion value type enumeration class. */
 export class IonType {
```
pbcornell: please mark all fields as `readonly`

pbcornell: if you were so inclined, this class could be written as:
```javascript
export class IonType {
    constructor(
        /** The binary type ID for this Ion Type. */
        public readonly binaryTypeId: number,
        /** The textual name of this type. */
        public readonly name: string,
        ...
    }
}
```

zslayton: Good idea, done.

zslayton: You bet I'm inclined! Done.

almann: Oh nice, I like this.

Issue #340, opened 2019-08-29
writer.writeInt(reader.numberValue(), 'a') should not work 


dblock:
Via https://github.com/amzn/ion-js/pull/319#discussion_r319165042

```
writer.writeInt(reader.numberValue(), 'a');
```

This shouldn't work at all, as the annotations parameter is supposed to be a `string[]`. Here 'a' is being treated as an array containing a single 1-character string (yay, javascript ;).

pbcornell:
As another example:  `writer.writeInt(123, 'abc')` produces `a::b::c::123`.  This is a bug for all write*() methods, as well as stepIn().  Haven't investigated whether similar behavior exists in the binary or pretty-print writers.

wesboyt:
should be a oneline fix: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray

wesboyt:
merged

PR #406, opened 2019-09-11
throws when annotations are not an array full of strings.

wesboyt:
*Issue #, if available:*
#340

*Description of changes:*
Checks that inputs for the function setAnnotations() is an array, and that each annotation within is a string while being written.

### DIFF
--- a/src/AbstractWriter.ts
+++ b/src/AbstractWriter.ts
@@ -15,6 +15,7 @@ import {Writer} from "./IonWriter";
 import {IonType} from "./IonType";
 import {Reader} from "./IonReader";
 import {IonTypes} from "./IonTypes";
+import {_validateAnnotations, _isString} from "./util"
 
 // TS workaround that avoids the need to copy all Writer method signatures into AbstractWriter
 export interface AbstractWriter extends Writer {}
@@ -23,14 +24,17 @@ export abstract class AbstractWriter implements Writer {
     protected _annotations = [];
 
     addAnnotation(annotation: string): void {
+        if(!_isString(annotation)) {
+            throw new Error('Annotation must be of type string.');
+        }
         this._annotations.push(annotation);
     }
 
     setAnnotations(annotations: string[]): void {
-        if (annotations === undefined || annotations === null) {
+        if(annotations === undefined || annotations === null){
             this._annotations = [];
-        } else if(!Array.isArray(annotations)){
-            throw new Error("Annotations must be an array of strings.")
+        } else if (!_validateAnnotations(annotations)) {
+            throw new Error('Annotations must be of type string[].');
         } else {
             this._annotations = annotations;
         }
--- a/src/AbstractWriter.ts
+++ b/src/AbstractWriter.ts
@@ -30,10 +30,8 @@ export abstract class AbstractWriter implements Writer {
     }
 
     setAnnotations(annotations: string[]): void {
-        if (annotations === undefined) {
-            throw new Error('Annotations were undefined.');
-        } else if (annotations === null) {
-            this._annotations = [];
+        if (annotations === undefined || annotations === null) {
+            throw new Error('Annotations were undefined or null.');
         } else if (!this._validateAnnotations(annotations)) {
             throw new Error('Annotations must be of type string[].');
         } else {
--- a/src/AbstractWriter.ts
+++ b/src/AbstractWriter.ts
@@ -15,7 +15,6 @@ import {Writer} from "./IonWriter";
 import {IonType} from "./IonType";
 import {Reader} from "./IonReader";
 import {IonTypes} from "./IonTypes";
-import {_validateAnnotations, _isString} from "./util"
 
 // TS workaround that avoids the need to copy all Writer method signatures into AbstractWriter
 export interface AbstractWriter extends Writer {}
@@ -24,7 +23,7 @@ export abstract class AbstractWriter implements Writer {
     protected _annotations = [];
 
     addAnnotation(annotation: string): void {
-        if(!_isString(annotation)) {
+        if(!this._isString(annotation)) {
             throw new Error('Annotation must be of type string.');
         }
         this._annotations.push(annotation);
@@ -33,7 +32,7 @@ export abstract class AbstractWriter implements Writer {
     setAnnotations(annotations: string[]): void {
         if(annotations === undefined || annotations === null){
             this._annotations = [];
-        } else if (!_validateAnnotations(annotations)) {
+        } else if (!this._validateAnnotations(annotations)) {
             throw new Error('Annotations must be of type string[].');
         } else {
             this._annotations = annotations;
@@ -99,6 +98,22 @@ export abstract class AbstractWriter implements Writer {
                 reader.stepOut();
             }
         }
+    }
+
+    private _validateAnnotations(input : string[]) : boolean {
+        if (!Array.isArray(input)) {
+            return false;
+        }
+        for (let ann in input) {
+            if (!this._isString(ann)) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    private _isString(input : string) : boolean {
+        return typeof input === 'string';
     }
 }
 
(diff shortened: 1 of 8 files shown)


src/IonTextWriter.ts
```diff
@@ -328,6 +328,9 @@ export class TextWriter extends AbstractWriter {
     protected writeAnnotations() : void {
         if (this._annotations === null || this._annotations === undefined) return;
         for (let annotation of this._annotations) {
+            if (typeof annotation !== 'string') {
```
raganhan: don't you need to check that `_annotations` is an array?

I believe it'd be better to do this check where `_annotations` array is being mutated. This way you fail fast when it's initialized by anything that's not an array or when trying to  add an element that's not a string.

src/AbstractWriter.ts
```diff
@@ -27,7 +27,13 @@ export abstract class AbstractWriter implements Writer {
     }
 
     setAnnotations(annotations: string[]): void {
-        this._annotations = annotations;
+        if (annotations === undefined || annotations === null) {
+            this._annotations = [];
+        } else if(!Array.isArray(annotations)){
+            throw new Error("Annotations must be an array of strings.")
+        } else {
+            this._annotations = annotations;
```
raganhan: don't you need to check the elements type as well? You are checking it in `IonBinaryWriter#encodeAnnotations`

src/IonTextWriter.ts
wesboyt: in the setAnnotations call yes we do.

src/AbstractWriter.ts
wesboyt: The elements are getting checked on write, the array is being checked when being set.
users are able to call writer.addAnnotation(string)

raganhan: It's always good to fail fast so it's easier for users to debug and find their mistake

wesboyt: done

src/util.ts
```diff
@@ -27,3 +27,17 @@ export function _hasValue(v: any): boolean {
     return v !== undefined && v !== null;
 }
 
+export function _validateAnnotations(input : string[]) : boolean {
```
pbcornell: please move to AbstractWriter

wesboyt: done

src/IonBinaryWriter.ts
```diff
@@ -311,10 +311,9 @@ export class BinaryWriter extends AbstractWriter {
   }
 
   private encodeAnnotations(annotations: string[]) : Uint8Array {
-    if (!annotations || annotations.length === 0) {
+    if (annotations.length === 0) {
```
raganhan: Since this is a private method I'm assuming the invariant here is that annotations is never `null` nor `undefined`, is this correct? 

If it is add a comment about it

wesboyt: yes, inputs are enforced or set internally.

wesboyt:
> You are only testing `IonBinaryWriter#setAnnotations`, please test all affect APIs:
> 
> * `IonBinaryWriter#addAnnotations`
> * `IonBinaryWriter#setAnnotations`
> * `IonTextWriter#addAnnotations`
> * `IonTextWriter#setAnnotations`
> 
> Also add tests for:
> 
> * Valid case
> * `null`
> * `undefined`
> * invalid types other than `int`

good call

tests/unit/IonTextWriterTest.js
```diff
@@ -311,6 +311,36 @@
...
+      badWriterTest('Cannot add array of chars.',
+          (writer) => { writer.addAnnotation(['t', 'a', 'c', 'o']), writer.writeInt(5) });
+      badWriterTest('Cannot pass annotations array without a string.',
+          (writer) => { writer.setAnnotations([5]), writer.writeInt(5) });
+      badWriterTest('Cannot pass annotations array containing a non string value.',
+          (writer) => { writer.setAnnotations(['a', 5,'t']), writer.writeInt(5) });
+      badWriterTest('Cannot pass annotations array containing undefined.',
+          (writer) => { writer.setAnnotations([undefined]), writer.writeInt(5) });
+      badWriterTest('Cannot pass annotations array containing null',
+          (writer) => { writer.setAnnotations([null]), writer.writeInt(5) });
```
raganhan: add a happy case test for `setAnnotations(null)`

tests/unit/IonBinaryWriterTest.js
```diff
@@ -813,6 +813,22 @@ define([
...
+    errorTest('Cannot add int as annotation.',
+          (writer) => { writer.addAnnotation(5), writer.writeInt(5) });
+    errorTest('Cannot add array of chars.',
+          (writer) => { writer.addAnnotation(['t', 'a', 'c', 'o']), writer.writeInt(5) });
+    errorTest('Cannot pass annotations array containing a non string value.',
+        (writer) => { writer.setAnnotations(['a', 5,'t']), writer.writeInt(5) });
+    errorTest('Cannot pass annotations array containing undefined.',
+        (writer) => { writer.setAnnotations([undefined]), writer.writeInt(5) });
+    errorTest('Cannot pass annotations array containing null',
+        (writer) => { writer.setAnnotations([null]), writer.writeInt(5) });
```
raganhan: add a happy case test for `setAnnotations(null)`

tests/unit/IonTextWriterTest.js
wesboyt: done

tests/unit/IonBinaryWriterTest.js
wesboyt: done

tests/unit/IonTextWriterTest.js
```diff
@@ -311,8 +311,44 @@
...
+          (writer) => {
+              writer.stepIn(ion.IonTypes.LIST);
+              writer.stepOut();
+              writer.stepOut();
+          });
+
 
+    // Symbols
+    writerTest('Writes null symbol by detecting null and allows user to pass null into set annotations',
+          (writer) => { writer.setAnnotations(null); writer.writeSymbol(null) },
```
raganhan: move it to the annotations section

tests/unit/IonBinaryWriterTest.js
```diff
@@ -625,10 +625,12 @@ define([
       ]);
 
     // Symbols
-
     writerTest('Writes null symbol by detecting null',
       (writer) => { writer.writeSymbol(null) },
         [0x7f]);
+    writerTest('Writes null symbol by detecting null and allows user to pass null into set annotations',
```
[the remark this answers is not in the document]
wesboyt: nope

tests/unit/IonTextWriterTest.js
wesboyt: okay

tests/unit/IonBinaryWriterTest.js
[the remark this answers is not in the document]
raganhan: If you disagree with a PR comment explain why or if need clarification say so. Answering "nope" is extremely unhelpful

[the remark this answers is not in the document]
wesboyt: sorry i meant were gonna throw on null instead as the apis no longer return null.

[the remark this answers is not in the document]
raganhan: ok. In any case the comment is to use a test description that describes the expected behavior and not what the test is doing. 

When you change to throw on `null` update the description with the expected behavior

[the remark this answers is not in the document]
wesboyt: gotcha, also the reason i dont plan on getting into parameterization is that were going to delete all these tests for mocha shortly.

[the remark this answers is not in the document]
raganhan: Ok that makes sense. As we talked offline if you can write the mocha tests now might as well do it, but since the test framework migration is in flight I'm ok waiting until mocha is fully integrated to migrate and reduce the duplication 

I'll let you decide what's the best path here

wesboyt:
tracking further progress in https://github.com/amzn/ion-js/issues/458

Issue #341, opened 2019-08-29
TEXT: stepIn(iontypes.STRUCT) whilst in a struct does not throw.


wesboyt:
We should be enforcing fieldnames be written before the next value but seem to be missing that logic when stepping into a container (on the text side) (we throw on scalars).

wesboyt:
https://github.com/amzn/ion-js/pull/319 dblock was able to serialize nested structs with no fieldnames.

wesboyt:
reproduction:
let data = "{ x: 1 }";
        let reader = ion.makeReader(data);
        let writer = ion.makeTextWriter();
        writer.stepIn(ion.IonTypes.STRUCT);
        writer.stepIn(ion.IonTypes.STRUCT, ['a', 'b']);

        reader.next();
        reader.stepIn();
        reader.next();

        util._writeValue(reader, writer, 1);

        reader.stepOut();

        writer.stepOut();
        writer.stepOut();
produces: 
{a::b::{{x:1}}}
which is illegal.

wesboyt:
merged

PR #407, opened 2019-09-11
Throws when containers are written inside of a struct without a field…

wesboyt:
…name.

*Issue #, if available:*
#341
*Description of changes:*
Throws when a user attempts to write a container within a struct without a paired fieldname.


src/IonTextWriter.ts
```diff
@@ -226,6 +226,9 @@ export class TextWriter extends AbstractWriter {
     }
 
     stepIn(type: IonType) : void {
+        if (this.currentContainer.containerType === IonTypes.STRUCT && this.currentContainer.state === State.STRUCT_FIELD) {
```
almann: Isn't this technically redundant as `this.currentContainer.state === State.STRUCT_FIELD` implies `this.currentContainer.containerType === IonTypes.STRUCT`?

src/IonBinaryWriter.ts
```diff
@@ -375,7 +375,7 @@ export class BinaryWriter extends AbstractWriter {
   private checkWriteValue() : void {
     this.checkClosed();
     if (this.state === States.STRUCT_FIELD) {
-      throw new Error("Expected a struct field name instead of a value");
+      throw new Error("Expected a struct field name instead of a value, Call writeFieldName(string) with the desired name before calling stepIn(IonType) or writeIonType()");
```
zslayton: Nit: The 'c' in "Call" follows a comma and so doesn't need to be capitalized.
> ...instead of a value. Call writeFieldName(string)...

or

> instead of a value, call writeFieldName(string)

would work.

dblock:
I feel like this could benefit from a higher level test. The original issue was doing this:

```js
        let data = "{ x: 1 }";
        let reader = ion.makeReader(data);
        writer.stepIn(ion.IonTypes.STRUCT);
        writer.stepIn(ion.IonTypes.STRUCT, ['a', 'b']);
        reader.next();
        reader.stepIn();
        reader.next();
        writer.writeValues(reader);
        reader.stepOut();
        writer.stepOut();
        writer.stepOut();
```

So maybe not something this long, but at least demonstrating the spirit of what was attempted and where it should fail. 

This would help prevent regressions in case lower level semantics change.

wesboyt:
this test is doing so?
https://github.com/amzn/ion-js/blob/master/tests/unit/IonTextWriterTest.js#L363

dblock:
At a first glance that felt too minimal, but you're right it's the minimum case for this problem. All set, thank you.

PR #409, opened 2019-09-11
Fixes decimal tests to work across Node 8/10/12.

### LINKED ISSUE #399: Decimal Test Regression in Node v12
As mentioned in #380, there is an error around decimal tests when running under Node v12:

```
× Decimal - 123d-5 (0.002s)
AssertionError: numberValue(): expected 0.00123 to equal 0.0012300000000000002
  at test  <tests/unit/IonDecimalTest.js:194:20>
  at Test.123d-5 [as test]  <tests/unit/IonDecimalTest.js:50:33>
```

The regression is likely around floating point changes in the release, so one should probably look there first.

almann:
The underlying issue is that we test using `Decimal.numberValue()`
which uses `Math.pow(10, exponent)`.  This API has different rounding
behaviors (which is allowed by the ECMAScript spec) in v8/v10 of
Node vs. v12 of Node.  This is likely to be an issue on different
engines as well.

The fix allows the `test` method to support an inclusive interval of
valid `number` values that the `numberValue()` representation can be
and fixes the one test that seems to trip on this.

Resolves #399

### DIFF
--- a/tests/unit/IonDecimalTest.js
+++ b/tests/unit/IonDecimalTest.js
@@ -47,7 +47,7 @@ define([
             '123d1'     : () => test('123d1', '123', 1, 1230, '1.23E+3'),
             '123d3'     : () => test('123d3', '123', 3, 123000, '1.23E+5'),
             '123d-1'    : () => test('123d-1', '123', -1, 12.3, '12.3'),
-            '123d-5'    : () => test('123d-5', '123', -5, 0.0012300000000000002, '0.00123'),
+            '123d-5'    : () => test('123d-5', '123', -5, [0.00123, 0.0012300000000000002], '0.00123'),
             '123d-10'   : () => test('123d-10', '123', -10, 1.2300000000000001e-8, '1.23E-8'),
             '-123d-12'  : () => test('-123d-12', '-123', -12, -0.000000000123, '-1.23E-10'),
             '12300d-2'  : () => test('12300d-2', '12300', -2, 123, '123.00'),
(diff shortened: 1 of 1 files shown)


tests/unit/IonDecimalTest.js
```diff
@@ -175,6 +175,14 @@ define([
             'intValue(-1.0000001)': () => () => assert.equal(ion.Decimal.parse('-1.0000001').intValue(), -1),
         });
 
+        /**
+         * @param decimalString The text representation of the decimal to test.
+         * @param expectedCoefficient The expected `number` or `string` (LongInt) for the coefficient.
+         * @param expectedExponent The expected `number` or `string` (LongInt) for the exponent part.
```
pbcornell: always a `number`;  i expect `string` to fail on `util._sign(expectedExponent)`

tests/unit/IonDecimalTest.js
```diff
@@ -175,6 +175,14 @@ define([
             'intValue(-1.0000001)': () => () => assert.equal(ion.Decimal.parse('-1.0000001').intValue(), -1),
         });
 
+        /**
+         * @param decimalString The text representation of the decimal to test.
+         * @param expectedCoefficient The expected `number` or `string` (LongInt) for the coefficient.
```
pbcornell: currently always a `string`, but agree a `number` could be passed

tests/unit/IonDecimalTest.js
almann: Oh interesting--I noted that it all plumbs down to `LongInt`, which is why I wrote it that way.

Issue #405, opened 2019-09-11
SharedSymbolTable leaks implementation details


zslayton:
SharedSymbolTable currently [allows users to directly access the `Array` being used to store its symbols' text](https://github.com/amzn/ion-js/blob/master/src/IonSharedSymbolTable.ts#L34-L36). This means that alternative `SymbolTable` implementations like [SubstituteSymbolTable](https://github.com/amzn/ion-js/blob/master/src/IonSubstituteSymbolTable.ts) *must* use an `Array` to store their data [even if it doesn't make sense to](https://github.com/amzn/ion-js/issues/404).

zslayton:
Closed by #424.

Issue #404, opened 2019-09-11
SubstituteSymbolTable should not allocate an array


zslayton:
While most of our `ion-tests` file tests take <0.01s to process, [`ion-tests/iontestdata/good/subfieldVarUInt.ion`](https://github.com/amzn/ion-tests/blob/f7a4999a416b98a4ab454ba18843a177404b7a45/iontestdata/good/subfieldVarUInt.ion) currently takes over 40s on my laptop. This is because the test [imports a non-existent symbol table with an enormous `maxId`](https://github.com/amzn/ion-tests/blob/f7a4999a416b98a4ab454ba18843a177404b7a45/iontestdata/good/subfieldVarUInt.ion).

Because that symbol table cannot be resolved via the `Catalog`, the `Reader` [falls back to instantiating](https://github.com/amzn/ion-js/blob/master/src/IonSymbols.ts#L66) a `SubstituteSymbolTable`, a symbol table for which [all symbols are undefined](https://github.com/amzn/ion-js/blob/master/src/IonSubstituteSymbolTable.ts#L17-L18).

Despite having no symbol data to manage, the `SubstituteSymbolTable` still [allocates an array of length `maxId`](https://github.com/amzn/ion-js/blob/master/src/IonSubstituteSymbolTable.ts#L23). When `maxId` is enormous, this can take tens of seconds to complete.

We should modify the implementation to simply return `0` for any symbol IDs it's asked to resolve that fall within in the range of `offset` to `maxId`. This will require [some refactoring of the `SharedSymbolTable` contract](https://github.com/amzn/ion-js/issues/405), which currently allows for direct access to the underlying `Array` that is usually in use.

zslayton:
Closed by #424.

Issue #121, opened 2018-04-06
SymbolTokens


wesboyt:
SymbolTokens and import descriptors will need to be implemented.

zslayton:
Related to #155.

PR #424, opened 2019-09-16
SymbolTable refactoring

zslayton:
*Issue #, if available:* #404, #405

### Overview

This PR refactors the `SharedSymbolTable`, `SubstituteSymbolTable`, and `Import` classes for better encapsulation. 

These changes allowed `SubstituteSymbolTable` to [avoid allocating](https://github.com/amzn/ion-js/issues/404), which reduced the time required to run `grunt test` on my laptop from `2m14.116s` to `0m30.987s`. 🎉 🍾 

### Low-level changes
- The `SharedSymbolTable#symbols()` method has been removed, as it required subclasses (like `SubstituteSymbolTable`) to maintain an array even if it wasn't called for.
- Added `numberOfSymbols()` accessor that can be used instead of `symbols().length`.
- `SubstituteSymbolTable` simply returns `undefined` when asked to resolve a symbol instead of allocating an array filled with `undefined` values and consulting it.
- `SharedSymbolTable` manages its own index from symbol text -> symbol ID rather than requiring outside assistance from `Import`.
- Renamed `getSymbol()` to `getSymbolText()` both to be more explicit and to preserve the `getSymbol` name for a possible later version of the API that returns a `SymbolToken`.

### Out of scope
There is still plenty of room for improvement in this portion of the code base. The following list contains changes which we can/should pursue, but which fell outside the scope of this PR:
* Introducing a `SymbolTable` interface that all of the symbol-table-like classes adhere to.
* Replacing all uses of `undefined` as a possible return value with a well-defined `UNKNOWN_SYMBOL`. (See #425.)
* Having the API center around a proper `SymbolToken` class instead of offering separate methods for `getSymbolText` and `getSymbolId`. (See #121.)

### DIFF
--- a/src/IonImport.ts
+++ b/src/IonImport.ts
@@ -28,7 +28,6 @@ import { SymbolIndex } from "./IonSymbolIndex";
 export class Import {
     private readonly _offset: number;
     private readonly _length: number;
-    private readonly index: SymbolIndex = {};
     private readonly _parent : Import;
     private readonly _symbolTable : SharedSymbolTable;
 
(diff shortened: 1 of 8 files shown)


src/IonImport.ts
```diff
@@ -28,39 +28,49 @@ import { SymbolIndex } from "./IonSymbolIndex";
...
             }
         }
 
         let index: number = symbolId - this.offset;
-        if (index < this.length) {
-            return this.symbolTable.symbols[index];
+        if (index >= 0 && index < this.length) {
+            return this.symbolTable.getSymbolText(index);
         }
+        return undefined;
```
almann: Do we really want this to return `undefined`, i.e. do we really want this to be the contract?

src/IonImport.ts
```diff
@@ -28,39 +28,49 @@ import { SymbolIndex } from "./IonSymbolIndex";
...
+        }
+
+        symbolId = this.symbolTable.getSymbolId(symbolText);
+        if (symbolId !== null
+            && symbolId !== undefined
+            && symbolId < this.length) {
+            return symbolId + this.offset;
+        }
+
+        return undefined;
```
almann: Same question around `undefined`.

src/IonSharedSymbolTable.ts
```diff
@@ -31,7 +63,7 @@ export class SharedSymbolTable {
     return this._version;
   }
 
-  get symbols() : string[] {
-    return this._symbols;
+  get numberOfSymbols(): number {
```
almann: Do we like this spelling over something like `size`, `length`, or `count`?

wesboyt: I prefer length but its purely an opinion

src/IonImport.ts
wesboyt: anywhere we can ripout undefined we should do so

zslayton: As noted in the `Out of Scope` section of the overview, we should replace all instances of `undefined` with an `UNKNOWN_SYMBOL` constant. Ideally, this would be combined with the introduction of a proper `SymbolToken` class.

PR #426, opened 2019-09-16
Bug? Single quote cobbled.

dblock:
Looks like `writeText` is garbling `’` as "â or `русский`. Something with encoding?

A more complete example where this appears in a value string:

```
"{Description:\"world’s most popular and authoritative source\"}"
```


dblock:
I'll take this back. We needed to decode those bytes correctly with `ion.decodeUtf8`.

PR #433, opened 2019-09-19
properly converts a UTF-16 surrogate pair into a 'character'

### LINKED ISSUE #430: text reader produces a string containing digits when reading a string containing an emoji
```javascript
let r = ion.makeReader('"👩"');       // this codepoint U+1F469 (see https://emojipedia.org/emoji/%F0%9F%91%A9/)
r.next();
console.log(r.stringValue());
```
produces:
```
111782
```
Note that this is the sum of the high- and low-surrogates (55357 + 56425 = 111782).

pbcornell:
Resolves #430 

### DIFF
--- a/src/IonParserTextRaw.ts
+++ b/src/IonParserTextRaw.ts
@@ -22,7 +22,7 @@ import * as IonText from "./IonText";
 
 import { IonType } from "./IonType";
 import { IonTypes } from "./IonTypes";
-import { StringSpan, Span } from "./IonSpan";
+import { StringSpan } from "./IonSpan";
 import {is_keyword, is_whitespace} from "./IonText";
 
 const EOF = -1;  // EOF is end of container, distinct from undefined which is value has been consumed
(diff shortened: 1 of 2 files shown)


src/IonParserTextRaw.ts
```diff
@@ -1060,9 +1062,13 @@ export class ParserTextRaw {
...
                         if (this.isLowSurrogate(tempChar)) {
-                            s += ch + tempChar;
+                            // convert from UTF-16 surrogate pair to a codepoint
+                            let hiSurrogate = ch;
+                            let loSurrogate = tempChar;
+                            let codepoint = 0x10000 + ((hiSurrogate & _UTF16_MASK) << 10) + (loSurrogate & _UTF16_MASK);
+                            s += String.fromCodePoint(codepoint);
                         } else {
-                            throw new Error("illegal high surrogate" + ch);
+                            throw new Error("illegal low surrogate: " + ch);
```
wesboyt: shouldn't this still say high surrogate?

pbcornell: This error is thrown in the else block of an `isLowSurrogate?` test, so I think "illegal low surrogate" is the right error message.  The code was expecting to find a low surrogate, but didn't.

wesboyt: means an unpaired high surrogate

pbcornell: I've attempted to clarify this as well as the "illegal low surrogate" error thrown a few lines below this.  Feel free to suggest better messages.

PR #434, opened 2019-09-19
improves escaping of non-printable ASCII characters

### LINKED ISSUE #431: text writer produces invalid Ion when writing strings containing non-printable ASCII
I suspect this issue applies to symbols and clobs, too.

```javascript
let w = ion.makeTextWriter();
for (let x = 0; x < 32; x++) {
    w.writeString(String.fromCharCode(x));
}
w.close();
console.log(String.fromCharCode.apply(null, w.getBytes()));
```

produces:
```
"\0"
"\u0001"
"\u0002"
"\u0003"
"\u0004"
"\u0005"
""                    // invalid Ion
"\a"
"\b"
"\t"
"\n"
"\v"
"\f"
"\r"
""                    // invalid Ion
""                    // invalid Ion
""                    // invalid Ion
""                    // invalid Ion
""                    // invalid Ion
""                    // invalid Ion
""                    // invalid Ion
""                    // invalid Ion
""                    // invalid Ion
""                    // invalid Ion
""                    // invalid Ion
""                    // invalid Ion
""                    // invalid Ion
""                    // invalid Ion
""                    // invalid Ion
""                    // invalid Ion
""                    // invalid Ion
""                    // invalid Ion
```

pbcornell:
Fills out the text escape tables a bit, and changes from `\u` to `\x` escapes so the a number of table entries can be shared with clobs (whereas `\u` is not a valid clob escape).

Resolves #431 

### DIFF
--- a/src/IonText.ts
+++ b/src/IonText.ts
@@ -267,16 +267,16 @@ function toCharCodes(s: string) {
   return charCodes;
 }
 
+const _HEX_ESCAPE_PREFIX = [CharCodes.BACKSLASH, CharCodes.LOWERCASE_X];
 function hexEscape(codePoint: number) : number[] {
-    let prefix: number[] = [CharCodes.BACKSLASH, CharCodes.LOWERCASE_X];
     let hexEscape: string = codePoint.toString(16);
     while (hexEscape.length < 2) {
         hexEscape = "0" + hexEscape;
     }
-    return prefix.concat(toCharCodes(hexEscape));
+    return _HEX_ESCAPE_PREFIX.concat(toCharCodes(hexEscape));
 }
 
-function hexEscapes(escapes: EscapeIndex, start: number, end?: number) {
+function populateWithHexEscapes(escapes: EscapeIndex, start: number, end?: number) {
     if (end === undefined) {
         escapes[start] = hexEscape(start);
     } else {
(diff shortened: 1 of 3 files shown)


src/IonText.ts
```diff
@@ -267,51 +267,44 @@ function toCharCodes(s: string) {
...
 
-function unicodeEscape(codePoint: number) : number[] {
-  let prefix: number[] = [CharCodes.BACKSLASH, CharCodes.LOWERCASE_U];
-  let hexEscape: string = codePoint.toString(16);
-  while (hexEscape.length < 4) {
-    hexEscape = "0" + hexEscape;
-  }
-  return prefix.concat(toCharCodes(hexEscape));
+function hexEscape(codePoint: number) : number[] {
+    let prefix: number[] = [CharCodes.BACKSLASH, CharCodes.LOWERCASE_X];
```
zslayton: Consider promoting `prefix` to a static value to avoid the cost of allocation/initialization on each call. The call to [`concat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat) below will create a new `Array` instance anyway.

src/IonText.ts
```diff
@@ -267,51 +267,44 @@ function toCharCodes(s: string) {
...
-ClobEscapes[CharCodes.BACKSLASH] = backslashEscape("\\");
-
-    function unicodeEscapes(escapes: EscapeIndex, start: number, end?: number) {
-        if (end === undefined) {
-            escapes[start] = unicodeEscape(start);
-        } else {
-            for (let i: number = start; i < end; i++) {
-                escapes[i] = unicodeEscape(i);
-            }
+function hexEscapes(escapes: EscapeIndex, start: number, end?: number) {
```
zslayton: Could you make the name more descriptive (maybe `populateHexEscapeIndex`?) or add a comment to that effect?

pbcornell: Renamed to `populateWithHexEscapes`

src/IonText.ts
pbcornell: Promoted to `_HEX_ESCAPE_PREFIX`

PR #435, opened 2019-09-22
Added Reader#deepValue.

dblock:
*Issue #, if available:*

#338, reopening #339 after rebasing and renaming `raw` to `deepValue`.

*Description of changes:*

Proposing an implementation to get deep data from an Ion reader. In my scenario we get an ion document then locate a path within the document, and extract a subdocument from it. This gives a text/binary representation of the sub document. 

Maybe there should be a way to get an ion document from any location as well?

---

### DIFF
--- a/src/IonReader.ts
+++ b/src/IonReader.ts
@@ -152,4 +152,14 @@ export interface Reader {
      * @throw Error when the reader is not positioned on a scalar value.
      */
     value(): ReaderScalarValue;
+
+    /**
+     * Returns a representation of the ion document.
+     * 
+     * @return `string` String value when text.
+     * @returns `Uint8Array` Binary array when binary.
+     * 
+     * @throws
+     */
+    deepValue: () => string | Uint8Array;
 }
--- a/tests/unit/IonBinaryReaderTest.js
+++ b/tests/unit/IonBinaryReaderTest.js
@@ -30,6 +30,12 @@ define([
                 reader.next();
                 assert.deepEqual(reader.timestampValue(), timestamp);
             },
+
+            'deepValue': () => {
+              var ivm = [0xe0, 0x01, 0x00, 0xea];
+              var ionReader = ion.makeReader(ivm);
+              assert.deepEqual(ionReader.deepValue(), new Uint8Array(ivm));
+            }
         });
     }
 );
--- a/tests/unit/IonTextReaderTest.js
+++ b/tests/unit/IonTextReaderTest.js
@@ -226,6 +226,19 @@ define(
 
 
         }
+
+        function testDeepValue(ionToRead, expectedOutput = ionToRead) {
+            var ionReader = ion.makeReader(ionToRead);
+            assert.equal(ionReader.deepValue(), expectedOutput);
+        }
+
+        suite['data'] = function() {
+            testDeepValue("123");
+            testDeepValue("{x : 123}", "{x:123}");
+            testDeepValue("{x:{y:[\"z\"]}}");
+            testDeepValue("a::123");
+        }
+
         registerSuite(suite);
     }
 );
--- a/src/IonBinaryReader.ts
+++ b/src/IonBinaryReader.ts
@@ -21,8 +21,6 @@
 import { Catalog } from "./IonCatalog";
 import { Decimal } from "./IonDecimal";
 import { defaultLocalSymbolTable } from "./IonLocalSymbolTable";
-import { getSystemSymbolTable } from "./IonSystemSymbolTable";
-import { Import } from "./IonImport";
 import { ion_symbol_table_sid } from "./IonSymbols";
 import { IonType } from "./IonType";
 import { IonTypes } from "./IonTypes";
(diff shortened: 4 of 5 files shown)


dblock:
I have not yet thought too deeply (paf.) about @pbcornell's comments in https://github.com/amzn/ion-js/pull/339#issuecomment-526680369, which seem entirely valid reasons to reject this PR.

src/IonReader.ts
```diff
@@ -152,4 +152,14 @@ export interface Reader {
...
+
+    /**
+     * Returns a representation of the ion document.
+     * 
+     * @return `string` String value when text.
+     * @returns `Uint8Array` Binary array when binary.
+     * 
+     * @throws
+     */
+    deepValue: () => string | Uint8Array;
```
almann: I am a little confused as to why this is called `deepValue()` since in theory it could be called at an point in the `Reader` including the top-level.  Also, the doc comment really should indicate the caveats to this API.  For example, I *assume* (though not sure it is the intent), that the local symbol table context is not necessarily available using this API and the caller needs to be aware of such things.  Is that correct?

src/IonReader.ts
```diff
@@ -152,4 +152,14 @@ export interface Reader {
...
      */
     value(): ReaderScalarValue;
-}
\ No newline at end of file
+
+    /**
+     * Returns a representation of the ion document.
+     * 
+     * @return `string` String value when text.
+     * @returns `Uint8Array` Binary array when binary.
```
almann: It may be the case that a text reader is over a `Uint8Array`, that is to say that the underlying type would be a function of the source that the reader is parsing over, but it is not necessarily tied to the actual text vs. binary format itself.  Let us presume that someone passed an array of numbers (as octets), what would an implementer have to do--do they have to normalize to `Uint8Array`?

*Minor:* `return` and `returns` (let's just choose one).

tests/unit/IonBinaryReaderTest.js
```diff
@@ -30,6 +30,12 @@ define([
                 reader.next();
                 assert.deepEqual(reader.timestampValue(), timestamp);
             },
+
+            'deepValue': () => {
+              var ivm = [0xe0, 0x01, 0x00, 0xea];
+              var ionReader = ion.makeReader(ivm);
+              assert.deepEqual(ionReader.deepValue(), new Uint8Array(ivm));
```
almann: Oh that is interesting, I would not expect `deepValue()` to work on a `Reader` that is not yet positioned on a value.  The IVM is not a value, per se, so I think we need to be a bit careful here.  E.g. what is the implication of doing this when there are local symbol table(s).

tests/unit/IonTextReaderTest.js
```diff
@@ -226,6 +226,19 @@ define(
 
 
         }
+
+        function testDeepValue(ionToRead, expectedOutput = ionToRead) {
+            var ionReader = ion.makeReader(ionToRead);
+            assert.equal(ionReader.deepValue(), expectedOutput);
```
almann: Again, to have this work the way it is asserted below, wouldn't you have to `next` first?

src/IonReader.ts
dblock: I think you are correct. I meant "deep" because it goes into the structure recursively.

dblock:
@pbcornell @almann Thank you for your comments. I think I want to step back from this implementation and maybe raise what use case I am trying to solve.

I get an ION document in some request. It looks very much like some JSON with some nested data. Then I pluck out that nested data. Example:

```
{
  x: {
    y: {
     z: 1
   }
  }
}
```

```
x.y => { z: 1 }
```

And I pass `{ z: 1 }` to some other service. 

The intent of `deepValue` is to be able to navigate within an ion document to `x.y`, then pluck out whatever is underneath. Currently I do the whole reader/write dance of getting that nested data, but it felt to me that it belonged in the ion library proper. LMK what you think.

wesboyt:
> @pbcornell @almann Thank you for your comments. I think I want to step back from this implementation and maybe raise what use case I am trying to solve.
> 
> I get an ION document in some request. It looks very much like some JSON with some nested data. Then I pluck out that nested data. Example:
> 
> ```
> {
>   x: {
>     y: {
>      z: 1
>    }
>   }
> }
> ```
> 
> ```
> x.y => { z: 1 }
> ```
> 
> And I pass `{ z: 1 }` to some other service.
> 
> The intent of `deepValue` is to be able to navigate within an ion document to `x.y`, then pluck out whatever is underneath (with `deepValue`). Currently I do the whole reader/write dance of getting that nested data, but it felt to me that it belonged in the ion library proper. LMK what you think.

Sounds like you want IonPathExtractor. We probably will end up adding it later on but it isn't a priority right now.

pbcornell:
> @pbcornell @almann Thank you for your comments. I think I want to step back from this implementation and maybe raise what use case I am trying to solve.

@dblock If I set aside concerns about the symbol table complexities and the potential memory implications (i.e., for large subdocuments), I remain concerned that the proposal is to add a *convenience* method to the `Reader` interface, and that `deepValue()` doesn't align well with the streaming-orientation of the interface.  We should avoid adding methods to `Reader` without a particularly strong cause; imo, `deepValue()` doesn't make the cut.

The contract of `Writer.writeValues()` addresses this use case without enabling full representation of a subdocument in memory (and yes, I am glossing over the fact that the library doesn't currently support reading from or writing to a stream).  `Writer.writeValues()` additionally supports use cases where the target encoding (binary/pretty/text) is different.

If at some point a DOM/Tree API is added to ion-js, I suspect that will lend itself more naturally to this use case and the convenience you're looking for (along with the memory implications, but then a DOM/Tree API shouldn't pretend to be a memory-efficient).

dblock:
Makes sense, feel free to close if you see fit.

PR #436, opened 2019-09-24
compares binary to text before returning the in memory representation.

wesboyt:
Makes sure that the binary serialization is equal to the text serialization on read.

### DIFF
--- a/src/IonEventStream.ts
+++ b/src/IonEventStream.ts
@@ -12,12 +12,11 @@
  * language governing permissions and limitations under the License.
  */
 
-import { Reader } from "./IonReader";
-import {IonEvent, IonEventFactory} from "./IonEvent";
-import { Writer } from "./IonWriter";
-import { IonType } from "./IonType";
-import { IonTypes } from "./IonTypes";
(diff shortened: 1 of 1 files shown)


src/IonEventStream.ts
```diff
@@ -301,6 +299,22 @@ export class IonEventStream {
             currentEvent['isNull'],
             currentEvent['value_text']
         );
+
+        let binaryEvent = this.eventFactory.makeEvent(
```
tgregg: This can go inside the `if(eventType === IonEventType.SCALAR){` block below since it's only needed in that case. To simplify it, you could also just do a data model equivalence comparison on `value_text` and `value_binary` if it's a scalar event. That way you don't need to create a dummy event for binary.

wesboyt:
> The change looks good, but could you comment on the impact of this PR? For example: does this allow us to remove tests from our skip lists? Does it make all of the existing event stream tests more thorough?

This does not allow us to remove tests from our skip lists afaik, it does increase the possible edge cases we can catch when reading eventstreams from other serializers than our own.

PR #437, opened 2019-09-25
Equivalence fix

wesboyt:
continues previously closed pr: https://github.com/amzn/ion-js/pull/302
*Description of changes:*
speeds up equivalence significantly, 


src/IonEvent.ts
```diff
@@ -395,20 +395,28 @@ class IonStructEvent extends AbsIonContainerEvent {//no embed support as of yet.
...
+                    let child = actualEvents[i]
+                    let expectedChild = expectedEvents[j]
+                    matchFound = child.equals(expectedChild);
+                    if (matchFound) paired[j] = true;
+                    if (matchFound && child.eventType === IonEventType.CONTAINER_START) {
+                        for (let k = j + 1; k < expectedChild.ionValue.length; k++) {
+                            paired[k] = true;
+                        }
+                        i += child.ionValue.length;
+                    }
```
almann: So this is basically doing a bag comparison, yes?  A comment explaining what the intent of this algorithm is would be helpful.  Also, explaining why ES6 `Map` cannot be used to reduce this O(N**2) to O(N) would be helpful.

wesboyt: it can be done in O(n)!

PR #438, opened 2019-09-25
Roundtrip fixes

wesboyt:
*Issue #, if available:*
continues https://github.com/amzn/ion-js/tree/roundtripFix work after master merge.
*Description of changes:*
Removes most of the repetitive logic within roundtripEventStream.
The last stream is not strictly necessary.


tests/unit/iontests.js
```diff
@@ -273,30 +273,46 @@ define([
...
+            let textWriter = new ion.makeTextWriter();
+            textWriter.writeValues(new ion.makeReader(input));
+            textWriter.close();
+            let text = textWriter.getBytes();
+            streams.push(new es.IonEventStream(new ion.makeReader(text)));
+            let binaryWriter = new ion.makeBinaryWriter();
+            binaryWriter.writeValues(new ion.makeReader(input));
+            binaryWriter.close();
+            let binary = binaryWriter.getBytes();
+            streams.push(new es.IonEventStream(new ion.makeReader(binary)));
```
almann: Minor: Some comments in this wall of code would be nice, what is the intention here, we have a stream of the original events, followed by the events from a serialize/read as text, then etc...  Being clear about the combinatorics would be rather nice to have for the next person.

tests/unit/iontests.js
```diff
@@ -273,30 +273,46 @@ define([
...
+
+            //second pass
+            textWriter = new ion.makeTextWriter();
+            textWriter.writeValues(new ion.makeReader(text));
+            textWriter.close();
+            streams.push(new es.IonEventStream(new ion.makeReader(textWriter.getBytes())));
```
almann: I can't help but notice that there sure is a lot of copy-paste, I would hoist the logic into a generic helper.

Perhaps something like:

```javascript
function makeStream(src, makeWriter) {
  const writer = makeWriter;
  writer.writeValues(new ion.makeReader(src));
  writer.close();
  const serialized = writer.getBytes();
  return new es.IonEventStream(new ion.makeReader(serialized));
}
```

Then the call sites would be:

```javascript
// second pass over input as serialized as text...
streams.push(makeStream(text, ion.makeTextWriter));
```

I think that would make this wall of code way easier to read.

tests/unit/iontests.js
```diff
@@ -273,30 +273,46 @@ define([
...
 
-            let eventWriter = ion.makeTextWriter();
-            streams[0].writeEventStream(eventWriter);
-            eventWriter.close();
-            streams.push(new es.IonEventStream(new ion.makeReader(eventWriter.getBytes())));
+            binaryWriter = new ion.makeBinaryWriter();
```
almann: Should this have a `new`?  I thought `makeBinaryWriter` was a simple function not a constructor.

tests/unit/iontests.js
wesboyt: done

tests/unit/iontests.js
wesboyt: i hope the limited comments I added made it a bit more clear.

tests/unit/iontests.js
wesboyt: doesnt seem necessary, I'll remove it

tests/unit/iontests.js
```diff
@@ -273,46 +273,43 @@ define([
             return fs.readFileSync(path, options);
         }
 
+        function makeStream(src, makeWriter) {
+            let writer = makeWriter;
```
almann: In my feedback, I suggested that `makeWriter` be a function argument, not the `writer` itself.

If you make this actually a *function* (i.e. callable), you can streamline the code below as:

```javascript
streams.push(makeStream(input, ion.makeTextWriter));
```

And so-on.  If you choose not to take this advice, please don't have a redundant parameter called `makeWriter` that is literally just assigned to a local named `writer`.

wesboyt: I went with streams.push(makeStream(input, ion.makeTextWriter()));

PR #439, opened 2019-09-25
Added support for Mocha testing.

zslayton:
*Description of changes:*

This is the first of a three part PR series:
1. **Add support for writing Mocha tests in Typescript.**
2. Migrate all existing intern.js 3 tests to Mocha.
3. Remove intern.js and migrate all grunt build targets to the new mocha tasks.

I've converted the unit tests for the `util` module to the new format as an example.

Once this branch is merged, you'll be able to run `grunt mocha` and `grunt mocha:coverage` from the command line. `grunt:test` will still point to our intern.js tests for now.

I'll add review comments detailing what each section achieves.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -169,17 +169,37 @@ module.exports = function(grunt) {
         }
       }
     },
-      uglify: { 
-          options: { 
-              compress: true, 
-              mangle: true, 
-              sourceMap: false
-          }, 
-          target: { 
-              src: './dist/browser/js/ion-bundle.js',
-              dest: './dist/browser/js/ion-bundle.min.js',
-          }
+    uglify: {
+        options: {
+            compress: true,
+            mangle: true,
+            sourceMap: false
+        },
+        target: {
+            src: './dist/browser/js/ion-bundle.js',
+            dest: './dist/browser/js/ion-bundle.min.js',
+        }
+    },
+    mochaTest: {
+      test: {
+        options: {
+          ui: 'mocha-typescript',
+          require: [
+              'ts-node/register',
+              'source-map-support/register',
+              'nyc'
+          ],
+          noFail: false // Run all tests even if an early test fails
+        },
+        src: ['test/**/*.ts']
       }
+    },
+    shell: {
+      // Uses the nyc configuration from package.json and the mocha settings from test/mocha.opts
+      "mochaWithCoverage": {
+        command: "npx nyc mocha --colors"
+      }
+    }
   });
 
   grunt.loadNpmTasks('grunt-babel');
@@ -188,6 +208,8 @@ module.exports = function(grunt) {
   grunt.loadNpmTasks('grunt-contrib-clean');
   grunt.loadNpmTasks('grunt-contrib-copy');
   grunt.loadNpmTasks('grunt-contrib-uglify');
+  grunt.loadNpmTasks('grunt-mocha-test');
+  grunt.loadNpmTasks('grunt-shell');
   grunt.loadNpmTasks('grunt-ts');
   grunt.loadNpmTasks('grunt-typedoc');
   grunt.loadNpmTasks('remap-istanbul');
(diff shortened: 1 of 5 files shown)


Gruntfile.js
```diff
@@ -169,17 +169,37 @@ module.exports = function(grunt) {
         }
       }
     },
-      uglify: {
```
zslayton: Correcting the indentation for the `uglify` config section.

Gruntfile.js
```diff
@@ -169,17 +169,37 @@ module.exports = function(grunt) {
...
+            compress: true,
+            mangle: true,
+            sourceMap: false
+        },
+        target: {
+            src: './dist/browser/js/ion-bundle.js',
+            dest: './dist/browser/js/ion-bundle.min.js',
+        }
+    },
+    mochaTest: {
```
zslayton: Configures the `grunt-mocha-test` package, which allows us to run Mocha tests via Grunt. Another file below (`mocha.opts`) configures the same settings, but Grunt can't use it directly.

Gruntfile.js
```diff
@@ -169,17 +169,37 @@ module.exports = function(grunt) {
...
+              'ts-node/register',
+              'source-map-support/register',
+              'nyc'
+          ],
+          noFail: false // Run all tests even if an early test fails
+        },
+        src: ['test/**/*.ts']
       }
+    },
+    shell: {
```
zslayton: The `nyc` command is the CLI interface to `Instanbul`, our code coverage tool. There's not an easy way to mix Grunt + Typescript + Mocha + nyc, so I've opted to simply invoke it as a child process.

package.json
```diff
@@ -31,6 +30,17 @@
...
+  "nyc": {
+    "extends": "@istanbuljs/nyc-config-typescript",
+    "all": true,
+    "include": [
+      "src/**.ts"
+    ],
+    "exclude": [
+      "src/.**"
+    ],
+    "check-coverage": false
```
zslayton: Prevents low test coverage from failing the build. Once more tests are migrated over we can configure the desired thresholds and re-enable this.

test/util.ts
```diff
@@ -0,0 +1,38 @@
+import {assert} from 'chai';
```
zslayton: An example of the new, declarative unit test format provided by `mocha-typescript`.

For the curious, it *is* possible to do fancier things like parameterized tests. It would look something like this:

```typescript
import {assert} from 'chai';
import {suite, test} from "mocha-typescript";
import * as util from "../src/util";

[
    {input: undefined, output: false},
    {input: null, output: false},
    {input: 0, output: true},
    {input: 1, output: true},
].forEach(({input, output}) => {
    @suite class GeneratedUtilTest {
        @test(`_hasValue(${input})`) test() {
            assert.equal(util._hasValue(input), output);
        }
    }
});

```

IMHO it's not worth it for small parameter sets like the above because:
1. This involves creating a new `@suite` (`GeneratedUtilTest`) for every test, polluting the testing report.
2. The IDE isn't able to run individual tests via a button press when you use runtime dynamism like this.

It's also possible to directly make calls against the `mocha` object, but it has similar gotchas.

Gruntfile.js
wesboyt: is this the right time to upgrade our grunt dependency as well?

almann: Should we do the same approach of running the unit tests by just invoking it rather than using the plugin?  In other words, if this is our test target from npm, does the grunt integration matter?

zslayton: @wesboyt Our package.json is pointed at `grunt` v1.0.4, which looks to be the latest release. Is that what you were referring to? I'm going to leave any package upgrades for another PR, but let me know what you were thinking.

@almann It's probably not in keeping with the spirit of `grunt`, but it sure seems easier to me. It would cut our dependency list down substantially and eliminate some config duplication. One downside is that it can't be done "natively"--you have to add a grunt package to run child processes without reinventing the wheel, and there are quite a few to choose from. (`grunt-exec`, `grunt-shell`, `grunt-run`, etc.) These have different properties for configuration, color pass-through, and presumably overhead.

PR #440, opened 2019-09-26
Adding mocha tests, big-int resolution changes.

zslayton:
This PR is follow-on work from #439, which added support for Mocha testing. It begins adding Mocha-flavored versions of our old InternJS tests. I'm only adding tests (not removing the old ones), which allows me to break the new tests into more easily-reviewed batches.

*Description of changes:*

This PR:
* Adds Mocha-flavored unit tests for:
  * `Timestamp`
  * `util`
  * `spans`
  * `TextNulls`
  * `Writeable`
* Adds a dependency on the npm module `big-integer`.
* Modifies `LongInt` to point to the node-imported `big-integer` instead of using a local copy.


test/IonTimestamp.ts
```diff
@@ -0,0 +1,231 @@
+/*
```
zslayton: The `timestamp` tests are so heavily parameterized that I fell back to using conventional Mocha syntax (`describe`/`it`) instead of `mocha-typescript`'s declarative syntax (`@suite`/`@test`).

The end result is pretty nice and still benefits from Typescript strictness. For example, the JS version of this file had tests stored in an object keyed by the timestamp text to parse (`{timestampText: testToRun}`). JS allowed repeated keys, which meant that some tests were being overwritten and silently dropped. Typescript required me to change the data structure to preserve them.

As with the `@suite`/`@test` tests in the other files, Webstorm is able to run/debug individual tests/suites from within the browser.

Gruntfile.js
```diff
@@ -91,29 +91,7 @@ module.exports = function(grunt) {
             dest: 'docs/',
           }
         ]
-      },
```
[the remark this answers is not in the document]
almann: :1st_place_medal:

test/IonTimestamp.ts
wesboyt: +1

tsconfig.json
```diff
@@ -10,6 +10,7 @@
   "compilerOptions": {
     "target": "es6",
     "module": "commonjs",
+    "moduleResolution": "node",
```
wesboyt: =)

Issue #444, opened 2019-09-27
Newest version of BigInteger.js breaks decimal -0


zslayton:
Now that we're pointing to npm's latest version of BigInteger.js, this portion of our `decimal` non-equivs test fails:

```
// positive zero and negative zero (with same precision)
[
    0.,
    -0.,
]
```

We're moving off of BigInteger.js, though, so rather than debugging this I'm tracking this problem here. I'll add `non-equivs/decimal` to the skip list for the new Mocha tests until this is fixed and link to this issue in a comment.

zslayton:
Fixed by #465.

Issue #425, opened 2019-09-16
Replace uses of `undefined` in symbol resolution code with a constant


zslayton:
`SharedSymbolTable`, `Import`, and similar return `undefined` when a symbol ID or text cannot be resolved. This is a pretty muddy API--the `undefined` value can be unintuitive to work with. Once we've introduced a proper `SymbolToken` class, we should create an `UNKNOWN_SYMBOL` constant that we can refer to when resolution is impossible.

PR #452, opened 2019-10-02
Continued Mocha test migration.

zslayton:
*Issue #, if available:*

Follow-on work from #451.

*Description of changes:*

Continued migration to mocha. Several of these tests are impacted by #444 and are being skipped for now.


test/IonImport.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  it('Long length pads symbols', () => {
+    let symbolTable = new ion.SharedSymbolTable('foo', 1, ['a', 'b', 'c']);
+    let parent = new ion.Import(null, symbolTable, 4);
+    let child = new ion.Import(parent, symbolTable);
+
+    assert.equal(child.getSymbolText(1), 'a');
+    assert.equal(child.getSymbolText(2), 'b');
+    assert.equal(child.getSymbolText(3), 'c');
+    assert.isUndefined(child.getSymbolText(4));
```
wesboyt: im not saying its in the purview of this pr but maybe we should open an issue to remove undefined from the apis.

zslayton: Agreed -- I think #425 covers that. We need to address that before we add the symbol table machinery to the public API.

PR #453, opened 2019-10-02
Mocha tests for BinaryWriter, Catalog.

zslayton:
Follow-on work from #452.

*Description of changes:*

Mocha tests for `BinaryWriter` and `Catalog`.

### DIFF
--- a/src/IonBinaryWriter.ts
+++ b/src/IonBinaryWriter.ts
@@ -221,7 +221,7 @@ export class BinaryWriter extends AbstractWriter {
 
   writeSymbol(value: string) : void {
     this.checkWriteValue();
-    if (value === null) {
+    if (value === null || value === undefined) {
         this.writeNull(IonTypes.SYMBOL);
     } else {
         let symbolId: number = this.symbolTable.addSymbol(value);
--- a/test/IonBinaryWriter.ts
+++ b/test/IonBinaryWriter.ts
@@ -1,0 +1,1139 @@
+/*
+ * Copyright 2012 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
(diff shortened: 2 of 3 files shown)


src/IonBinaryWriter.ts
```diff
@@ -221,7 +221,7 @@ export class BinaryWriter extends AbstractWriter {
 
   writeSymbol(value: string) : void {
     this.checkWriteValue();
-    if (value === null) {
+    if (value === null || value === undefined) {
```
wesboyt: do we want to allow undefined input here?

test/IonBinaryWriter.ts
```diff
@@ -0,0 +1,1139 @@
...
+        expected: [0x10]
+    },
+    {
+        name: "Writes null boolean by detecting null",
+        instructions: (writer) => writer.writeBoolean(null),
+        expected: [0x1f]
+    },
+    {
+        name: "Writes null boolean by detecting undefined",
+        instructions: (writer) => writer.writeBoolean(undefined),
```
wesboyt: should we open an issue here tracking removal undefined from the boolean api?

src/IonBinaryWriter.ts
zslayton: > do we want to allow undefined input here? I personally would rather throw.

Me too. I've opened [an issue](https://github.com/amzn/ion-js/issues/456) to discuss and track this change.

In the meantime, `writeSymbol` is the only odd-man-out that doesn't accept `undefined`. This required a test to be skipped, so I addressed it here. We can change the behavior across the board all in a future PR.

test/IonBinaryWriter.ts
zslayton: #456 covers this.

Issue #190, opened 2019-05-29
EventStream.writeIon() cant handle null containers


wesboyt:
- /home/travis/build/amzn/ion-js/ion-tests/iontestdata/good/testfile22.ion (39ms)
Error: unexpected type: list

zslayton:
I checked in my local environment: `testfile22.ion` can be safely removed from the skip list. I'll leave this issue open until that change lands.

wesboyt:
needs to PR reverted changes back in

PR #459, opened 2019-10-07
Readers throw in .value() calls whilst positioned on a container type.

wesboyt:
*Issue #, if available:*
https://github.com/amzn/ion-js/issues/190
*Description of changes:*
This PR makes .value() calls on containers throw. It includes updates to both the mocha and intern test suites which allowed us to remove testfile22 from the skiplist which was failing our reader compare tests and causing the eventstream roundtrip logic to throw.


src/IonTextReader.ts
```diff
@@ -328,9 +328,6 @@ export class TextReader implements Reader {
 
     value(): any {
         if (this._type && this._type.isContainer) {
-            if (this.isNull()) {
```
wesboyt: .value() calls now throw on any container

src/IonEventStream.ts
```diff
@@ -154,7 +158,7 @@ export class IonEventStream {
         let currentContainerIndex : number[] = [];
         while(true) {
             if(this.reader.isNull()){
-                this.eventStream.push(this.eventFactory.makeEvent(IonEventType.SCALAR, tid, this.reader.fieldName(), this.reader.depth(), this.reader.annotations(), true, this.reader.value()));
+                this.eventStream.push(this.eventFactory.makeEvent(IonEventType.SCALAR, tid, this.reader.fieldName(), this.reader.depth(), this.reader.annotations(), true, null));
```
dlurton: Can `reader` be on a container type here (i.e. `null.list`) and if so is it valid to make an event with `IonEventType.SCALAR` for container types?

wesboyt: null containers are scalars

dlurton: According to what, exactly?

wesboyt: because I designed it that way.

dlurton: Ok, well, the rest of the changes this PR are acceptable, but calling a `null.list` a scalar value seems wrong to me.  Whether or not a value is a scalar possibly should only be determined by its type, not its type and nulliness.  

I might be being pedantic here, but if is the event stream that is fed into to the compliance test tool, I'd suspect that you'd have to do whatever it expects, and I would be surprised if it considered null containers to be scalar.  

I'm approving this for now in the interest of moving forward.  

Besides, if my hunch is correct you'll have to fix it at a later date anyhow.

Issue #350, opened 2019-09-04
TextReader/writer roundtrip ( '\ a'   '\ \ \ a'   'a' ) using illegal characters.


wesboyt:
This is causing the symbol bugs in equivs.

wesboyt:
Seems like the binary writer should be stripping all quotes including reserved keywords

wesboyt:
needs unit tests verifying all symbols are stripped of quotes before they enter any symboltable.

Issue #348, opened 2019-09-04
Text: the text parser is not removing \ or \n from the input correctly before setting raw


wesboyt:
good/equivs/emptystrings.ion has a few examples and whilst the text roundtrip may be similar the binary output is not. the text parser should be altering the raw value after lexing out the start/end indexes

wesboyt:
merged

PR #461, opened 2019-10-08
String fixes

wesboyt:
*Issue #, if available:*
https://github.com/amzn/ion-js/issues/350
https://github.com/amzn/ion-js/issues/348
Removes equivs-string.ion and emptystrings.ion from the skiplists.
*Description of changes:*
Triple quote strings are currently corrupting the state of the parser when reading escaped terminator characters (single quotes). Values like: '''abc\\'''' are returning abc instead of abc'
This fix flags escaped characters in the parsing loop as non terminating.

It also makes sure characters read from read_escape_sequences() are valid  before adding them to the result string which fixes the escaped newline issues within strings.'


src/IonParserTextRaw.ts
```diff
@@ -1072,13 +1074,9 @@ export class ParserTextRaw {
                         }
                     } else if (this.isLowSurrogate(ch)) {
                         throw new Error("unexpected low surrogate: " + ch);
-                    } else if (t === T_STRING3 && ch === CH_SQ) {
-                        if (this.verifyTriple(index)) {
+                    } else if (t === T_STRING3 && ch === CH_SQ && !bs && this.verifyTriple(index)) {
                             index = this._skip_triple_quote_gap(index, this._end, /*acceptComments*/ true);
-                        } else {
```
wesboyt: this character check skips characters like escaped newlines which return negative values.

src/IonParserTextRaw.ts
```diff
@@ -1072,13 +1074,9 @@ export class ParserTextRaw {
                         }
                     } else if (this.isLowSurrogate(ch)) {
                         throw new Error("unexpected low surrogate: " + ch);
-                    } else if (t === T_STRING3 && ch === CH_SQ) {
```
wesboyt: the important check is !bs, the other alterations are just a simplification.

src/IonParserTextRaw.ts
```diff
@@ -1049,10 +1049,12 @@ export class ParserTextRaw {
             case T_STRING2:
             case T_STRING3:
                 for (index = this._start; index < this._end; index++) {
+                    let bs = false;
```
wesboyt: bs is set to false before each character is read.

src/IonParserTextRaw.ts
```diff
@@ -1072,13 +1074,9 @@ export class ParserTextRaw {
...
                         throw new Error("unexpected low surrogate: " + ch);
-                    } else if (t === T_STRING3 && ch === CH_SQ) {
-                        if (this.verifyTriple(index)) {
+                    } else if (t === T_STRING3 && ch === CH_SQ && !bs && this.verifyTriple(index)) {
                             index = this._skip_triple_quote_gap(index, this._end, /*acceptComments*/ true);
-                        } else {
-                            s += String.fromCharCode(ch);
-                        }
-                    } else {
+                    } else if (ch >= 0) {
```
wesboyt: this character check skips characters like escaped newlines which return negative values.

src/IonParserTextRaw.ts
```diff
@@ -1072,13 +1074,9 @@ export class ParserTextRaw {
                         }
                     } else if (this.isLowSurrogate(ch)) {
                         throw new Error("unexpected low surrogate: " + ch);
-                    } else if (t === T_STRING3 && ch === CH_SQ) {
-                        if (this.verifyTriple(index)) {
+                    } else if (t === T_STRING3 && ch === CH_SQ && !bs && this.verifyTriple(index)) {
```
wesboyt: the important check is !bs, the other alterations are just a simplification.

src/IonParserTextRaw.ts
tgregg: For readability, consider naming `bs` something like `isEscaped` (or, at least, `backslash`).

PR #446, opened 2019-09-27
Depth API addition

wesboyt:
*Description of changes:*
removes toplevel calls, removes toplevel method, adds public depth method to writer interface and writer implementations, adds depth() unit tests.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #462, opened 2019-10-09
Depth fix

wesboyt:
*Issue #, if available:*
Cleans up dirty diff being experienced in: https://github.com/amzn/ion-js/pull/446
*Description of changes:*
adds depth API, unit testing, removes istoplevel in favor of public depth interface.


test/AbstractWriter.ts
```diff
@@ -27,6 +27,71 @@ function testWriteValues(reader, expected) {
...
+    });
+
+    it('Stepping into a sexp results in a depth of 1.', () => {
+        depthTest((writer) => {writer.stepIn(ion.IonTypes.SEXP)}, 1);
+    });
+
+    it('Stepping into 2 lists results in a depth of 2.', () => {
+        depthTest((writer) => {writer.stepIn(ion.IonTypes.LIST); writer.stepIn(ion.IonTypes.LIST)}, 2);
+    });
+
```
almann: *Minor*: I might consider looping through the container types to do this since they are redundant per each.  It does mean you might have to do some separate logic for `STRUCT` (or ignore it as the above does).

Issue #464, opened 2019-10-10
Enable writing arbitrarily sized ints in binary


zslayton:
Recent work on #354 and #111 has enabled reading and writing text BigInts as well as reading binary BigInts. Writing BigInts is possible, but only by converting them to `number`s first, losing precision.

zslayton:
Closed by #465.

Issue #354, opened 2019-09-05
Migrate from custom LongInt class to BigInt


zslayton:
Internally,`ion-js` uses a custom class called `LongInt` to represent integers larger than `(2^53)-1`. `LongInt` is a thin wrapper around a `big-integer.js`'s BigInteger, which is in turn a shim around [the new `BigInt` primitive](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt).

In order [to avoid users taking a dependency on the API offered by `LongInt`](https://github.com/amzn/ion-js/issues/310), we've removed support for Integers that [cannot be stored in 31 bits](https://github.com/amzn/ion-js/issues/210).

`BigInt` is [already supported](https://caniuse.com/#feat=bigint) in many popular JS runtimes--we may be able to migrate to it relatively soon depending on our runtime support requirements.

almann:
I don't think we can directly support native `BigInt` without breaking ES5 compatibility (polyfilling this is apparently [hard](https://github.com/GoogleChromeLabs/jsbi#why)).  We should probably choose a `BigInt` object wrapper and expose that.

Some of the options

* [peterolson/BigInteger.js](https://github.com/peterolson/BigInteger.js) - the one we are presumably using currently.
* [GoogleChromeLabs/JSBI](https://github.com/GoogleChromeLabs/jsbi) - an implementation that is tracking the ECMAScript standard.

nakedible:
I would definitely suggest using JSBI:

- Code matches native BigInt operations 1:1, so future translation is easy
- Can be transpiled to native BigInt by Babel for environments that support it today
- Very good performance
- Written by the Chrome team

In short, the closest you can get to a native BigInt today.

Unless you are willing to break ES5 compatibility, which would be my number one suggestion, in which case native BigInts would be great.

zslayton:
Closed by #465, #469, and #470.

Issue #215, opened 2019-07-08
Convert test suite to Typescript


zslayton:
While all of `ion-js`'s source files are Typescript, all of its tests are written in Javascript. Apart from being inconsistent, this limits the amount of help that IDEs like Webstorm can provide when writing new tests.

zslayton:
#465 introduced our new Typescript + Mocha test suite.

Issue #91, opened 2017-11-02
Transition Intern to V4 to have the library entirely in typescript.


wesboyt:
Intern is now natively typescript, need to transition from js to ts.

zslayton:
#465 introduced our new Typescript + Mocha test suite.

PR #465, opened 2019-10-11
Adds JSBI, removes Intern

zslayton:
*Issue #, if available:* #91, #215, #354, #444, #464 

*Description of changes:*

I recommend looking at the three major commits in this PR independently of one another.

#### Changes
* Added a `bigIntValue(): JSBI` method to the `Reader` interface, allowing users to read arbitrarily large integers.
* `reader.value()` now returns a `JSBI` when the current value is an `int`.
* Creates an `IntSize` enum that indicates whether a value can fit in a `number` or if a `BigInt` is needed.
* Adds an `intSize(): IntSize` method to the `Reader` interface, allowing users to tell whether it's safe to read the current value using `numberValue()`.
* Removes the `LongInt` class and associated tests.
* Rewrites the innards of `Decimal` to use `JSBI` instead of `LongInt`.
* Adds the `JsbiSupport` class, which houses helper functions and static variables that simplify working with JSBI values.
* Adds the `JsbiSerde` class, which contains the data needed to encode and decode `UInt` and `Int` values.
* Removed a couple of dozen tests from the skip lists. 🎉 

#### Follow-on work
* ~~Writing binary BigInts: #464. Currently, writing a JSBI value in binary will cause the value to be turned into a `number`, losing precision. This change will involve adding more functionality to `JsbiSerde` to cover `VarInt` and `VarUInt`.~~ This is addressed in commit 830a14d, described below.
* Rename and expose `Decimal`'s `_getCoefficient(): JSBI` method and `_fromJsbiCoefficient(): JSBI` static constructor as public methods.
* Still need unit tests showing that high-precision Timestamps work as expected.
* The `intSize()` function currently works by reading the value into a `JSBI` and checking to see if it's too large for a `number`. This is somewhat self-defeating, as you won't be saving any effort by calling `intSize()` and then `numberValue()`. The interface allows for a more efficient implementation to be added down the road.

#### Changes
* Removes all of the intern.js tests. (~4,000 lines deleted!)
* Removes our dependencies on:
  * `intern`
  * `grunt-mocha-test`
  * `big-integer.js`
* Deletes our local copy of `big-integer.js`. (This raised our test coverage by quite a bit!)
* Adds `mocha` to the build targets.
* We're now testing using `ts-node`, which allows us to run our test suite without first compiling all of our build targets. This PR re-orders the build targets to run the tests (~9 seconds) before compiling all build targets (another ~11 seconds), allowing you to see if something's not working much quicker.

#### Changes
* JSBI values are no longer clamped to `Number`'s safe integer range.
* `IonBinaryWriter#writeInt` and all downstream methods can accept either a `number` or a `JSBI`.
* Some of the `ion-tests` construct absurdly large numbers that can take thousands of bytes to encode as a `UInt`. Previously, `JsbiSerde` would bomb out if asked to serialize an integer that required more than a given size in bytes. Now it will accommodate any size, while taking steps to make sure that the more common, smaller-sized integers are reasonably fast.
* Removed 19 more `ion-tests` from the skip list

#### Follow-on work
* We should consider creating a type alias for `number | JSBI`.
* There's plenty of room for optimization in `JsbiSerde`'s `getUnsignedIntSizeInBytes` method.


src/IonParserBinaryRaw.ts
```diff
@@ -208,74 +211,26 @@ export class ParserBinaryRaw {
...
-            byte = this._in.next();
-            stopBit = byte & 0x80;
-            byte &= 0x7F;
-            bytes.push(byte);
-        }
-        return LongInt.fromVarIntBytes(bytes, isNegative);
-    }
-
-    static _readSignedIntFrom(input: BinarySpan, numberOfBytes: number) : LongInt {
+    static _readSignedIntFrom(input: BinarySpan, numberOfBytes: number) : SignAndMagnitudeInt {
```
zslayton: `SignAndMagnitudeInt` is a simple class that pairs a `magnitude: JSBI` with an `isNegative: boolean` for the sake of returning a single value representing a signed int that might be `-0`. It is only used internally.

src/IonParserBinaryRaw.ts
```diff
@@ -208,74 +211,26 @@ export class ParserBinaryRaw {
...
-            return 0;
-
-        while (bytesRead < numberOfBytes) {
-            byte = input.next();
-            bytesRead++;
-            value = value << 8;
-            value = value | (byte & 0xff);
-        }
-
-        // We overflowed
```
zslayton: I was able to delete this section because we no longer overflow. 💃

test/IonBinaryWriter.ts
```diff
@@ -237,7 +237,6 @@ let decimalWriterTests = [
     },
     {
         name: "Writes negative zero decimal",
-        skip: true, // FIXME: https://github.com/amzn/ion-js/issues/444
```
zslayton: Quick link to #444. We've removed Intern, so this is no longer an issue.

src/IonBinaryReader.ts
```diff
@@ -18,22 +18,20 @@
...
+import {IonType} from "./IonType";
+import {IonTypes} from "./IonTypes";
+import {IVM} from "./IonConstants";
+import {ParserBinaryRaw} from "./IonParserBinaryRaw";
+import {Reader} from "./IonReader";
+import {BinarySpan} from "./IonSpan";
+import {Timestamp} from "./IonTimestamp";
+import JSBI from "jsbi";
+import IntSize from "./IntSize";
+import {JsbiSupport} from "./JsbiSupport";
```
therapon: This seems like formatting changes. Did we pin down a formatter? It would be nice to avoid these sort of changes in the diffs.

src/IonBinaryReader.ts
```diff
@@ -176,6 +174,17 @@ export class BinaryReader implements Reader {
     return this._parser.decimalValue();
   }
 
+  bigIntValue() : JSBI {
```
therapon: I forget, is this method going to be private? 

If this method is exposed does this mean that we now have JSBI in our external facing API? 
Is this the long term goal? If not should we wrap it with our own interface such that we can swap out implementations of BigInteger to shield us from client braking changes?

src/IonBinaryWriter.ts
```diff
@@ -194,13 +197,18 @@ export class BinaryWriter extends AbstractWriter {
     this.addNode(new BytesNode(this.writer, this.getCurrentContainer(), IonTypes.FLOAT, this.encodeAnnotations(this._annotations), bytes));
   }
 
-  writeInt(value: number) : void {
+  writeInt(value: number | JSBI | null) : void {
```
therapon: perhaps a typealias for `value`?

src/IonDecimal.ts
```diff
@@ -12,8 +12,9 @@
  * language governing permissions and limitations under the License.
  */
 
-import {LongInt} from "./IonLongInt";
+import JSBI from 'jsbi';
```
therapon: nit: be consistent on the use of single VS double quotes

src/IonReader.ts
```diff
@@ -120,10 +122,27 @@ export interface Reader {
...
     numberValue(): number | null;
 
+    /**
+     * Returns the current value as a `BigInt`.  This is only valid if `type() == IonTypes.INT`.
+     *
+     * @return `null` if the current Ion value [[isNull]] or a BigInt containing the deserialized integer.
+     *
+     * @throw Error when the reader is not positioned on an `int` typed value.
+     */
+    bigIntValue(): JSBI | null;
```
therapon: Aha so we are exposing it through our API.

src/JsbiSupport.ts
```diff
@@ -0,0 +1,92 @@
...
+ * A copy of the License is located at:
+ *
+ *     http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ */
+
+import JSBI from 'jsbi';
```
therapon: nit: usage of single quotes

src/JsbiSupport.ts
```diff
@@ -0,0 +1,92 @@
...
+     */
+    public static clampToSafeIntegerRange(value: JSBI): number {
+        if (JSBI.greaterThan(value, this.NUMBER_MAX_SAFE_INTEGER)) {
+            return Number.MAX_SAFE_INTEGER;
+        }
+        if (JSBI.lessThan(value, this.NUMBER_MIN_SAFE_INTEGER)) {
+            return Number.MIN_SAFE_INTEGER;
+        }
+        return JSBI.toNumber(value);
+    }
```
therapon: minor: I would add a third if for the case that now falls through to the last return. Use the static method below to see that the input is in the valid range and return. I assume this would be the most common case so it would be better to rub one if test rather than 2.

src/SignAndMagnitudeInt.ts
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Represents a signed, arbitrarily sized integer.
+ *
+ * BigInts cannot represent negative zero. This class should be used in situations where negative zero is
+ * a supported value, such as when decoding binary Ion Int/VarInt.
+ *
+ * http://amzn.github.io/ion-docs/docs/binary.html#uint-and-int-fields
+ *
+ */
+export default class SignAndMagnitudeInt {
```
therapon: A comment earlier on in this PR said that this class is for internal use only, Should we prefix the name with `_`?

test/IonDecimal.ts
```diff
@@ -32,8 +34,10 @@ function test(decimalString,
 
     let decimal = ion.Decimal.parse(decimalString);
 
-    assert.deepEqual(decimal._getCoefficient(), new ion.LongInt(expectedCoefficient), '_getCoefficient()');
-    assert.equal(decimal._getCoefficient().signum(), new ion.LongInt(expectedCoefficient).signum(), 'coefficient sign');
+    // assert.deepEqual(decimal._getCoefficient(), new ion.LongInt(expectedCoefficient), '_getCoefficient()');
+    // assert.equal(decimal._getCoefficient().signum(), new ion.LongInt(expectedCoefficient).signum(), 'coefficient sign');
```
therapon: remove?

test/IonDecimal.ts
```diff
@@ -171,7 +175,7 @@ let decimalComparisonTests = [
 ];
 
 let decimalEqualsTests = [
-    {input1: '0', input2: '0', expected: true},
+    {input1: '0', input2: '0', expected: true, skip: false},
```
therapon: Not sure why we have `skip: false` here? Could you add a comment.

src/IonBinaryReader.ts
zslayton: > This seems like formatting changes. Did we pin down a formatter?

We were holding off on selecting a formatter until we had migrated the tests to Typescript, which this PR completes. This change, however, came from my opinionated IDE and was not intentional. 

>  It would be nice to avoid these sort of changes in the diffs.

Agreed, sorry about that.

src/IonBinaryReader.ts
zslayton: > I forget, is this method going to be private?

It's meant to be public to give users access to ints greater than `(2^53)-1`.

> If this method is exposed does this mean that we now have JSBI in our external facing API? Is this the long term goal?

The long term goal is to move from JSBI to using [proper `BigInt`s](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt), which are already broadly (but not universally) supported by JS runtimes. JSBI gives us the ability to [add a babel build step](https://github.com/GoogleChromeLabs/babel-plugin-transform-jsbi-to-bigint) that will convert all of our JSBI static method uses to `BigInt` syntax when we're ready. Then customers can decide whether they want native `BigInt`s or the JSBI shim. See [the "Why?" section](https://github.com/GoogleChromeLabs/jsbi/#why) of the JSBI repo for more details.

> [...] should we wrap it with our own interface such that we can swap out implementations of BigInteger to shield us from client braking changes?

JSBI strictly adheres to [the BigInt TC39 proposal](https://tc39.es/proposal-bigint/), which moved to [stage 4 ("Final")](https://tc39.es/process-document/) last month, so breaking changes are very unlikely.

src/IonBinaryWriter.ts
zslayton: What alias would you suggest? I'm a little afraid to introduce an important-sounding type named `Integer` or `IonInteger`.

src/IonDecimal.ts
zslayton: Ah, sure. I happen to like single quotes for imports, but I'll switch them for consistency's sake.

src/SignAndMagnitudeInt.ts
zslayton: We haven't been using underscore prefixes for internal classes thus far, just for member data / functions. Only [a handful of our classes](https://github.com/amzn/ion-js/blob/master/src/Ion.ts#L89-L97) are part of the public API.

test/IonDecimal.ts
zslayton: Fixed, thanks.

test/IonDecimal.ts
zslayton: Typescript is able to infer a (pretty verbose) data type for `decimalEqualsTests` based on the fields/types found in the object literals in the array declaration.

Previously, some of the tests needed to be skipped, so they included a `skip: true` field. This PR obviated the need to skip them, but when I removed `skip: true`, I found that Typescript would complain about `skip: boolean` not being part of `decimalEqualsTests`'s data type when I used object destructuring:

```typescript
    describe('equals', () => {
        decimalEqualsTests.forEach(
            ({input1, input2, expected, skip = false}) => { // <--- Object destructuring
               // ...
                if(skip) {
                    // ...
                } else {
                    // ...
                }
            }
        );
    });
```

Adding `skip: false` to the first test as an example fixed the issue.

It's difficult to find documentation that describes this combination of behaviors that I could link to in a comment, and as you can see it takes a fair bit of text to explain the issue. 😛 I've opted to explicitly declare `decimalEqualsTests`' data type, allowing me to remove `skip: false`.

```typescript
let decimalEqualsTests: ({input1: string, input2: string, expected: boolean, skip?: boolean})[] = [
    // ...
];
```
Hopefully that's a bit more self-explanatory.

src/JsbiSupport.ts
zslayton: Do you mean something like this?

```typescript
if (JsbiSupport.isSafeInteger(value)) { // Involves 2 tests, a GTE(MIN) && LTE(MAX)
  return JSBI.toNumber(value);
}
if (JSBI.greaterThan(value, this.NUMBER_MAX_SAFE_INTEGER)) {
  return Number.MAX_SAFE_INTEGER;
}
if (JSBI.greaterThan(value, this.NUMBER_MAX_SAFE_INTEGER)) {
  return Number.MAX_SAFE_INTEGER;
}
// Unreachable, I guess we would throw?
```

If so, I'm unsure I understand the benefit. We would end up needing to compare the value against the `MIN` and `MAX` two times instead of just once, and the code is less intuitive to me. Have I misunderstood?

therapon: Ignore me, I was not thinking.

src/IonBinaryWriter.ts
therapon: By type alias I mean http://www.typescriptlang.org/docs/handbook/advanced-types.html#type-aliases

Mostly a rename to aid readability, `MaybeInt` -- not sure if there is a `Maybe<X>` or `Option<X>` and how it is implemented but generics might make things more complex

test/IonDecimal.ts
therapon: This seems to be an issue with Typescripts inference engine. Seems like a bug to me if the inferred type relies on the first instance of a value and not the rest. I would expect a union type for the last field `skip`.

Explicitly pinning the type for the variable is the sane thing to do here. 
Thanks

src/IonTextWriter.ts
```diff
@@ -216,8 +218,8 @@ export class TextWriter extends AbstractWriter {
         writer.writeUtf8(text);
     }
 
-    writeInt(value: number) : void {
-        this._serializeValue(IonTypes.INT, value, (value: number) => {
+    writeInt(value: number | JSBI) : void {
+        this._serializeValue(IonTypes.INT, value, (value: number | JSBI) => {
```
almann: This can be `null` right?

src/IonBinaryWriter.ts
```diff
@@ -18,11 +18,13 @@ import {Import} from "./IonImport";
 import {IonType} from "./IonType";
 import {IonTypes} from "./IonTypes";
 import {LocalSymbolTable} from "./IonLocalSymbolTable";
-import {LongInt} from "./IonLongInt";
 import {LowLevelBinaryWriter} from "./IonLowLevelBinaryWriter";
 import {TimestampPrecision,Timestamp} from "./IonTimestamp";
 import {Writeable} from "./IonWriteable";
 import {_sign} from "./util";
+import JSBI from "jsbi";
```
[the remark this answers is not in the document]
zslayton: It means something different.
```typescript
import JSBI from "jsbi";
```
means that `JSBI` is the `default export` in that module. The naming is irrelevant, I could write:
```typescript
import walnuts from "jsbi";
```
and `walnuts` would be the same value as `JSBI` was in the prior example.

If something is the default export, it can't be imported selectively.
```
import {JSBI} from "jsbi";
```
requires that there be a non-default exported value in `"jsbi"` that is named `JSBI`. Running this results in:
```
TS2305: Module ".../jsbi" has no exported member 'JSBI'.
```

nakedible:
A couple comments, ignore if not relevant:

- Should there be an API to get BigInts as strings and to write BigInts from strings? Since not everyone uses JSBI, strings are the common denominator between different implementations. Base-10 always, `0|-?[1-9][0-9]*` format.
- Should this pull add IonDecimal constructors and accessors for BigInts? Without that, there still is no way to access the accurate value of IonDecimals or produce such values.
- Can IonDecimal **exponent** even theoretically be larger than JavaScript integers? I found no mention of limitations in the Ion specification. This would matter for example if calculating Ion Hash for documents containing such values - and could produce a security vulnerability if one implementation calculates a different hash or crashes with such documents.

wesboyt:
-We've limited exponent to number by design but theres nothing in the specification limiting it.

src/IonParserBinaryRaw.ts
```diff
@@ -208,74 +211,26 @@ export class ParserBinaryRaw {
...
-                + " Use readUnsignedLongIntFrom instead. UInt size: " + numberOfBytes + ", value: " + value
-            );
-        }
-        // Fewer bytes than the required `numberOfBytes` were available in the input
-        if (byte === EOF) {
-            throw new Error("Ran out of data while reading a " + numberOfBytes + "-byte unsigned int.");
-        }
-
-        return value;
+    static _readUnsignedIntFrom(input: BinarySpan, numberOfBytes: number) : JSBI {
```
wesboyt: We should not be replacing the lower level ints with JSBI we should only be creating JSBI values for decimal/timestamp decimals and Ion INTs lower level values should still be consumed as numbers.

src/IonTextWriter.ts
zslayton: It can. By default, Typescript treats `null` and `undefined` as [subtypes of all other types](https://www.typescriptlang.org/docs/handbook/basic-types.html#null-and-undefined), meaning they're allowed even if the type signature of a method doesn't accept it explicitly.

While the `Writer` interface's methods spell out where `null` is acceptable, the implementations don't for some reason.

You can opt into package-wide type checking that requires methods to explicitly allow `null`s via the `--strictNullChecks` flag. I'd be for doing this, but it would be a sweeping change better suited to its own PR.

src/IonParserBinaryRaw.ts
zslayton: As discussed offline, the `UInt` and `Int` primitives are only used [in the `int`, `decimal`, and `symbol` types](http://amzn.github.io/ion-docs/docs/binary.html#uint-and-int-fields) (and `timestamp`, which has a `decimal` field), so this isn't as far-reaching a change as it seemed at first glance.

zslayton:
> Should this pull add IonDecimal constructors and accessors for BigInts? 
> Without that, there still is no way to access the accurate value of IonDecimals or produce such values.

Yep, you're right. I mentioned this briefly in the "Follow on work" section of the PR description:

> Rename and expose Decimal's _getCoefficient(): JSBI method and _fromJsbiCoefficient(): JSBI static constructor as public methods.

I'll have another PR out for this soon, likely Monday.

nakedible:
@wesboyt 
> -We've limited exponent to number by design but theres nothing in the specification limiting it.

I added an issue about this here https://github.com/amzn/ion-tests/issues/50. I'm not sure if there's a better place for it. Again, feel free to ignore and close if you believe this to be a non-issue.

Issue #419, opened 2019-09-14
Decimal interface does not give access to coefficient and exponent


nakedible:
The `Decimal` interface does not give the user any access to the coefficient and exponent values. The only options are `intValue`, which truncates fractional portion and `numberValue` which is inexact. This means that it is impossible to accurately decipher `Decimal` values for any users.

Obviously BigNum issues should be solved first as both coefficient and exponent (theoretically) may be larger than what safely can be represented in a number.

jonwilsdon:
Related to #354

zslayton:
Closed by #469 and #470.

PR #469, opened 2019-10-15
Decimal constructors for number, JSBI, and string.

zslayton:
*Issue #, if available:* #354, #419 

*Description of changes:*

`Decimal`'s constructor now supports three sets of parameters:

```typescript
// Parsing a string (same behavior as Decimal.parse)
new Decimal(text: string)

// Providing an explicit coefficient and exponent as numbers
new Decimal(coefficient: number, exponent: number)

// Providing a BigInt coefficient, a number exponent, 
// and an optional explicit sign to accommodate -0.
new Decimal(coefficient: number, exponent: number, isNegative?: boolean)
```

### DIFF
--- a/src/IonDecimal.ts
+++ b/src/IonDecimal.ts
@@ -13,7 +13,7 @@
  */
 
 import JSBI from "jsbi";
-import {_sign} from "./util";
+import {_hasValue, _sign} from "./util";
 import {JsbiSupport} from "./JsbiSupport";
 
 /**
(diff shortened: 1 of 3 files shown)


src/IonDecimal.ts
```diff
@@ -49,27 +49,68 @@ export class Decimal {
...
+     *      The provided text will be parsed into a Decimal value.
+     *
+     *  (coefficient: number, coefficient: number)
+     *      A Decimal value will be constructed using the provided coefficient and exponent values.
+     *
+     *  (coefficient: BigInt, coefficient: number, isNegative?: boolean)
+     *      A Decimal value will be constructed using the provided coefficient and exponent values.
+     *      Because the BigInt data type cannot represent -0, a third parameter called `isNegative` may be supplied
+     *      to explicitly set the sign of the Decimal following construction. If this flag is not specified,
+     *      the provided coefficient's sign will be used.
```
almann: Technically aren't both variants `number | JSBI` for the coefficient?

What is the implication of a negative coefficient and `isNegative` being in conflict or not with that?  Is the sign of the coefficient ignored--that seems to be the implication, but I would be explicit.

src/IonDecimal.ts
```diff
@@ -49,27 +49,68 @@ export class Decimal {
...
+            return Decimal.parse(coefficient);
+        }
+
+        if (!_hasValue(exponent)) {
+            throw new Error("Decimal's constructor was called with a numeric coefficient but no exponent.");
+        }
+
+        if (typeof coefficient === "number") {
+            return Decimal._fromNumberCoefficient(coefficient, exponent);
+        }
```
almann: Oh interesting, what is supposed to happen if `number` is a coefficient and `isNegative` is set?

src/IonDecimal.ts
```diff
@@ -49,27 +49,68 @@ export class Decimal {
...
+        if(coefficient instanceof JSBI) {
+            if (!_hasValue(isNegative)) {
+                // If isNegative was not specified, infer isNegative from the coefficient.
+                // This will work for all values except -0.
+                isNegative = JsbiSupport.isNegative(coefficient);
+            } else if (isNegative && !JsbiSupport.isNegative(coefficient)) {
+                // If isNegative was specified, make sure that the coefficent's sign agrees with it.
+                coefficient = JSBI.unaryMinus(coefficient);
+            }
+            return Decimal._fromBigIntCoefficient(isNegative, coefficient, exponent);
```
almann: What about `! isNegative` and `JsbiSupport.isNegative(coefficient)`?  Is this a valid state?

src/IonDecimal.ts
zslayton: An unfortunate consequence of Javascript/Typescript not supporting constructor overloading is that all of the parameter sets get intermingled.

> Technically aren't both variants number | JSBI for the coefficient?

I would say they're not because the code path for `(number, number)` will ignore a third parameter (`isNegative`) while the code path for `(BigInt, number, isNegative?)` will respect `isNegative` when it's supplied.

> What is the implication of a negative coefficient and isNegative being in conflict or not with that? Is the sign of the coefficient ignored--that seems to be the implication, but I would be explicit.

Good point. The sign of the coefficient will be overridden by `isNegative` if specified. I'll add that to the documentation, thanks.

src/IonDecimal.ts
```diff
@@ -49,27 +49,68 @@ export class Decimal {
...
     /**
-     * Creates a new Decimal value.
-     * @param coefficient   See the {@link Decimal} class-level documentation for details.
-     * @param exponent      See the {@link Decimal} class-level documentation for details.
+     * Creates a new Decimal value. Three parameter sets are supported:
+     *
+     *  (text: string)
+     *      The provided text will be parsed into a Decimal value.
+     *
+     *  (coefficient: number, coefficient: number)
```
pbcornell: --> `(coefficient: number, exponent: number)`

src/IonDecimal.ts
```diff
@@ -49,27 +49,68 @@ export class Decimal {
...
+     *      A Decimal value will be constructed using the provided coefficient and exponent values.
+     *
+     *  (coefficient: BigInt, coefficient: number, isNegative?: boolean)
```
pbcornell: --> `(coefficient: BigInt, exponent: number, ...)`

src/IonDecimal.ts
zslayton: Thanks, fixed!

src/IonDecimal.ts
zslayton: Fixed.

src/IonDecimal.ts
zslayton: I just learned that Typescript allows you to declare which subsets of a signature's possible parameter combinations are permitted with the following syntax:
```typescript
    // Publicly visible/usable signatures
    constructor(coefficient: string);
    constructor(coefficient: number, exponent: number);
    constructor(coefficient: JSBI, exponent: number, isNegative?: boolean);

    // Consolidated, non-public implementation signature
    constructor(coefficient: number | JSBI | string, exponent?: number, isNegative?: boolean) {
```

I'm going to take advantage of this feature to eliminate the usage of unintended combinations.

wesboyt:
looks good to me, I'm not a huge fan of relying on isnegative and the coefficient as a fall back but don't see a way around it.

PR #473, opened 2019-10-16
Fixes bug for writeInt(BigInt('0')), exports IntSize type, updates ReaderScalarValue type

### LINKED ISSUE #471: binary writer produces incorrect output for writeInt(BigInt('0'))
```javascript
let w = ion.makeBinaryWriter();
w.writeInt(JSBI.BigInt('0'));
w.close();
console.log('bytes: ' + toHexString(w.getBytes()));

function toHexString(byteArray) {
    let sb = '';
    byteArray.forEach(b => {
        if (sb != '') { sb += ' ' }
        sb += ('0' + (b & 0xFF).toString(16)).slice(-2);
    });
    return sb;
}
```
produces:
```
bytes: e0 01 00 ea 30
```

but should produce: `e0 01 00 ea 20`

pbcornell:
Resolves #471, exports the new IntSize type, and updates the ReaderScalarValue type.


src/IonBinaryWriter.ts
```diff
@@ -654,7 +654,7 @@ class IntNode extends LeafNode {
   constructor(writer: LowLevelBinaryWriter, parent: Node, annotations: Uint8Array, private readonly value: number | JSBI) {
     super(writer, parent, IonTypes.INT, annotations);
 
-    if (this.value === 0) {
+    if (this.value == 0) {
```
wesboyt: JSBI(0) == "0" -> true
" " == 0 -> true
false == 0 -> true
[] == 0 -> true
We should be checking the input type instead.

almann: We should be using `JSBI.equals()` or `JSBI.EQ()` for `===` and `==` respectively.  That said, I agree here that `==` (and its equivalent `JSBI.EQ()`) is not quite right.  What about:

```typescript
this.value === 0 || JsbiSupport.isZero(this.value)
```

Maybe rolling the above into a utility method might be worthwhile.

pbcornell: Today, this appears to be only place where we perform a problematic `===` comparison.  I've updated this PR to avoid it altogether via a @zslayton gambit:  `0` is now handled in the else block.  Avoids a lengthy condition, as well as any discussion around where such utility method should reside.

That said, we'd be wise to build a little internal infrastructure around number/JSBI to avoid pitfalls like this in the future, as well as only allocating a JSBI when doing so adds value.

src/IonBinaryWriter.ts
```diff
@@ -654,18 +654,15 @@ class IntNode extends LeafNode {
...
       this.bytes = writer.getBytes();
-    } else {
+
+    } else if (this.value < 0) {
       this.intTypeCode = TypeCodes.NEGATIVE_INT;
       let magnitude: number | JSBI;
-      let uintSize: number;
       if (value instanceof JSBI) {
         if(JsbiSupport.isNegative(value)) {
           magnitude = JSBI.unaryMinus(value);
```
almann: I am pretty sure we should be using `JSBI.GT` and `JSBI.LT` instead of the operator.  In fact, I am a bit surprised that this works, do we have tests that invoke this code path with `JSBI` values?

Should the preamble of this method check for `typeof this.value === 'number' || this.value isinstance JSBI` as per Wes's feedback to avoid strange behavior for abstract  comparisons?

pbcornell: I've added binary `writeInt` tests that pass JSBI values.  These new tests pass locally when the implementation uses `>`/`<` operators as well as `JSBI.GT`/`JSBI.LT` calls; not sure why.  I agree the latter approach feels safer.

I've also added a check to assert that `this.value` is a number or JSBI; we might want to have a discussion about whether similar assertions would make sense in other parts of the impl.

src/IonBinaryWriter.ts
```diff
@@ -654,13 +654,17 @@ class IntNode extends LeafNode {
   constructor(writer: LowLevelBinaryWriter, parent: Node, annotations: Uint8Array, private readonly value: number | JSBI) {
     super(writer, parent, IonTypes.INT, annotations);
 
-    if (this.value > 0) {
+    if (!(typeof this.value === 'number' || this.value instanceof JSBI)) {
+      throw new Error('Expected ' + this.value + ' to be a number or JSBI');
+    }
+
+    if (JSBI.GT(this.value, 0)) {
```
zslayton: This works fine, but:
* Using `JSBI.GT` will construct `JSBI` values from `number` values that it wishes to compare. That's a pretty high overhead for a simple `>` or `<` test.
* You can pass the static value `JsbiSupport.ZERO` instead of `0` to avoid instantiating `JSBI(0)` each time.

How about lifting the comparisons out?
```typescript
let isGreaterThanZero: boolean;
let isLessThanZero: boolean;
if (typeof value === 'number') {
   isGreaterThanZero = value > 0;
   isLessThanZero = value < 0;
} else if(value instanceof JSBI) {
   isGreaterThanZero = JsbiSupport.greaterThan(value, JsbiSupport.ZERO);
   isLessThanZero = JsbiSupport.greaterThan(value, JsbiSupport.ZERO);
} else {
  throw new Error('Expected ' + this.value + ' to be a number or JSBI');
}

if (isGreaterThanZero) {
  //...
} else if (isLessThanZero) {
  //...
} else {
  //...
}
```
I realize this is a larger restructuring than we had hoped for, but it's very explicit and avoids unnecessary overhead. It also folds the `assert` into the type checking.

pbcornell: > Using JSBI.GT will construct JSBI values from number values that it wishes to compare.

@wesboyt made this observation the other day as well.  However, the logic of [JSBI.__compare()](https://github.com/GoogleChromeLabs/jsbi/blob/master/jsbi.mjs#L1077) indicates that this is not the case.  The JSBI source we depend on is a bit different, but doesn't appear to construct new JSBIs for ease-of-comparison either.

A case could be made that our code can perform less type-checking logic than JSBI internals, but I'm not sure that's warranted here (esp. in the absence of performance data).  If we're going to make such changes, let's make sure we're clear about *why* we're doing so.

Based on these findings, my inclination is to not lift the comparisons out as you've proposed.  Let me know if you still feel doing so would add clarity and/or material performance improvements.

zslayton: > However, the logic of [JSBI.__compare()](https://github.com/GoogleChromeLabs/jsbi/blob/master/jsbi.mjs#L1077) indicates that this is not the case. The JSBI source we depend on is a bit different, but doesn't appear to construct new JSBIs for ease-of-comparison either.

I stand corrected! It looks like the `(number, number)` case boils down to a number of type checks followed by [the expected operators](https://github.com/GoogleChromeLabs/jsbi/blob/master/jsbi.mjs#L1114-L1119).

> A case could be made that our code can perform less type-checking logic than JSBI internals, but I'm not sure that's warranted here (esp. in the absence of performance data). If we're going to make such changes, let's make sure we're clear about why we're doing so.

Agreed.

> Based on these findings, my inclination is to not lift the comparisons out as you've proposed. Let me know if you still feel doing so would add clarity and/or material performance improvements.

Your current code is sensible. Let's stay the course until we have performance data.

Thanks for looking into `JSBI.__compare()`!

PR #482, opened 2019-10-23
Removes text parser logic that performed UTF-16 and UTF-8 encoding on clob bytes

### LINKED ISSUE #481: bytes produced by the text reader for {{"\xc2\x80"}} don't match expectations
```javascript
let reader = ion.makeReader('{{"\\xc2\\x80"}}');
reader.next();
console.log('bytes: ' + toHexString(reader.byteValue()));

function toHexString(byteArray) {
    let sb = '';
    byteArray.forEach(b => {
        if (sb != '') { sb += ' ' }
        sb += ('0' + (b & 0xFF).toString(16)).slice(-2);
    });
    return sb;
}
```
produces:
```bytes: c3 82 c2 80```
but should produce:
```bytes: c2 80```

### PR DESCRIPTION
**Pull request by pbcornell**
The text reader currently converts clob bytes to a string via `String.fromCharCode` (which applies UTF-16 encoding), then `byteValue()` converts that string UTF-8 bytes.  This implementation is incorrect.

An Ion clob value does have any particular encoding--it is just a collection of bytes intended to represent text.  Any encoding is an application-level concern, and encoding should thereby *not* be performed by an Ion parser implementation.

Resolves #481 

### DIFF
--- a/src/IonParserTextRaw.ts
+++ b/src/IonParserTextRaw.ts
@@ -1035,7 +1035,6 @@ export class ParserTextRaw {
     get_value_as_string(t: number) : string {
         let index : number;
         let ch : number;
-        let escaped : number;
         let s : string = "";
         switch (t) {
             case T_NULL:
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/IonParserTextRaw.ts [hunk 1]
```diff
@@ -1096,45 +1095,54 @@ export class ParserTextRaw {
...
+            default:
+                throw new Error("can't get this value as a string");
+        }
+        return s;
+    }
+
+    get_value_as_uint8array(t: number) : Uint8Array {
+        let bytes = [];
+        switch (t) {
             case T_CLOB2:
```
zslayton: Seems like there's an opportunity to consolidate the implementations of the `T_CLOB2` and `T_CLOB3` cases a fair amount in the future.

pbcornell: Potentially; leaving it up to the future for now

Issue #478, opened 2019-10-22
Text reader doesn't throw upon encountering the annotation '$0'


pbcornell:
`Reader.value()` throws on `$0`, `Reader.fieldName()` throws on `{$0: 1}`, but `Reader.annotations()` *doesn't* throw on `$0::1`.  Until SID 0 is supported, I suspect it should.

```javascript
readValue('$0');
readValue('{$0: 1}');
readValue('$0::1');

function readValue(s) {
    try {
        let r = ion.makeReader(s);
        r.next();
        if (r.type() === ion.IonTypes.STRUCT) {
            r.stepIn();
            r.next();
        }
        r.value();        // throws on $0
        r.fieldName();    // throws on {$0: 1}
        r.annotations();  // doesn't throw on $0::1
    } catch (e) {
        console.log('Error on ' + s + ': ' + e.message);
    }
}
```
produces:
```
Error on $0: Unresolvable symbol ID, symboltokens unsupported.
Error on {$0:1}: Unresolveable symbol ID, symboltokens unsupported.
```

Also note that "Unresolveable" is incorrect spelling.

wesboyt:
fixed

PR #486, opened 2019-10-28
Sid0 fix

wesboyt:
*Issue #, if available:*
https://github.com/amzn/ion-js/issues/478
*Description of changes:*
-Adds unit test for reader encountering annotation with sid0
-Alters binary parser to use null instead of -1 on missing field IDs
-Textparser now throws when it encounters a symbol identifier of $0
-Adds sidzero testfiles to the skiplist.


test/IonReader.ts
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific
+ * language governing permissions and limitations under the License.
+ */
+
+import {assert} from 'chai';
+import * as ion from '../src/IonTests';
+
+describe('Annotation tests', () => {
```
zslayton: It seems like this test file could be named something a bit narrower in scope than `IonReader`--maybe `test/annotations.ts`?

src/IonParserBinaryRaw.ts
```diff
@@ -494,6 +494,7 @@ export class ParserBinaryRaw {
             a = (a << VINT_SHIFT) | (b & VINT_MASK);  // OR in the 7 useful bits
             if ((b & VINT_FLAG) !== 0) {
                 // once we have the last byte, add it to our list and start the next
+                if(a === 0) throw new Error('Symbol ID zero is unsupported.');
```
zslayton: Matter of preference: I like having the contents of an `if` statement wrapped in braces and on the following line(s), even if it's only a single statement. If WebStorm's code formatter doesn't mind though, I'm happy to leave it as is.

PR #512, opened 2019-11-19
Adds tslint, tslint:src, and tslint:test build targets

### LINKED ISSUE #345: add a linter to the build process
For example:  [tslint](https://palantir.github.io/tslint/)

pbcornell:
Initial config simply uses tslint's "recommended" set of rules.  Some rules come with "fixers" that modify the source to align with the rule's notion of "correct."  I'd suggest we configure and apply fixers as time and inclination permits.

Resolves #345 

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -24,9 +24,9 @@ module.exports = function(grunt) {
     jshint: {
       files: []
     },
-      /**
-       * Typescript document generator
-       */
+    /**
+     * Typescript document generator
+     */
     typedoc: {
       build: {
         src: 'src/**/*'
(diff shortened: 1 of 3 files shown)


Gruntfile.js
```diff
@@ -148,6 +153,7 @@ module.exports = function(grunt) {
   grunt.loadNpmTasks('grunt-contrib-uglify');
   grunt.loadNpmTasks('grunt-shell');
   grunt.loadNpmTasks('grunt-ts');
+  grunt.loadNpmTasks('grunt-tslint');
```
zslayton: How does one kick off a `tslint` run given this config? I expected a build target or npm script to be required.

pbcornell: The additions within the `grunt.initConfig` section define the targets.  With these changes, `grunt tslint:src`, `grunt tslint:test`, and `grunt tslint:all` (all source) are all valid.

PR #517, opened 2019-11-26
Updates IonWriter method implementations to throw if value is undefined

### LINKED ISSUE #456: writeType(...) methods should not accept undefined
`Writer`'s `writeType` methods ([`writeTimestamp`](https://github.com/amzn/ion-js/blob/31871dc0e9404bef52037ae22c13d8aa49839362/src/IonBinaryWriter.ts#L236), [`writeString`](https://github.com/amzn/ion-js/blob/31871dc0e9404bef52037ae22c13d8aa49839362/src/IonBinaryWriter.ts#L214), etc) all accept `undefined` as a synonym for `null`. This means that calling `writeInt()` (with no parameters) produces `null.int`, which is rather surprising. Passing an `undefined` should probably cause an `Error` to be thrown.

pbcornell:
Resolves #456 

### DIFF
--- a/src/IonBinaryWriter.ts
+++ b/src/IonBinaryWriter.ts
@@ -98,7 +98,7 @@ export class BinaryWriter extends AbstractWriter {
     writeBlob(value: Uint8Array): void {
         _assertDefined(value);
         this.checkWriteValue();
-        if (value === null || value === undefined) {
+        if (value === null) {
             this.writeNull(IonTypes.BLOB);
             return;
         }
@@ -108,7 +108,7 @@ export class BinaryWriter extends AbstractWriter {
     writeBoolean(value: boolean): void {
         _assertDefined(value);
         this.checkWriteValue();
-        if (value === null || value === undefined) {
+        if (value === null) {
             this.writeNull(IonTypes.BOOL);
             return;
         }
@@ -119,7 +119,7 @@ export class BinaryWriter extends AbstractWriter {
     writeClob(value: Uint8Array): void {
         _assertDefined(value);
         this.checkWriteValue();
-        if (value === null || value === undefined) {
+        if (value === null) {
             this.writeNull(IonTypes.CLOB);
             return;
         }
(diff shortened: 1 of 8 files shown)


src/IonBinaryWriter.ts
```diff
@@ -96,6 +96,7 @@ export class BinaryWriter extends AbstractWriter {
     }
 
     writeBlob(value: Uint8Array): void {
+        _assertDefined(value);
         this.checkWriteValue();
         if (value === null || value === undefined) {
```
zslayton: There are quite a few places where the new call to `_assertDefined(value)` happens before an 

```typescript
if (value === null || value === undefined) {
//...
}
```

The above can now be simplified to `if (value === null)`.
```

File: test/IonWriter.ts [hunk 2]
```diff
@@ -0,0 +1,83 @@
...
+    writeNull(type: IonType): void { }
+    writeString(value: string | null): void { }
+    writeSymbol(value: string | null): void { }
+    writeTimestamp(value: Timestamp | null): void { }
+    writeValue(reader: Reader): void { }
+    writeValues(reader: Reader): void { }
+}
+
+describe('IonWriter interface', () => {
+    let testMethods = Reflect.ownKeys(TestWriter.prototype)
```
zslayton: Whoa, fancy!

test/IonWriter.ts
```diff
@@ -0,0 +1,83 @@
...
+ */
+
+import {assert} from "chai";
+import * as ion from "../src/IonTests";
+import {Decimal, IonType, IonTypes, Reader, Timestamp} from "../src/IonTests";
+import JSBI from "jsbi";
+
+// This dummy impl exists as a substitute for reflection directly against
+// the Writer interface (which TypeScript doesn't currently support)
+class TestWriter implements ion.Writer {
```
zslayton: This file/class might be worth naming something more specific. `UndefinedValueWriter`?

src/IonBinaryWriter.ts
pbcornell: indeed, simplified

test/IonWriter.ts
pbcornell: Used generic terms with the idea that other tests against the `Writer` interface could go here, but I've renamed the file to `IonWriterUndefinedParameters`, and renamed `TestWriter` to `NoopWriter`.  Hope that helps a bit

Issue #130, opened 2018-04-09
Need to remove all undefined behavior logic to throw early


wesboyt:
Replace with null values or refactor out.

pbcornell:
Impl shouldn't check (or rely upon) `undefined`.  Consider use of `undefined` to be a bug;  use `null` (or a value) instead.

Issue #90, opened 2017-11-02
Removal of strict typing compiling errors.


wesboyt:
Currently the js library cannot statically compile due to numerous typing errors around unidentified reference error handling.

PR #522, opened 2019-12-03
Enables typescript compiler's strictNullChecks option and updates code as needed

### LINKED ISSUE #234: enable the typescript compiler's strictNullChecks option
See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html for details.

pbcornell:
Resolves #234 

* in `tsconfig.json`, sets `strictNullChecks` to `true`
* aligns method signatures with those of the implemented interface
* adds a number of not-null (!) assertions
* removes tests that can no longer compile (e.g., attempts to call `writeNull()` or `writeNull(undefined)`)
* removes logic that supported `writeDecimal('123.456')` in the binary writer
* does not attempt to remove `undefined` (to be done as part of #90 / #130)
* misc. cleanup when I noticed IDE suggestions, including:
  * addition of `readonly` to properties
  * removal of unused artifacts

### DIFF
--- a/src/IonDecimal.ts
+++ b/src/IonDecimal.ts
@@ -82,9 +82,9 @@ export class Decimal {
     constructor(coefficient: JSBI, exponent: number, isNegative?: boolean);
 
     // This is the unified implementation of the above signatures and is not visible to users.
-    constructor(coefficient: number | JSBI | string, exponent?: number, isNegative?: boolean) {
+    constructor(coefficient: number | JSBI | string, exponent?: number, isNegative: boolean = false) {
         if (typeof coefficient === "string") {
-            return Decimal.parse(coefficient);
+            return Decimal.parse(coefficient)!;
         }
 
         if (!_hasValue(exponent)) {
@@ -92,7 +92,7 @@ export class Decimal {
         }
 
         if (typeof coefficient === "number") {
-            return Decimal._fromNumberCoefficient(coefficient, exponent);
+            return Decimal._fromNumberCoefficient(coefficient, exponent!);
         }
 
         if (coefficient instanceof JSBI) {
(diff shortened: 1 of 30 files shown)


src/IonDecimal.ts
```diff
@@ -138,22 +138,22 @@ export class Decimal {
...
-            exponentDelimiterIndex = d.index;
+        if (d !== undefined && d !== null) {
+            exponent = Number(str.substring(d.index! + 1, str.length));
+            exponentDelimiterIndex = d.index!;
         }
         let f = str.match('\\.');
         let coefficientText: string;
         if (f) {
-            let exponentShift = d ? (d.index - 1) - f.index : (str.length - 1) - f.index;
+            let exponentShift = d ? (d.index! - 1) - f.index! : (str.length - 1) - f.index!;
```
zslayton: I wonder if `tsc` is smart enough to minimize the number of `!`s that actually expand to non-null checks or if we're paying for that branch+throw every time it appears.

src/IonParserBinaryRaw.ts
```diff
@@ -58,15 +58,8 @@ import {Timestamp, TimestampPrecision} from "./IonTimestamp";
...
 
-const DEBUG_FLAG = true;
-
 const EOF              = -1;  // EOF is end of container; distinct from undefined which is value has been consumed
-const ERROR            = -2;
-const TB_UNUSED__      = 15;
 const TB_DATAGRAM      = 20;  // fake type of the top level
-const TB_SEXP_CLOSE    = 21;
-const TB_LIST_CLOSE    = 22;
-const TB_STRUCT_CLOSE  = 23;
```
zslayton: Thanks for cleaning these up.

src/IonDecimal.ts
pbcornell: The not-null assertions appear to be compile-time only.  For example, the generated JavaScript of this line is simply:
```javascript
let exponentShift = d ? (d.index - 1) - f.index : (str.length - 1) - f.index;
```

And there's no generated logic to assert non-null-ness at runtime in this block.

PR #526, opened 2019-12-06
Adds unit tests for ParserTextRaw

### LINKED ISSUE #330: We are not unit testing IonParserTextRaw
We are only testing it with integration tests, not ideal.

pbcornell:
Code coverage for ParserTextRaw prior to this PR is already quite reasonable:
* statements:  90.48%
* branches:  84.75%
* functions:  100%
* lines:  91.25%

The changes in this PR do not attempt to provide complete coverage, but rather add unit tests in hopes of catching changes in behavior if/as the implementation evolves over time.

### DIFF
--- a/test/IonParserTextRaw.ts
+++ b/test/IonParserTextRaw.ts
@@ -1,0 +1,287 @@
+/*!
+ * Copyright 2012 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *  
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *  
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *  
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import {assert} from 'chai';
+import {StringSpan} from '../src/IonSpan';
+import {ParserTextRaw} from '../src/IonParserTextRaw';
+import {IonTypes} from "../src/IonTypes";
+import {IonType} from "../src/IonType";
+import JSBI from "jsbi";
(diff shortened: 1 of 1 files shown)


pbcornell:
> Looks like a good start, I think theres alot of value in testing the machinery within the parser itself. checking the this._start, this._end, this._esc_len as well as the machinery around ops.

The machinery you mention is all private, and thereby not visible (or testable) outside of ParserTextRaw--am I missing something here?

> I think the biggest source of bugs in the future will be our devs refactoring the lexing/parsing explicitly as we actually lex everything twice instead of tokenizing our inputs.

If we rewrite/refactor the internals, we want to know if the contract of this class changes--that's what these new tests are designed to catch, albeit incompletely (as Ion has so many edge cases).

wesboyt:
> > Looks like a good start, I think theres alot of value in testing the machinery within the parser itself. checking the this._start, this._end, this._esc_len as well as the machinery around ops.
> 
> The machinery you mention is all private, and thereby not visible (or testable) outside of ParserTextRaw--am I missing something here?
> 
> > I think the biggest source of bugs in the future will be our devs refactoring the lexing/parsing explicitly as we actually lex everything twice instead of tokenizing our inputs.
> 
> If we rewrite/refactor the internals, we want to know if the contract of this class changes--that's what these new tests are designed to catch, albeit incompletely (as Ion has so many edge cases).

Well you're saying that this pr solves the issue listed which is specific to unit tests. if we arent going to unit test it we shouldn't close the issue.

pbcornell:
> Well you're saying that this pr solves the issue listed which is specific to unit tests. if we arent going to unit test it we shouldn't close the issue.

Perhaps I've missed the intent of #330.  This PR adds unit tests of `ParserTextRaw`'s API, but not its internals.  If the intent of #330 includes unit testing the internals, I wonder what doing so would look like due to the nature of the internal methods themselves (many of which appear to be highly interdependent and thereby difficult to isolate), and the `private` accessibility now enforced by TypeScript.

I've updated the description of this PR so #330 shouldn't auto-close with it.  Perhaps you could clarify the expectations of that issue and suggest how it might be accomplished.

Issue #458, opened 2019-10-07
Annotations unit tests need to be migrated to mocha. Specifically the binary/text writer tests.


wesboyt:
https://github.com/amzn/ion-js/pull/406

PR #529, opened 2019-12-11
reapplies missing annotations tests.

### LINKED ISSUE #253: API allows annotations[] to be undefined
APIs such as `Writer.writeList()` accepts `annotations` as either string[] or `undefined`.  Similarly, `Reader.annotations()` returns string[] or `undefined`.  I'd suggest excluding `undefined` from such APIs.

wesboyt:
*Issue #, if available:*
Fixes #253 #458

### DIFF
--- a/test/IonBinaryWriter.ts
+++ b/test/IonBinaryWriter.ts
@@ -1052,6 +1052,99 @@ let badWriterTests: BadTest[] = [
         }
     },
     {
+        name:'Should throw when passing a single string as an annotation.',
+        instructions: (writer) => {
+            // @ts-ignore
+            writer.setAnnotations('taco');
+            writer.writeInt(5);
+        }
+    },
+    {
+        name:'Should throw when setting annotations to null.',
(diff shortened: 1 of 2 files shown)


wesboyt:
> I'm surprised to see `@ts-ignore` for the binary writer, and not the text writer. What's the reason for this discrepancy?
> 
> Might be nice to avoid some of the duplication by coding these as parameterized tests (see [IonWriterUndefinedParameters.ts#L59](https://github.com/amzn/ion-js/blob/master/test/IonWriterUndefinedParameters.ts#L59) as an example). Writer type (binary|text) and the value passed to the annotation method being tested could both be parameterized.
> 
> But I'm fine with these tests as-written.

I believe the binary tests structure allows for static type checking, and the text side does not have any types defined within the describe itself.
Text:
`let badWriterTest = function (name, instructions) {
    let test = function () {
        instructions(writer);
        writer.close();
    };
    it(name, () => {
        assert.throws(test, Error);
    });
};`

PR #530, opened 2019-12-13
fixes incorrect escape sequences within symbols, strings, and tripleq…

### LINKED ISSUE #515: CR-LF newline pairs inject \r and \n into strings incorrectly
roundtrip we turn [string_cr_nl.ion](https://github.com/amzn/ion-tests/blob/master/iontestdata/good/strings_cr_nl.ion) into:
"short1\n\nmulti-line string\r\nwith embedded\nnew line\r\ncharacters\n\"
it should become:
"short1multi-line string\nwith embedded\nnew line\ncharacters"

roundtrip we turn [symbolEmptyWithCRLF.ion](https://github.com/amzn/ion-tests/blob/master/iontestdata/good/symbolEmptyWithCRLF.ion) into:
'\n'
it should become:
''

wesboyt:
…uoted strings.

*Issue #, if available:*
Fixes #515
*Description of changes:*
-correctly adjusts indexes by 2 instead of 3 within the text parser when escaped \r\n pairs are found
-removes \r from unescaped \r\n pairs within triple quotes.

### DIFF
--- a/src/IonParserTextRaw.ts
+++ b/src/IonParserTextRaw.ts
@@ -396,6 +396,10 @@ export class ParserTextRaw {
                         if (isEscaped) {
                             s += String.fromCodePoint(ch);
                         } else {
+                            if(t === T_STRING3 && ch === ESC_nl3 && this._in.valueAt(index + 1) === ESC_nl2) {
+                                ch = ESC_nl2;
+                                index++;
+                            }
                             s += String.fromCharCode(ch);
                         }
                     }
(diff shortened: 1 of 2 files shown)


wesboyt:
> > removes \r from unescaped \r\n pairs within triple quotes.
> 
> Do I understand correctly that this will subtly modify the string that's read from input?
> 
> e.g. The file contains:
> 
> ```
> """hello \r\n world"""
> ```
> 
> But `reader.stringValue()` will return `"hello \n world"`?

yes, that is the behavior I found in ionjava, ionc, and talked to greggt about.

pbcornell:
Do these changes allow us to remove anything from the skiplists?  If not, let's add unit tests to verify this fix (and prevent regression in the future).

zslayton:
> yes, that is the behavior I found in ionjava, ionc, and talked to greggt about.
> Effectively if we see the pair 13 and 10 we remove the carriage return.

It took me a while to find it, but the relevant bit of the spec is [here](http://amzn.github.io/ion-docs/docs/stringclob.html#text-format):

> Within long string literals unescaped newlines are normalized such that U+000D U+000A pairs (CARRIAGE RETURN and LINE FEED respectively) and U+000D are replaced with U+000A. This is to facilitate compatibility across operating systems.

Issue #454, opened 2019-10-03
IonAnnotationsTest relies on _writeValues


zslayton:
See [here](https://github.com/amzn/ion-js/blob/6fa07b4f5a820f28ee02e49fac4676a132559186/tests/unit/IonAnnotationsTest.js#L79). `_writeValues()` is not part of `Writer`'s public API.

jonwilsdon:
Avoid adding the depth parameter to the public API.

zslayton:
I reviewed `ion-java`'s implementation of `writeValues` to clarify the expected behavior of that method in `ion-js`. It performs [limited checks](https://github.com/amzn/ion-java/blob/master/src/com/amazon/ion/impl/bin/AbstractIonWriter.java#L90-L96) regarding the starting state of the reader and writer.

If a field name is set on the reader (which should only happen when it's inside a struct), the writer will attempt to set one if it is also inside a struct. If the writer is not positioned within a struct, the reader's field name will be ignored.

If a field name is not set on the reader but the writer IS in a struct, it will [throw an `IllegalStateException`](https://github.com/amzn/ion-java/blob/master/src/com/amazon/ion/impl/bin/IonRawBinaryWriter.java#L887-L890). 

Ideally I think there would be an early sanity check for the reader-not-in-struct+writer-in-struct use case, but that would require new APIs to be added to test `Reader#isInStruct()` from the `Writer` class. 

This could also be addressed by an addendum to [`IonWriter#writeValues()`' documentation](https://github.com/amzn/ion-java/blob/master/src/com/amazon/ion/IonWriter.java#L327-L340), which currently doesn't mention it.

zslayton:
Fixed by PR #532

PR #532, opened 2019-12-18
writeValues() works when writers/readers at depth>0

zslayton:
*Issue #, if available:* #454 

*Description of changes:*

This PR makes it legal to call `IonWriter#writeValues()` when the `Writer` and `Reader` are at depths greater than zero. It does not change the public-facing API.

* Added a protected abstract method `_isInStruct: boolean` to `AbstractWriter`. Subclasses (`BinaryWriter`, `TextWriter`) each provide their own implementations.
* Added docs to the `IonWriter` interface explaining the behavior of `IonWriter#writeValues()`.
* Removed `IonAnnotationsTest`'s sneaky call to `writer['_writeValues']`.
* Added unit tests.

### DIFF
--- a/src/AbstractWriter.ts
+++ b/src/AbstractWriter.ts
@@ -42,6 +42,10 @@ export abstract class AbstractWriter implements Writer {
         }
     }
 
+    // This method is not part of the Writer interface, but subclasses of AbstractWriter must implement it
+    // in order to use the default implementations of writeValue() and writeValues()
+    protected abstract _isInStruct(): boolean
+
     writeValues(reader: Reader): void {
         this._writeValues(reader);
     }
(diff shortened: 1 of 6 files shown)


src/AbstractWriter.ts
```diff
@@ -54,53 +58,58 @@ export abstract class AbstractWriter implements Writer {
...
+                throw new Error("Cannot call writeValue() when the Writer is in a Struct but the Reader is not.");
             }
+            this.writeFieldName(fieldName);
         }
+
         this.setAnnotations(reader.annotations());
+
         if (reader.isNull()) {
             this.writeNull(type);
-        } else {
```
zslayton: This is just a formatting fix--this block was indented too far.

Issue #414, opened 2019-09-12
Equivs/nonEquivs testing should include a roundtrip phase


zslayton:
In issue #408, the text and binary `Reader`s could distinguish between `0` and `-0` but the `Writer`s couldn't. This meant that simply reading the data in `ion-tests` was not enough to notice a problem, and round-tripping it produced the same (incorrect) values in both the text and binary files produced.

We should consider expanding our equivs/non-equivs tests to also roundtrip the data and make sure the tests still pass after the data has been serialized/deserialized.

tgregg:
There is some discussion of this in the [ion-test-driver](https://github.com/amzn/ion-test-driver) README (for example, see [this section](https://github.com/amzn/ion-test-driver#goodnon-equivs-file-with-embedded-streams)). We should all review that document and determine whether anything is missing or unclear, or whether there is anything else that is present in the document but missing from the ion-js implementation.

PR #539, opened 2020-01-15
Embedded document support

wesboyt:
*Issue #, if available:*
Fixes https://github.com/amzn/ion-js/issues/270
Fixes https://github.com/amzn/ion-js/issues/414

*Description of changes:*
Adds embedded document support and refactors some related redundant codepaths.

### DIFF
--- a/test/iontests.ts
+++ b/test/iontests.ts
@@ -110,11 +110,14 @@ function equivsTest(path: string, expectedEquivalence = true) {
         let comparisons : any = [];
         if (event.annotations[0] === 'embedded_documents') {
             for (let j = 0; j < event.ionValue.length - 1; j++) {
-                comparisons.push([new IonEventStream(ion.makeReader(event.ionValue)), new IonEventStream(ion.makeReader(textEvent.ionValue)), new IonEventStream(ion.makeReader(binEvent.ionValue))]);
+                comparisons.push([new IonEventStream(ion.makeReader(event.ionValue[j].ionValue)), new IonEventStream(ion.makeReader(textEvent.ionValue[j].ionValue)), new IonEventStream(ion.makeReader(binEvent.ionValue[j].ionValue))]);
             }
         } else {//were in an sexp
-            for (let j = 0; j < event.ionValue.length - 1; event.ionValue[j].eventType === IonEventType.SCALAR ? j++ : j += event.ionValue[j].ionValue.length) {
+            for (let j = 0; j < event.ionValue.length - 1; j++) {
                 comparisons.push([event.ionValue[j], textEvent.ionValue[j], binEvent.ionValue[j]]);
+                if (event.ionValue[j].eventType === IonEventType.CONTAINER_START) {
+                    j += event.ionValue[j].ionValue.length
+                }
             }
         }
 
@@ -122,7 +125,7 @@ function equivsTest(path: string, expectedEquivalence = true) {
         for (let j = 0; j < comparisons.length - 1; j++) {
             for (let k = j + 1; k < comparisons.length; k++) {
                 for (let l = 0; l < width; l++) {
-                    for (let m = l + 1; m < width; m++) {
+                    for (let m = 0; m < width; m++) {
                         assert.isTrue(comparisons[j][l].equals(comparisons[k][m]) === expectedEquivalence);
                     }
                 }
--- a/test/iontests.ts
+++ b/test/iontests.ts
@@ -18,6 +18,7 @@ import * as ion from '../src/Ion';
 import * as fs from 'fs';
 import * as path from 'path';
 import {IonEventStream} from '../src/IonEventStream';
+import {IonEvent} from '../src/IonEvent';
 import {IonType, Reader, Timestamp, Writer} from '../src/Ion';
 
 function findFiles(folder: string, files: string[] = []) {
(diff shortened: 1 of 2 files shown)


test/iontests.ts
```diff
@@ -45,65 +46,50 @@ function exhaust(reader: Reader) {
...
+    binWriter.writeValues(ion.makeReader(bytes));
+    binWriter.close();
+    let binBytes = binWriter.getBytes();
+    let binEvents = (new IonEventStream(ion.makeReader(binBytes))).getEvents();
+    for (let i = 0; i < originalEvents.length - 2; i += originalEvents[i].ionValue.length + 1) {
+        let event = originalEvents[i];
+        let textEvent = textEvents[i];
+        let binEvent = binEvents[i];
+        //if were equal everybody equals everybody
+        let comparisons : any = [];
```
pbcornell: Comment please.  looks like each element in this array will be either:
- [original IonEventStream, text IonEventStream, binary IonEventStream]
- or [original ionValue, text ionValue, binary ionValue]

Or perhaps I'm confused as to what `event.ionValue[j]` actually contains...?

test/iontests.ts
```diff
@@ -45,65 +46,50 @@ function exhaust(reader: Reader) {
...
-    let textValues = loadValues(path, () => ion.makeTextWriter());
-
-    for (let c = 0; c < binaryValues.length; c++) {
-        for (let i = 0; i < binaryValues[c].length; i++) {
-            for (let j = i + 1; j < binaryValues[c].length; j++) {
-                compare(c, i, j, binaryValues[c][i], binaryValues[c][j], expectedEquivalence);
-                compare(c, i, j, binaryValues[c][i], textValues[c][j], expectedEquivalence);
-                compare(c, i, j, textValues[c][i], textValues[c][j], expectedEquivalence);
+        let width = comparisons[0].length;
+        for (let j = 0; j < comparisons.length - 1; j++) {
```
pbcornell: I've never seen 4 nested for loops (and we're still within the `i` loop, so we're actually nested 5 levels deep here).  This deserves some clarifying comments.

And what is `width`, really?

pbcornell: and let's have `j` range from `0` to `comparisons.length` -- shouldn't cause any change in behavior, but it's cleaner/more-readable to let the range of `k` govern this (instead of also limiting the range of `j`)

test/iontests.ts
```diff
@@ -45,65 +46,50 @@ function exhaust(reader: Reader) {
...
+            for (let k = j + 1; k < comparisons.length; k++) {
+                for (let l = 0; l < width; l++) {
+                    for (let m = 0; m < width; m++) {
+                        if (equivsTimelines) {
+                            assert.isTrue(comparisons[j][l].ionValue.compareTo(comparisons[k][m].ionValue) === 0);
+                        } else {
+                            assert.isTrue(comparisons[j][l].equals(comparisons[k][m]) === expectedEquivalence);
```
pbcornell: or `assert.equals(comparisons[j][l].equals(comparisons[k][m]), expectedEquivalence);`

test/iontests.ts
```diff
@@ -45,65 +46,50 @@ function exhaust(reader: Reader) {
...
-                compare(c, i, j, binaryValues[c][i], binaryValues[c][j], expectedEquivalence);
-                compare(c, i, j, binaryValues[c][i], textValues[c][j], expectedEquivalence);
-                compare(c, i, j, textValues[c][i], textValues[c][j], expectedEquivalence);
+        let width = comparisons[0].length;
+        for (let j = 0; j < comparisons.length - 1; j++) {
+            for (let k = j + 1; k < comparisons.length; k++) {
+                for (let l = 0; l < width; l++) {
+                    for (let m = 0; m < width; m++) {
+                        if (equivsTimelines) {
+                            assert.isTrue(comparisons[j][l].ionValue.compareTo(comparisons[k][m].ionValue) === 0);
```
pbcornell: or `assert.equals(comparisons[j][l].ionValue.compareTo(comparisons[k][m].ionValue), 0);`

test/iontests.ts
wesboyt: Each top level value is a container (either sexp or list of strings). The ionValue is the backing array of inner IonEvents.

test/iontests.ts
wesboyt: Width is just the number of streams were comparing, the "width" of the array of comparisons.

test/iontests.ts
```diff
@@ -45,65 +46,50 @@ function exhaust(reader: Reader) {
...
+    let textWriter = ion.makeTextWriter();
+    textWriter.writeValues(ion.makeReader(bytes));
+    textWriter.close();
+    let textBytes = textWriter.getBytes();
+    let textEvents = (new IonEventStream(ion.makeReader(textBytes))).getEvents();
+    let binWriter = ion.makeBinaryWriter();
+    binWriter.writeValues(ion.makeReader(bytes));
+    binWriter.close();
+    let binBytes = binWriter.getBytes();
+    let binEvents = (new IonEventStream(ion.makeReader(binBytes))).getEvents();
```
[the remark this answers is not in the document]
wesboyt: I disagree

test/iontests.ts
pbcornell: We're missing a comment here that describes what the `comparisons` variable contains, as it is not obvious from reading the code, and it appears to differ between 'embedded_documents' and non-embedded_documents.

For an `IonEvent` representing a scalar value, `IonEvent.ionValue` appears to be the value returned by `Reader.value()`.  For an `IonEvent` representing a container, `IonEvent.ionValue` is an array of `IonEvents` representing the container start, contents, and end.

That said, each element of the `comparisons` array appears to be either:
* [IonEventStream (original), IonEventStream (text), IonEventStream (binary)]
* or [IonEvent.ionValue (original), IonEvent.ionValue (text), IonEvent.ionValue (binary)] -- where `IonEvent.ionValue` may be either a scalar value or an array of `IonEvents` (for containers)

Is that accurate?

wesboyt: yes, both have the same comparison method equals().

PR #560, opened 2020-02-19
Merged Value interface and DomValue mixin.

zslayton:
*Description of changes:*

This PR uses Typescript's [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) feature to combine the `Value` interface and the `DomValue` mixin into a single combined entity.

Classes can now extend `Value(...)` instead of `DomValue(...)`.

This allows you to test whether a given object is an instance of an Ion dom `Value` by checking: 
```typescript
if (value instanceof Value) {
    //...
}
```

(Previously, this would have required `instanceof DomValue`, which is slightly incongruous.)

This change also makes it possible to store functions on the `Value` namespace, simulating static methods. This will enable a nice way to convert native JS values to Ion dom `Value`s:

```typescript
let v: Value = Value.from(7);
```

Note: this PR is currently targeting the `zslayton-load-dom` branch, which contains the code from PR #557. Once that PR is merged, I'll change the target to `master` (rebasing if needed).


src/dom/Value.ts
```diff
@@ -126,4 +135,141 @@ export interface Value {
  *
  * See: Value#get
  */
-export type PathElement = string | number;
\ No newline at end of file
+export type PathElement = string | number;
```
zslayton: The contents of the `DomValue.ts` file have been moved to `Value.ts`. The only changes are:
* The `DomValue` mixin was renamed `Value` so it would merge with the `Value` interface above.
* A `Value` namespace was set up at the bottom of the file to accommodate upcoming functionality.

zslayton:
Closing in favor of #561.

Issue #579, opened 2020-03-13
Create dump() convenience method to complement Value.from() and ion.load()


zslayton:
Now that #562 and #568 have been merged, it's straightforward to write any given JS value to an Ion stream via:

```typescript
let jsValue = ...;
let writer = ion.makeBinaryWriter(); // or makeTextWriter(), makePrettyTextWriter()
dom.Value.from(jsValue).writeTo(writer);
writer.close();
let ionBytes: Uint8Array = writer.getBytes();
```

Since this will be a common pattern, we should provide convenience methods to streamline this even further. I suggest offering a version of `dump()` to accompany each existing flavor of `makeWriter()`: 

```typescript
function dumpBinary(...values: any[]): Uint8Array {}
function dumpText(...values: any[]): string {}
function dumpPrettyText(...values: any[]): string {}
```

zslayton:
Addressed by #580.

PR #580, opened 2020-03-13
Added dumpBinary(), dumpText(), dumpPrettyText().

zslayton:
*Issue #, if available:* #579 

CC: @yohanmartin

*Description of changes:*

### Added `ion.dump*()` functions

Adds 3 new top-level functions in the `ion` module:

```typescript
function dumpBinary(...values: any[]): Uint8Array {}
function dumpText(...values: any[]): string {}
function dumpPrettyText(...values: any[]): string {}
```

Each of these constructs the `Writer` implementation that corresponds to the requested format, writes each value to the writer, closes the writer, and returns the complete serialized stream to the user.

### Moves some test functions to a common `src` file

It makes sense to re-use some test functions and sample values across multiple `.ts` files in `test`. However, the Typescript compiler will throw an error if you reference functions from another file in `test`. I've instead moved the relevant functions to `src/IonTests.ts`, which already existed for this reason.

### Avoids using re-exports of load() and loadAll() within `test/dom`

`ion.load()` and `ion.loadAll()` are re-exports of `ion.dom.load()` and `ion.dom.loadAll()` respectively. The compiler will sometimes complain that `ion.load()` is not yet defined if it is referenced from within `ion.dom` test code. Changing these references to `ion.dom.*` in the test code prevents this issue.

### DIFF
--- a/src/Ion.ts
+++ b/src/Ion.ts
@@ -88,7 +88,7 @@ export function makeBinaryWriter(): Writer {
 }
 
 // Used by the dump*() functions to write each of a sequence of values to the provided Writer.
-function writeAllTo(writer: Writer, values: any[]): Uint8Array {
+function _writeAllTo(writer: Writer, values: any[]): Uint8Array {
     for (let value of values) {
         dom.Value.from(value).writeTo(writer);
     }
@@ -101,7 +101,7 @@ function writeAllTo(writer: Writer, values: any[]): Uint8Array {
  * @param values Values to encode in Ion.
  */
 export function dumpBinary(...values: any[]): Uint8Array {
-    return writeAllTo(makeBinaryWriter(), values);
+    return _writeAllTo(makeBinaryWriter(), values);
 }
 
 /**
@@ -109,7 +109,7 @@ export function dumpBinary(...values: any[]): Uint8Array {
  * @param values Values to encode in Ion.
  */
 export function dumpText(...values: any[]): string {
-    return decodeUtf8(writeAllTo(makeTextWriter(), values));
+    return decodeUtf8(_writeAllTo(makeTextWriter(), values));
 }
 
 /**
@@ -118,7 +118,7 @@ export function dumpText(...values: any[]): string {
  * @param values Values to encode in Ion.
  */
 export function dumpPrettyText(...values: any[]): string {
-    return decodeUtf8(writeAllTo(makePrettyWriter(), values));
+    return decodeUtf8(_writeAllTo(makePrettyWriter(), values));
 }
 
 export {Reader, ReaderScalarValue} from "./IonReader";
--- a/src/Ion.ts
+++ b/src/Ion.ts
@@ -87,6 +87,40 @@ export function makeBinaryWriter(): Writer {
     return new BinaryWriter(localSymbolTable, new Writeable());
 }
 
+// Used by the dump*() functions to write each of a sequence of values to the provided Writer.
+function writeAllTo(writer: Writer, values: any[]): Uint8Array {
+    for (let value of values) {
+        dom.Value.from(value).writeTo(writer);
+    }
+    writer.close();
+    return writer.getBytes();
+}
+
+/**
+ * Returns a binary Ion representation of the provided values.
+ * @param values Values to encode in Ion.
+ */
+export function dumpBinary(...values: any[]): Uint8Array {
+    return writeAllTo(makeBinaryWriter(), values);
+}
+
+/**
+ * Returns a compact text Ion representation of the provided values.
+ * @param values Values to encode in Ion.
+ */
+export function dumpText(...values: any[]): string {
+    return decodeUtf8(writeAllTo(makeTextWriter(), values));
+}
+
+/**
+ * Returns a text Ion representation of the provided values that is generously spaced for
+ * easier human readability.
+ * @param values Values to encode in Ion.
+ */
+export function dumpPrettyText(...values: any[]): string {
+    return decodeUtf8(writeAllTo(makePrettyWriter(), values));
+}
+
 export {Reader, ReaderScalarValue} from "./IonReader";
 export {Writer} from "./IonWriter";
 export {Catalog} from "./IonCatalog";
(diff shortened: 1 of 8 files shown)


src/IonTests.ts
```diff
@@ -25,3 +29,55 @@ export {BinaryWriter} from "./IonBinaryWriter";
 export {ParserBinaryRaw} from "./IonParserBinaryRaw";
 export {BinarySpan} from "./IonSpan";
 export {Decimal} from "./IonDecimal";
+
+// Generates a mocha-friendly string representation of the provided value that can be used in test names
+export function valueName(value: any): string {
```
zslayton: This code was moved from `test/dom/Value.ts` so it could be referenced from multiple test files.

test/dom/dom.ts
```diff
@@ -2,7 +2,7 @@ import {assert} from "chai";
 import JSBI from "jsbi";
 import * as ion from "../../src/Ion";
 import {IonTypes} from "../../src/Ion";
-import {Value} from "../../src/dom";
+import {Value, load, loadAll} from "../../src/dom";
```
zslayton: References to `ion.load()` (a re-export of `ion.dom.load()`) in this file have been changed to direct references to `ion.dom.load()` to avoid a compile time issues that arise inconsistently.

test/dump.ts
```diff
@@ -0,0 +1,84 @@
...
+
+// Dumps the provided values to Ion using the specified dump*() function, uses loadAll() to read the values
+// back in, then compares the input values to the loaded values.
+function roundTripTest(dumpFn: (...values: any[]) => any,
+                       inputValues: any[],
+                       expectedValues: dom.Value[]) {
+    let actualValues: dom.Value[] = dom.loadAll(dumpFn(inputValues));
+    let testComparisons: [dom.Value, dom.Value][] = actualValues.map(
+        (actualValue, index) => [actualValue, expectedValues[index]]
+    );
```
pbcornell: does `testComparisons` add value over having the `for` loop below iterate over `actualValues`?

test/dump.ts
```diff
@@ -0,0 +1,84 @@
...
+        writer.writeInt(3);
+        writer.stepOut();
+        writer.close();
+        let expectedBytes: Uint8Array = writer.getBytes();
+        let byteComparisons: [number, number][] = Array.from(ionBinary).map(
+            (byte, index) => [byte, expectedBytes[index]]
+        );
+        for (let [actualByte, expectedByte] of byteComparisons) {
+            assert.equal(actualByte, expectedByte);
+        }
```
pbcornell: consider simplifying to `assert.deepEqual(ionBinary, expectedBytes)`, which i think accomplishes the same thing

test/dump.ts
```diff
@@ -0,0 +1,84 @@
...
+    let testComparisons: [dom.Value, dom.Value][] = actualValues.map(
+        (actualValue, index) => [actualValue, expectedValues[index]]
+    );
+    for (let [actualValue, expectedValue] of testComparisons) {
+        //TODO: This tests for a very weak common definition of equality.
+        //      We should make this a stronger test when .equals() is added to the Value interface[1].
+        //      In the meantime, the core writing/equality logic is more strictly tested by
+        //      test/dom/Value.ts, which evaluates the data produced by the Value.writeTo() function
+        //      at the core of the ion.dump*() methods.
+        //      [1] https://github.com/amzn/ion-js/issues/576
```
pbcornell: understood, thanks for calling this out

src/IonTests.ts
zslayton: Some investigation determined that the issue stemmed from `tsconfig.json` setting the project's `rootDir` to `src` rather than the actual root, which caused `test` to be considered a non-source directory. Setting `rootDir` to `.` (the same directory as `tsconfig.json` itself) lifted this restriction; it does not appear to have affected anything else.

I've created `test/exampleValues.ts` to house the functionality I had previously placed in `src/IonTests.ts`.

> I'd like to see `src/IonTests.ts` removed at some point

Now that we have a path forward, I've opened #582 to track this work.

src/Ion.ts
```diff
@@ -87,6 +87,40 @@ export function makeBinaryWriter(): Writer {
     return new BinaryWriter(localSymbolTable, new Writeable());
 }
 
+// Used by the dump*() functions to write each of a sequence of values to the provided Writer.
+function writeAllTo(writer: Writer, values: any[]): Uint8Array {
+    for (let value of values) {
+        dom.Value.from(value).writeTo(writer);
+    }
+    writer.close();
```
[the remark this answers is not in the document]
zslayton: Yes, I certainly wouldn't have done it for a public-facing function. I've renamed `writeAllTo` to `_writeAllTo` (as I should've done initially) to emphasize that this internal only.

test/dump.ts
zslayton: Based on serialization peculiarities I've experienced using `Uint8Array`, I had assumed that `deepEqual` would evaluate properties of the object itself rather than comparing the stored bytes. I'm pleased to discover that I was wrong. I've converted this to `deepEqual`, thanks.

test/dump.ts
zslayton: I need access to the index of `actualValue`, which I can't get from
```typescript
for (let actualValue of actualValues) {
  let expectedValue = expectedValues[???];
  // Compare them
}
```

Zipping the two arrays together in advance makes for a nice, self-explanatory loop header:
```typescript
for (let [actualValue, expectedValue] of testComparisons) {}
```

Unfortunately, Javascript doesn't provide a `zip` method out of the box so you end up seeing how the sausage gets made.

I've converted this to a `forEach` in the hopes that that's more palatable:
```typescript
expectedValues.forEach((expectedValue, index) => {
   let actualValue = actualValues[index];
   // Compare them
}
```

src/IonTests.ts
pbcornell: Thank you for taking a closer look.  I wonder if changing `rootDir` changes the paths produced within the `dist` folder...?

zslayton: It doesn't appear to. That behavior is set by `outDir`, which is not specified relative to `rootDir`.

PR #589, opened 2020-03-20
Changes jsbi to be a peerDependency

### LINKED ISSUE #588: Change jsbi to be a peerDependency
Per [this article](https://lexi-lambda.github.io/blog/2016/08/24/understanding-the-npm-dependency-model/), since jsbi is exposed as part of the API of this package.

And while [semver](https://semver.org/) indicates that an incompatible API change must result in incrementing the major version, the consensus in https://github.com/semver/semver/issues/502 seems to be that adding a peerDependency is a breaking change (unless the consuming package already depends on a compatible version of the peerDependency).

pbcornell:
Resolves #588 

For development purposes, I'm proposing we pin the devDependency of jsbi to 3.1.1.  When ^3.1.1 is specified, npm currently installs 3.1.2, which is fine; however, when linking a local ion-js to a local ion-hash-js, npm install for ion-hash-js replaces ion-js's jsbi with version 3.1.1 for some odd reason, which tsc identifies as incompatible with 3.1.2 (from ion-hash-js).  It's quite confusing.

As noted in #588, this is a breaking change, as it has the potential to break the build of any consuming package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "grunt-ts": "^6.0.0-beta.21",
     "grunt-tslint": "^5.0.2",
     "grunt-typedoc": "^0.2.4",
+    "jsbi": "3.1.1",
     "mocha": "^6.2.0",
     "mocha-typescript": "^1.1.17",
     "nyc": "^14.1.1",
@@ -79,7 +80,7 @@
     "typedoc": "^0.16.10",
     "typescript": "^3.6.2"
   },
-  "dependencies": {
+  "peerDependencies": {
     "jsbi": "^3.1.1"
   }
 }


pbcornell:
> Why not pin to `3.1.2` if that's the latest and we're already making a breaking change?

As noted in the initial description of this PR:  "when linking a local ion-js to a local ion-hash-js, npm install for ion-hash-js replaces ion-js's jsbi with version 3.1.1 for some odd reason, which tsc identifies as incompatible with 3.1.2 (from ion-hash-js)".  I don't understand why this occurs, but pinning both devDependencies to 3.1.1 seems reasonable for the time being.  Note that consumers of either ion-js or ion-hash-js (or both) must provide their own dependency on jsbi, so they are not pinned to 3.1.1.

PR #593, opened 2020-03-27
Updated READMEs with DOM examples.

### LINKED ISSUE #586: Add Ion DOM examples to the README
Now that ion-js has a DOM API, we should update `README.md` to show examples of the `load()`, `Value.from(...)`, and `dump*()` functions.

zslayton:
*Issue #, if available:* Fixes #586.

*Description of changes:* 
* Adds a new `README.md` file under `/src/dom` demonstrating the new APIs.
* Updates the main `README.md` file's example of reading Ion to use `ion.load()` instead of `Reader`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,13 +20,18 @@ You can use this library either as a Node.js module or inside an HTML page.
     ```
     npm install --save ion-js
     ```
-1. Use the library to read/write ion data. Here is an example that reads Ion data from a Javascript string:
+1. Use the library to read/write Ion data. Here is an example that reads Ion data from a JavaScript string:
     ```javascript
     let ion = require("ion-js");
-   
+    
+    // Reading
     let ionData = '{ greeting: "Hello", name: "Ion" }';
     let value = ion.load(ionData);
     console.log(value.greeting + ", " + value.name + "!");
+   
+    // Writing
+    let ionText = ion.dumpText(value);
+    console.log("Serialized Ion: " + ionText);
     ```
    
    For more examples, see the [Ion DOM `README`](/src/dom/README.md).
--- a/README.md
+++ b/README.md
@@ -20,20 +20,16 @@ You can use this library either as a Node.js module or inside an HTML page.
     ```
     npm install --save ion-js
     ```
-1. Use the library to read/write ion data. Here is an example that reads Ion data from a Javascript string
+1. Use the library to read/write ion data. Here is an example that reads Ion data from a Javascript string:
     ```javascript
-    var ionJs = require("ion-js")
-
-    var ionData = "{ hello: \"Ion\" }";
-    var ionReader = ionJs.makeReader(ionData);
-    ionReader.next();
-    ionReader.stepIn();
-    ionReader.next();
-    var hello = ionReader.fieldName();
-    var ion = ionReader.stringValue();
-    ionReader.stepOut();
-    console.log(ion.concat(", ").concat(hello));
+    let ion = require("ion-js");
+   
+    let ionData = '{ greeting: "Hello", name: "Ion" }';
+    let value = ion.load(ionData);
+    console.log(value.greeting + ", " + value.name + "!");
     ```
+   
+   For more examples, see the [Ion DOM `README`](/src/dom/README.md).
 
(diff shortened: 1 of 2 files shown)


src/dom/README.md
```diff
@@ -0,0 +1,253 @@
...
+// Write the number 7 to a binary Ion stream and store the bytes in `data`
+let data = ion.dumpBinary(7);
+// Use ion.load() to parse `data`
+let i = ion.load(data);
+// Verify that the value found in the stream is the one we started with
+assert.equal(7, i);
+```
+
+### ion.dumpText() vs ion.dumpPrettyText() example
+```
```
pbcornell: enable javascript syntax highlighting

src/dom/README.md
```diff
@@ -0,0 +1,253 @@
...
+let ionText = ion.dumpText([1, 2, 3]);
+let ionPrettyText = ion.dumpPrettyText([1, 2, 3]);
+
+assert.equal(ionText, '[1,2,3]');
+assert.equal(ionPrettyText, '[\n  1,\n  2,\n  3\n]');
+```
+
+## `ion.dom` Data Types
+
+The Ion DOM data types each extend a native Javascript type:
```
pbcornell: JavaScript

src/dom/README.md
```diff
@@ -0,0 +1,253 @@
...
+`ion.dom.Clob` | `Uint8Array`
+`ion.dom.Blob` | `Uint8Array`
+`ion.dom.List` | `Array`
+`ion.dom.SExpression` | `Array`
+`ion.dom.Struct` | `Object`
+
+### Testing Equality
+
+Note that the JS types being extended are "[wrapper objects](https://developer.mozilla.org/en-US/docs/Glossary/Primitive#Primitive_wrapper_objects_in_JavaScript)"
+that box the primitive types themselves. This matters when testing for equality; Javascript's strict definition of equality
```
pbcornell: JavaScript's

README.md
```diff
@@ -20,20 +20,16 @@ You can use this library either as a Node.js module or inside an HTML page.
...
-    var ion = ionReader.stringValue();
-    ionReader.stepOut();
-    console.log(ion.concat(", ").concat(hello));
+    let ion = require("ion-js");
+   
+    let ionData = '{ greeting: "Hello", name: "Ion" }';
+    let value = ion.load(ionData);
+    console.log(value.greeting + ", " + value.name + "!");
     ```
+
```
pbcornell: it would be nice to have an example of using one of the `dump*()` methods, too

src/dom/README.md
```diff
@@ -0,0 +1,253 @@
...
+
+You can upgrade a JS value to an Ion value using the `from()` method: 
+
+```typescript
+let i = ion.dom.Value.from(7);
+assert.equal(IonTypes.INT, i.getType());
+assert.equal(7, i.numberValue());
+
+let s = ion.dom.Value.from("foo", ["bar", "baz"]); // from() accepts an optional list of annotations
+assert.equal(IonTypes.STRING, s.getType());
```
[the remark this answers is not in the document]
zslayton: This was a typo on my part; the example was meant to be `ion.dom.Value.from('"foo"')`, producing a string. Good catch!

src/dom/README.md
zslayton: Done.

README.md
zslayton: Added.

src/dom/README.md
```diff
@@ -0,0 +1,253 @@
...
+```typescript
+// `s` is a `dom.String`, which implements the `Value` interface.
+let s: Value = ion.load('"foo"');
+
+// The Typescript compiler won't allow a direct comparison of a dom.String and a JS string, so
+// we'll need to down-convert our Ion value before comparing them. We can do this either by calling 
+// the Value interface's stringValue() method:
+assert.equal("foo", s.stringValue()); 
+// or concatenate our dom.String with the empty string, which is more concise but less explicit:
+assert.equal("foo", +s);
```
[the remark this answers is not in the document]
zslayton: Yes it should.

Issue #595, opened 2020-03-27
Timestamps created from a Date have an incorrect timezone offset


zslayton:
The [`timestamp` section](http://amzn.github.io/ion-docs/docs/binary.html#6-timestamp) of the binary spec says
> Timestamp representations have 7 components, where 5 of these components are optional depending on the precision of the timestamp. The 2 non-optional components are offset and year. The 5 optional components are (from least precise to most precise): month, day, hour and minute, second, fraction_exponent and fraction_coefficient. **All of these 7 components are in Universal Coordinated Time (UTC).**

> **The offset denotes the local-offset portion of the timestamp, in minutes difference from UTC.**

(Emphasis mine.)

The method that creates an `ion.Timestamp` from a `Date` attempted to follow this description, [storing the UTC values of each time unit in the `Date` and then asking for its timezone offset](https://github.com/amzn/ion-js/blob/6e43ff42b595f08bcc890b81b7cd045e1fc3f1b1/src/dom/Timestamp.ts#L42-L50). However, the `getTimezoneOffset()` method on Javascript's `Date` object does not return an offset that is specific to that `Date` object; it returns the host system's configured timezone offset. If a timezone offset is specified when the `Date` is created, that offset is discarded when the constructor finishes running.

```javascript
// Specify a datetime with offset UTC+17 hours
let date = new Date('1970-01-01T00:00:00.000+17:00');

// This fails if the host system is not configured with a local time of UTC+17 hours
assert.equal(date.getTimezoneOffset(), -1020);

// We can only request values in terms of the host's configured timezone or UTC; the
// original timezone specified at creation time has been discarded.
assert.equal(date.geUTCHours(), 7); // UTC time

// Requesting the non-UTC hours returns the host's local time, not the time 
// zone specified at creation
assert.equal(date.getHours(), 23); // pacific time
```

This is currently causing `dom.Timestamp`s created from a `Date` object to have an offset that conflicts with the intended datetime. This issue was not detected by our existing round-trip tests for `dom.Timestamp` because the error occurred before the value was written out. The wrong value was written out and the same wrong value was read in, satisfying the round trip test.

We need to modify `dom.Timestamp` to either:
1. Always convert `Date` to `Timestamp` as a UTC value
2. Always convert `Date` to `Timestamp` using the client system's configured local timezone.

[...]

PR #596, opened 2020-03-27
Fixed timezone issue in dom.Timestamp

zslayton:
*Issue #, if available:* #595 

*Description of changes:*

* When creating a `dom.Timestamp` from a provided `Date`, stores the UTC datetime represented by the `Date`. (See issue #595 for an explanation of why storing the `Date`'s offset is not meaningful.)
* Modifies an existing unit test to catch this sort of error in the event of a regression.


zslayton:
>  can we still assert `domValue.dateValue()` behavior is correct?

I removed this check because `Value.from(myDate).dateValue()` is guaranteed to be `myDate`. However, I only know this is true because I wrote the implementation; it certainly doesn't hurt to test it. I've re-added the check.
 
> the new comment suggests passing in an ion.Timestamp -- do we have a test for this?

Yes, in a few places. For example, [these round-trip tests](https://github.com/amzn/ion-js/blob/42ce6a42a9293605f194049ee8b12146f822b89f/test/dom/Value.ts#L361-L363) create `dom.Timestamp`s from `ion.Timestamps`.

> we have a test for offset Z, but I'd suggest having a test for a non-zero offset

I've added a bunch of example `Date` tests to check for this and other corner cases. In the process, I fixed an issue that could crop with fractional seconds (which, it turns out, `ion.Timestamp` requires be specified as an `ion.Decimal`). 👍

pbcornell:
> In the process, I fixed an issue that could crop with fractional seconds (which, it turns out, `ion.Timestamp` requires be specified as an `ion.Decimal`). 👍

Ok great...but now I'm looking for tests that assert expected behavior for non-zero milliseconds (for Date), nor any tests for micro-/nano-seconds (when using Timestamp) and I don't see any.

zslayton:
Changes:
* I refactored `exampleValues.ts` so it now has a list of ISO-8601 strings that is used to provide arrays of both sample `Date`s and `Timestamp`s.
* I added lots of different fractional-second ISO-8601 examples.
* I changed some existing `Timestamp` unit tests to use the new `exampleTimestampsWhere()` values, which should be much more comprehensive.

test/exampleValues.ts
```diff
@@ -41,4 +41,49 @@ export function exampleIonValuesWhere(filter: (v: Value) => boolean = acceptAnyV
...
+    "2020-03-28T01:00:00.123456+01:00",
+    "2020-03-28T01:00:00.123456789+01:00",
+];
+
+// A common collection of Date values that can be reduced to a relevant subset using
+// the provided filter function.
+export function exampleDatesWhere(filter: (v: Date) => boolean = acceptAnyValue): Date[] {
+    return _exampleIsoStrings
+    .map((isoString) => new Date(isoString))
+    .filter(filter);
```
pbcornell: indent

PR #607, opened 2020-04-21
Provides friendlier decimal text serialization

### LINKED ISSUE #606: Provide a more friendly text serialization for decimals
The text writer currently serializes a decimal such as `41.2` as `412d-1`.  While this is technically correct, it is not particularly friendly to a human reader.

I propose we update the logic to follow that used in ion-java's [_Private_IonTextAppender.printDecimal()](https://github.com/amzn/ion-java/blob/master/src/com/amazon/ion/impl/_Private_IonTextAppender.java#L719).

pbcornell:
Resolves #606 

Follows the logic used in ion-java's [_Private_IonTextAppender.printDecimal()](https://github.com/amzn/ion-java/blob/master/src/com/amazon/ion/impl/_Private_IonTextAppender.java#L719).

### DIFF
--- a/src/IonTextWriter.ts
+++ b/src/IonTextWriter.ts
@@ -54,6 +54,8 @@ export class Context {
     }
 }
 
+let _jsbiZero = JSBI.BigInt(0);
+
 export class TextWriter extends AbstractWriter {
 
     protected containerContext: Context[];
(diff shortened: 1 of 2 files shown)


src/IonTextWriter.ts
```diff
@@ -54,6 +54,8 @@ export class Context {
     }
 }
 
+let _jsbiZero = JSBI.BigInt(0);
```
zslayton, line 57: A constant for this is [available in in `JsbiSupport`](https://github.com/amzn/ion-js/blob/fa1fe7a6a8590d9fc5647f3a123495e7bad2a4d3/src/JsbiSupport.ts#L22).

pbcornell: thank you :)

PR #611, opened 2020-05-04
Updates IonTextReader.annotations() to avoid returning special annotation strings within single quotes

### LINKED ISSUE #610: IonTextReader.annotations() returns some symbols within single quotes
In Ion text encoding, some annotations must be enclosed within single quotes.  However, `IonTextReader.annotations()` should not treat such quotes as part of the value; rather, the quotes are part of the text encoding.

For example, for the value `'$3'::x`, the annotation returned by `IonTextReader.annotations()` is the JavaScript string `"'$3'"`, but it should be `"$3"`.

pbcornell:
* introduces `SymbolToken` class, currently only used internally
* IonParserTextRaw's `ann` array is now of type `SymbolToken[]` instead of `any[]` (well, `string[]` in practice)
* removes logic that would incorrectly identify `'$1.00'` as SID 1, in favor of a stronger test

Resolves #610 

### DIFF
--- a/src/IonSymbolToken.ts
+++ b/src/IonSymbolToken.ts
@@ -14,8 +14,9 @@
  */
 
 /**
- * An Ion symbol token (field name, annotation, and symbol values)
- * providing both the symbol text and the assigned symbol ID.
+ * An Ion symbol token (used to represent field names, annotations,
+ * and symbol values) providing both the symbol text and the assigned
+ * symbol ID.
  */
 export class SymbolToken {
     private static _UNKNOWN_SYMBOL_ID = -1;
--- a/src/IonSymbolToken.ts
+++ b/src/IonSymbolToken.ts
@@ -1,0 +1,41 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+/**
+ * An Ion symbol token (field name, annotation, and symbol values)
+ * providing both the symbol text and the assigned symbol ID.
+ */
+export class SymbolToken {
+    private static _UNKNOWN_SYMBOL_ID = -1;
+
+    constructor(private text: string | null,
+                private sid: number = SymbolToken._UNKNOWN_SYMBOL_ID) {
+    }
+
+    /**
+     * Returns the text of this symbol, or null if the text is unknown.
+     */
+    public getText(): string | null {
+        return this.text;
+    }
+
+    /**
+     * Returns the symbol ID (sid) of this symbol, or -1 if the sid is unknown.
+     */
+    public getSid(): number {
+        return this.sid;
+    }
+}
+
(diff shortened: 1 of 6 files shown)


src/IonTextReader.ts
```diff
@@ -217,20 +217,19 @@ export class TextReader implements Reader {
...
-            ann.push(str);
-        }
-        for (let i = 0; i < ann.length; i++) {
-            if(ann[i].length > 1 && ann[i][0] === '$') {
-                let tempStr = ann[i].substr(1, ann[i].length);
-                if (+tempStr === +tempStr) {//look up sid, +str === +str is a one line is integer hack
-                    let symbol = this._symtab.getSymbolText(Number(tempStr));
-                    if(symbol === undefined || symbol === null) throw new Error("Unresolvable symbol ID, symboltokens unsupported.");
-                    ann[i] = symbol;
+        let ann: string[] = [];
```
zslayton, line 220: Rather than initializing a new array and populating it with `.forEach()`, I suggest using [`.map()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map).

src/IonTextReader.ts
```diff
@@ -217,20 +217,19 @@ export class TextReader implements Reader {
...
+        this._parser.annotations().forEach((st) => {
+            let text = st.getText();
+            if (text !== null) {
+                ann.push(text);
+            } else {
+                let symbol = this._symtab.getSymbolText(st.getSid());
+                if (symbol === undefined || symbol === null) {
+                    throw new Error("Unresolvable symbol ID, symboltokens unsupported.");
```
zslayton, line 228: I know this isn't your code change, but: I wonder if we should be returning unresolvable symbols as `null` strings instead of throwing an `Error`. That would allow users to at least read the other values returned by `annotations()`  if they're resolvable.

src/IonSymbolToken.ts
```diff
@@ -0,0 +1,41 @@
...
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+/**
+ * An Ion symbol token (field name, annotation, and symbol values)
```
zslayton, line 17: I suggest tweaking this to

> (used to represent field names, annotations, and symbol values)

src/util.ts
```diff
@@ -36,3 +36,17 @@ export function _assertDefined(value: any): void {
         throw new Error("Expected value to be defined");
     }
 }
+
+/**
+ * If s consists solely of chars ['0'..'9'] and can be parsed as an int,
+ * returns the int value;  otherwise returns NaN.
+ */
+export function _toInt(s: string): number {
```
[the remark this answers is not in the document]
pbcornell: good point; moved into IonParserTextRaw, renamed to _parseSymbolId, and updated it to include testing the first char for `'$'`.

src/IonTextReader.ts
pbcornell: huh, this is my first `map()` in TS/JS :)

src/IonTextReader.ts
pbcornell: ion-hash-js currently expects an exception to be thrown, otherwise it will simply proceed and return an incorrect hash.  `$0` is an edge case, and I'm not inclined to change this behavior until we just go ahead and support symbol tokens.  Additionally, ion-java's `IonReader.getTypeAnnotations()` throws when it encounters `$0` (but `getTypeAnnotationSymbols()` doesn't); that's probably a good precedent to follow.

src/IonSymbolToken.ts
pbcornell: done

src/IonTextReader.ts
zslayton: > Additionally, ion-java's IonReader.getTypeAnnotations() throws when it encounters $0 (but getTypeAnnotationSymbols() doesn't); that's probably a good precedent to follow.

Yeah, this was my motivation. In the absence of a `SymbolToken` API, it seemed unfortunate that the only available method would `throw`. However, if `ion-hash-js` depends on this behavior I'm fine with leaving it in place.

Issue #622, opened 2020-07-17
Add support for deleting a key in Struct


alexis779:
There is no **delete** method in [Struct](https://amzn.github.io/ion-js/api/classes/_dom_struct_.struct.html) class.

To delete a Struct key, I run those steps:

- parse ION
- convert to JSON
- delete key on Object
- convert back to ION

_test_ion_struct_delete_key.js_ content:
```Javascript
const ion = require('ion-js');

var ionStr = '[ { key: "value" } ]';
console.log(ionStr);

const list = ion.load(ionStr);

var struct = list[0];
struct = JSON.parse(JSON.stringify(struct))
delete struct['key'];
list[0] = ion.load(JSON.stringify(struct))

ionStr = ion.dumpText(list);
console.log(ionStr);
```

Output:
```
$ node test_ion_struct_delete_key.js
[ { key: "value" } ]
[{}]
```

zslayton:
Addressed in #623.

PR #623, opened 2020-07-28
add deleteField method for deleting keys of Struct type

desaikd:
*Issue #, if available:*   #622 

*Description of changes:*
added a deleteField method which will delete the keys for Struct type. 
added unit-tests to check if deleteField works with built-in and user defind properties.

### DIFF
--- a/src/dom/Struct.ts
+++ b/src/dom/Struct.ts
@@ -62,10 +62,10 @@ export class Struct extends Value(Object, IonTypes.STRUCT, FromJsConstructor.NON
             },
             deleteProperty: function (target, name): boolean {
                 // Property is deleted only if it's in _field Collection
-                // reference Document: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete#Return_value
                 if (name in target._fields) {
                     delete target._fields[name];
                 }
+                // Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete#Return_value
                 return true;
             }
         });
--- a/src/dom/Struct.ts
+++ b/src/dom/Struct.ts
@@ -62,6 +62,7 @@ export class Struct extends Value(Object, IonTypes.STRUCT, FromJsConstructor.NON
             },
             deleteProperty: function (target, name): boolean {
                 // Property is deleted only if it's in _field Collection
+                // reference Document: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete#Return_value
                 if (name in target._fields) {
                     delete target._fields[name];
                 }
--- a/src/dom/Struct.ts
+++ b/src/dom/Struct.ts
@@ -59,6 +59,13 @@ export class Struct extends Value(Object, IonTypes.STRUCT, FromJsConstructor.NON
                     return target[name];
                 }
                 return target._fields[name];
+            },
+            deleteProperty: function (target, name): boolean {
+                // Property is deleted only if it's in _field Collection
+                if (name in target._fields) {
+                    delete target._fields[name];
+                }
+                return true;
             }
         });
     }
@@ -113,6 +120,14 @@ export class Struct extends Value(Object, IonTypes.STRUCT, FromJsConstructor.NON
             value.writeTo(writer);
         }
         writer.stepOut();
+    }
+
+    deleteField(name: string): boolean {
+        if (name in this._fields) {
+            delete this._fields[name];
+            return true;
+        }
+        return false;
     }
 
     toJSON() {
--- a/src/dom/Value.ts
+++ b/src/dom/Value.ts
@@ -140,8 +140,8 @@ export interface Value {
     writeTo(writer: Writer): void;
 
     /**
-     * For the Struct type, deletes the key/field if it's in the _fields collection
-     * otherwise throws an Error.
+     * For the Struct type, deletes the field with provided name if present. Returns a boolean
+     * indicating whether the Struct was modified. For all other types, throws an Error.
      */
     deleteField(name: string): boolean;
 }
@@ -296,7 +296,7 @@ export function Value<Clazz extends Constructor>(BaseClass: Clazz, ionType: IonT
         }
 
         deleteField(name: string): boolean {
-            this._unsupportedOperation('delete');
+            this._unsupportedOperation('deleteField');
         }
 
         // Returns the IonType associated with a particular dom.Value subclass. Useful for testing.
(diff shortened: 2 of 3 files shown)


src/dom/Struct.ts
```diff
@@ -59,6 +59,13 @@ export class Struct extends Value(Object, IonTypes.STRUCT, FromJsConstructor.NON
                     return target[name];
                 }
                 return target._fields[name];
+            },
+            deleteProperty: function (target, name): boolean {
+                // Property is deleted only if it's in _field Collection
+                if (name in target._fields) {
+                    delete target._fields[name];
+                }
+                return true;
```
zslayton, line 68: Because we had to do some research to understand what `true` means in the context of the `delete` operator, I suggest adding a comment with a link to [the relevant documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete#Return_value).

src/dom/Value.ts
```diff
@@ -138,6 +138,12 @@ export interface Value {
      * @param writer    An Ion Writer to which the value should be written.
      */
     writeTo(writer: Writer): void;
+
+    /**
+     * For the Struct type, deletes the key/field if it's in the _fields collection
+     * otherwise throws an Error.
```
zslayton, line 144: Two minor thoughts here:
* This is a typedoc comment (like Javadoc, but for Typescript) and will appear in [the public-facing documentation](https://amzn.github.io/ion-js/api/interfaces/_dom_value_.value-1.html#fieldnames) that's generated during the release. Folks reading that won't be aware of the `_fields` collection, which is an implementation detail in the `Struct` class.
* While the field name is basically a key, I suggest using either the Ion terminology ("field name" + "field value") or the JS terminology ("property name" + "property value") for consistency.

```suggestion
     * For the Struct type, deletes the field with provided name if present. Returns a boolean 
     * indicating whether the Struct was modified. For all other types, throws an Error.
```

src/dom/Value.ts
```diff
@@ -289,6 +295,10 @@ export function Value<Clazz extends Constructor>(BaseClass: Clazz, ionType: IonT
             this._unsupportedOperation('writeTo');
         }
 
+        deleteField(name: string): boolean {
+            this._unsupportedOperation('delete');
```
zslayton, line 299: The rename from `delete` to `deleteField` missed this string:
```suggestion
            this._unsupportedOperation('deleteField');
```

test/dom/propertyShadowing.ts
```diff
@@ -1,19 +1,25 @@
 import {assert} from "chai";
 import {dom, IonTypes} from "../../src/Ion";
+import {Err} from "typedoc/dist/lib/utils/result";
```
zslayton, line 3: Is this import necessary? It looks like it's using a type from the `typedoc` package.

test/dom/propertyShadowing.ts
```diff
@@ -22,6 +28,29 @@ describe('dom.Struct property shadowing', () => {
...
+        assert.deepEqual(s.fieldNames(), ['getType', 'getAnnotations', 'fieldNames', 'toString', 'age']);
+
+        // deleteField Struct method
+        assert.isTrue(s.deleteField('age'));
+        assert.deepEqual(s.fieldNames(), ['getType', 'getAnnotations', 'fieldNames', 'toString']);
+
+        // delete for properties that match built-in
+        assert.equal(s.get('toString')!.numberValue(), 10);
+        assert.isTrue(delete s['toString']);
+        assert.equal(s.toString(),'{getType: baz, getAnnotations: 56, fieldNames: [dog, cat, mouse]}');
```
zslayton, line 42: This test might be somewhat fragile; I don't know how rigidly the behavior of `toString()` is defined by the ECMA spec. If a web browser or NodeJS decided to add newlines to the returned string, for example, this would break.

We can achieve the same thing by just making sure that `toString` is still present and is a function:
```suggestion
        assert.equal(typeof s.toString, "function");
```

test/dom/propertyShadowing.ts
```diff
@@ -22,6 +28,29 @@ describe('dom.Struct property shadowing', () => {
...
+        assert.deepEqual(s.fieldNames(), ['getType', 'getAnnotations', 'fieldNames']);
+
+        // delete for field that doesn't exist
+        assert.isFalse(s.deleteField('toString'));
+        assert.isFalse(s.deleteField('greetings'));
+        assert.isFalse(s.deleteField('name'));
+        assert.isUndefined(s['greetings']);
+
+        // deleteField will throw an error if it's called on a dom.Value that isn't a struct
+        assert.throws(() => l.deleteField("1"), Error);
```
zslayton, line 52: Since the list `l` is only used in this one check, could you move its definition down to just below the comment?

```suggestion
        // deleteField will throw an error if it's called on a dom.Value that isn't a struct
        let l = dom.Value.from([1, 2, 3]);
        assert.throws(() => l.deleteField("1"), Error);
```

src/dom/Struct.ts
```diff
@@ -62,6 +62,7 @@ export class Struct extends Value(Object, IonTypes.STRUCT, FromJsConstructor.NON
             },
             deleteProperty: function (target, name): boolean {
                 // Property is deleted only if it's in _field Collection
+                // reference Document: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete#Return_value
                 if (name in target._fields) {
                     delete target._fields[name];
                 }
```
zslayton: Because this comment explains the meaning of the return value, I think it would be better to have it immediately above the `return true` on line 69.

```suggestion
                if (name in target._fields) {
                    delete target._fields[name];
                }
                // Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete#Return_value
```

If you agree with this suggestion, note that you can accept the change directly from the GitHub UI using the "commit suggestion" button above. You don't need to make another commit and push it.

src/dom/Struct.ts
```diff
@@ -59,6 +59,14 @@ export class Struct extends Value(Object, IonTypes.STRUCT, FromJsConstructor.NON
...
                 return target._fields[name];
+            },
+            deleteProperty: function (target, name): boolean {
+                // Property is deleted only if it's in _field Collection
+                // reference Document: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete#Return_value
+                if (name in target._fields) {
+                    delete target._fields[name];
+                }
+                return true;
```
zslayton, line 69: Because this comment explains the meaning of the return value, I think it would be better to have it immediately above the `return true` on line 69.

```suggestion
                if (name in target._fields) {
                    delete target._fields[name];
                }
                // Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete#Return_value
                return true;
```

If you agree with this suggestion, note that you can accept the change directly from the GitHub UI using the "commit suggestion" button above. You don't need to make another commit and push it.

PR #624, opened 2020-08-06
Implement process for ion-js cli

desaikd:
**Description:**

As part of ion-test-driver, this commit is for ion-js-cli process. It supports:

 - --output(-o) : rewrite given input files to specified output format
 - --output-fomrat(-f): specify ion output format
 - --error-report(-e): generate error report to specified location 

 
**Completeted Tasks:**

1. Implement processFiles function for binary/text/pretty formats to rewrite given input files.
2. Implement processEvents for generating event stream out of ion stream.
3. Supports help menu with all options and usage
4. Supports multiple input files rewritten into same output file

**TO DO:**

- [X] Support symbol[READ | WRITE | STATE] for error-report.
- [X] Support multiple input files for error-report.
- [X] Support None output format 
- [ ] Implement other options --catalog, --imports, --filter

**Test:**

tested using multiple input files for all output formats and events.
e.g. ts-node process example.ion --output-format text --output output.ion --error-report error_report

### DIFF
--- a/src/Cli.ts
+++ b/src/Cli.ts
@@ -119,11 +119,11 @@ export class IonComparisonReport {
     comparisonReportFile: Writable;
     comparisonType: ComparisonType;
 
-    constructor(lhs: ComparisonContext, rhs: ComparisonContext, comparisonReportFile: Writable, comprisonType: ComparisonType) {
+    constructor(lhs: ComparisonContext, rhs: ComparisonContext, comparisonReportFile: Writable, comparisonType: ComparisonType) {
         this.lhs = lhs;
         this.rhs = rhs;
         this.comparisonReportFile = comparisonReportFile;
-        this.comparisonType = comprisonType;
+        this.comparisonType = comparisonType;
     }
 
     writeComparisonReport(result: ComparisonResultType, message: string, event_index_lhs: number, event_index_rhs: number) {
--- a/src/Cli.ts
+++ b/src/Cli.ts
@@ -6,12 +6,14 @@ import {ComparisonContext, ComparisonType} from "./Compare";
 
 /** common CLI arguments structure
  *
- * @params outputFile,errorReportFile : 'any' datatype to use stdout as well as file write stream to write ion data
+ * @params
  * */
 export class IonCliCommonArgs {
     inputFiles: Array<string>;
-    outputFile: any;
     outputFormatName: OutputFormat;
+
+    // outputFile,errorReportFile: 'any' datatype to use stdout as well as file write stream to write ion data
+    outputFile: any;
     errorReportFile: any;
 
     constructor(argv: yargs.Arguments) {
--- a/src/Cli.ts
+++ b/src/Cli.ts
@@ -2,15 +2,17 @@ import * as yargs from "yargs";
 import fs from "fs";
 import {OutputFormat} from "./OutputFormat";
 import {IonTypes, makeTextWriter} from "./Ion";
-import {Writable} from "stream";
 import {ComparisonContext, ComparisonType} from "./Compare";
 
-/** common CLI arguments structure */
+/** common CLI arguments structure
+ *
+ * @params outputFile,errorReportFile : 'any' datatype to use stdout as well as file write stream to write ion data
+ * */
 export class IonCliCommonArgs {
     inputFiles: Array<string>;
-    outputFile: Writable;
+    outputFile: any;
     outputFormatName: OutputFormat;
-    errorReportFile: Writable;
+    errorReportFile: any;
 
     constructor(argv: yargs.Arguments) {
         this.outputFile = argv["output"] ? fs.createWriteStream(argv["output"] as string, {flags: 'w'}) : process.stdout;
@@ -20,11 +22,11 @@ export class IonCliCommonArgs {
         this.inputFiles = argv["input-file"] as Array<string>;
     }
 
-    getOutputFile(): Writable {
+    getOutputFile(): any {
         return this.outputFile;
     }
 
-    getErrorReportFile(): Writable {
+    getErrorReportFile(): any {
         return this.errorReportFile;
     }
 
@@ -116,10 +118,10 @@ export class ComparisonResult {
 export class IonComparisonReport {
     lhs: ComparisonContext;
     rhs: ComparisonContext;
-    comparisonReportFile: Writable;
+    comparisonReportFile: any;
     comparisonType: ComparisonType;
 
-    constructor(lhs: ComparisonContext, rhs: ComparisonContext, comparisonReportFile: Writable, comparisonType: ComparisonType) {
+    constructor(lhs: ComparisonContext, rhs: ComparisonContext, comparisonReportFile: any, comparisonType: ComparisonType) {
         this.lhs = lhs;
         this.rhs = rhs;
         this.comparisonReportFile = comparisonReportFile;
--- a/src/Cli.ts
+++ b/src/Cli.ts
@@ -1,5 +1,5 @@
-import yargs from "yargs";
-import fs, {WriteStream} from "fs";
+import * as yargs from "yargs";
+import fs from "fs";
 import {OutputFormat} from "./OutputFormat";
 import {IonTypes, makeTextWriter} from "./Ion";
 import {Writable} from "stream";
--- a/src/Cli.ts
+++ b/src/Cli.ts
@@ -13,6 +13,7 @@ export class IonCliCommonArgs {
     outputFormatName: OutputFormat;
 
     // outputFile,errorReportFile: 'any' datatype to use stdout as well as file write stream to write ion data
+    // for more information : https://github.com/amzn/ion-js/issues/627
     outputFile: any;
     errorReportFile: any;
 
--- a/src/Cli.ts
+++ b/src/Cli.ts
@@ -2,13 +2,14 @@ import yargs from "yargs";
 import fs, {WriteStream} from "fs";
 import {OutputFormat} from "./OutputFormat";
 import {IonTypes, makeTextWriter} from "./Ion";
+import {Writable} from "stream";
 
 /** common CLI arguments structure */
 export class IonCliCommonArgs {
     inputFiles: Array<string>;
-    outputFile: WriteStream | NodeJS.WriteStream;
+    outputFile: Writable;
     outputFormatName: OutputFormat;
-    errorReportFile: WriteStream | NodeJS.WriteStream;
+    errorReportFile: Writable;
 
     constructor(argv: yargs.Arguments) {
         this.outputFile = argv["output"] ? fs.createWriteStream(argv["output"] as string, {flags: 'w'}) : process.stdout;
@@ -18,11 +19,11 @@ export class IonCliCommonArgs {
         this.inputFiles = argv["input-file"] as Array<string>;
     }
 
-    getOutputFile(): WriteStream | NodeJS.WriteStream {
+    getOutputFile(): Writable {
         return this.outputFile;
     }
 
-    getErrorReportFile(): WriteStream | NodeJS.WriteStream {
+    getErrorReportFile(): Writable {
         return this.errorReportFile;
     }
 
--- a/src/Cli.ts
+++ b/src/Cli.ts
@@ -1,27 +1,28 @@
 import yargs from "yargs";
-import fs from "fs";
+import fs, {WriteStream} from "fs";
 import {OutputFormat} from "./OutputFormat";
+import {IonTypes, makeTextWriter} from "./Ion";
 
 /** common CLI arguments structure */
 export class IonCliCommonArgs {
     inputFiles: Array<string>;
-    outputFile: any;
+    outputFile: WriteStream | NodeJS.WriteStream;
     outputFormatName: OutputFormat;
-    errorReportFile: any;
+    errorReportFile: WriteStream | NodeJS.WriteStream;
 
     constructor(argv: yargs.Arguments) {
-        this.outputFile = argv["output"] ? fs.createWriteStream(<string>argv["output"], {flags: 'w'}) : process.stdout;
+        this.outputFile = argv["output"] ? fs.createWriteStream(argv["output"] as string, {flags: 'w'}) : process.stdout;
         // create error output stream (DEFAULT: stderr)
-        this.errorReportFile = argv["error-report"] ? fs.createWriteStream(<string>argv["error-report"], {flags: 'w'}) : process.stderr;
-        this.outputFormatName = <OutputFormat>argv["output-format"];
-        this.inputFiles = <Array<string>>argv["inputfile"];
+        this.errorReportFile = argv["error-report"] ? fs.createWriteStream(argv["error-report"] as string, {flags: 'w'}) : process.stderr;
+        this.outputFormatName = argv["output-format"] as OutputFormat;
+        this.inputFiles = argv["input-file"] as Array<string>;
     }
 
-    getOutputFile(): any {
+    getOutputFile(): WriteStream | NodeJS.WriteStream {
         return this.outputFile;
     }
 
-    getErrorReportFile(): any {
+    getErrorReportFile(): WriteStream | NodeJS.WriteStream {
         return this.errorReportFile;
     }
 
@@ -44,22 +45,43 @@ export enum ErrorType {
 /** Error structure for error report */
 export class IonCliError {
     errorType: ErrorType;
+    eventIndex: number;
     location: string;
     message: string;
     errorReportFile: any;
 
-    constructor(errorType: ErrorType, location: string, message: string, errorReportFile: any) {
+    constructor(errorType: ErrorType, location: string, message: string, errorReportFile: any, eventIndex: number = 0) {
         this.errorType = errorType;
         this.location = location;
         this.message = message;
         this.errorReportFile = errorReportFile;
+        this.eventIndex = eventIndex;
     }
 
     writeErrorReport() {
-        this.errorReportFile.write(this.errorType + ": " + this.location + " " + this.message + "\n");
+        let writer = makeTextWriter();
+        writer.stepIn(IonTypes.STRUCT);
+        writer.writeFieldName('error_type');
+        writer.writeSymbol(this.errorType);
+        writer.writeFieldName('message');
+        writer.writeString(this.message);
+        writer.writeFieldName('location');
+        writer.writeString(this.location);
+        writer.writeFieldName('event_index');
+        writer.writeInt(this.eventIndex);
+        writer.stepOut();
+        this.errorReportFile.write(writer.getBytes());
+        this.errorReportFile.write("\n");
     }
 }
 
+/**
+ * This will cause `yargs` to look in other .ts files in the same directory for command modules.
+ *
+ * For more information, see:
+ * Command modules: https://github.com/yargs/yargs/blob/master/docs/advanced.md#providing-a-command-module
+ * commandDir: https://github.com/yargs/yargs/blob/master/docs/advanced.md#commanddirdirectory-opts
+ */
 const argv = yargs
     .commandDir(__dirname,{
         extensions: ['ts'],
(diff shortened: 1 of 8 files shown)


src/Process.ts
```diff
@@ -0,0 +1,133 @@
...
+            + "'process' command, and outputs a serialized EventStream\n"
+            + "representing the input Ion stream(s)."
+    },
+    'error-report': {
+        alias: 'e',
+        description: 'Error report file',
+    }
+}).argv;
+
+class ProcessArgs {
```
tgregg, line 28: These args will apply to other commands too, not just process.

src/Process.ts
```diff
@@ -0,0 +1,133 @@
...
+        return this.inputFiles;
+    }
+
+    getOutputFormatName(): OutputFormat{
+        return <OutputFormat> this.outputFormatName;
+    }
+}
+
+/** Error structure for error report */
+class ProcessError {
```
tgregg, line 61: This ErrorReport structure applies to all commands, not just process.

src/Process.ts
```diff
@@ -0,0 +1,133 @@
...
+    }
+}
+
+/**
+ *  Read the input file(s) and re-write in the format
+ *  specified by --output-format
+ *
+ *  Create error report through --error-report and output file specified by --output
+ */
+class Process {
```
tgregg, line 85: Process is just one command available via the CLI. Compare is another. They should be accessible via the same executable. Will this support that?

src/Process.ts
```diff
@@ -0,0 +1,133 @@
...
+    'error-report': {
+        alias: 'e',
+        description: 'Error report file',
+    }
+}).argv;
+
+class ProcessArgs {
+    inputFiles: Array<string>;
+    outputFile: any;
+    outputFormatName: string;
```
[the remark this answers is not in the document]
desaikd: Yes. Changing it in next commit.

src/Process.ts
desaikd: No. This will not support that. Hence modularizing the code by creating another class to support both compare and process.

src/Cli.ts
```diff
@@ -0,0 +1,89 @@
...
+    writeErrorReport() {
+        this.errorReportFile.write(this.errorType + ": " + this.location + " " + this.message + "\n");
+    }
+}
+
+const argv = yargs
+    .commandDir(__dirname,{
+        extensions: ['ts'],
+    })
+    .options({
```
tgregg, line 67: How do you differentiate between the process command and other commands?

src/OutputFormat.ts
```diff
@@ -0,0 +1,33 @@
...
+            case OutputFormat.PRETTY:
+                return makePrettyWriter();
+            case OutputFormat.TEXT:
+                return makeTextWriter();
+            case OutputFormat.BINARY:
+                return makeBinaryWriter();
+            case OutputFormat.EVENTS:
+                return makePrettyWriter();
+            case OutputFormat.NONE:
+                return makePrettyWriter();
```
tgregg, line 28: It looks like makePrettyWriter is just a placeholder here because NONE causes nothing to be written. Is that right? If that's the case, is there a more obvious placeholder that could be used (like null)? I think a comment would be helpful, as a minimum.

src/Process.ts
```diff
@@ -0,0 +1,104 @@
...
+        } catch(Error) {
+            this.clearIonOutputWriter(ionOutputWriter);
+            new IonCliError(ErrorType.WRITE, path, Error.message, args.getErrorReportFile()).writeErrorReport();
+        }
+    }
+
+    processFiles(ionOutputWriter: Writer, args: IonCliCommonArgs): void {
+        for (let path of args.getInputFiles()) {
+                this.processFile(ionOutputWriter, args, path);
+            }
```
tgregg, line 84: It looks like these two lines ended up with one extra level of indentation.

src/Cli.ts
desaikd: To differentiate between commands I have used a constant called command and a handler for the command file like Process.ts(8-16) which has a handler stating to call process.(Yargs support having a command directory in which you can have command files specified using a command and a handler constants).

src/Process.ts
```diff
@@ -0,0 +1,104 @@
+import fs from 'fs';
+import {OutputFormat} from './OutputFormat';
+import {IonEventStream} from "./IonEventStream";
+import {Writer} from "./IonWriter";
+import {ErrorType, IonCliCommonArgs, IonCliError} from "./Cli";
+import {makeReader, Reader} from "./Ion";
+
```
zslayton, line 7:
```suggestion
// The `command`, `describe`, and `handler` exports below are part of the yargs command module API
// See: https://github.com/yargs/yargs/blob/master/docs/advanced.md#providing-a-command-module
```

src/OutputFormat.ts
```diff
@@ -0,0 +1,33 @@
+import {Writer} from './IonWriter';
+import {
+    makePrettyWriter,
+    makeBinaryWriter,
+    makeTextWriter
+} from './Ion';
+
+export enum OutputFormat {
```
zslayton, line 8: This could use a `typedoc` comment.

src/Process.ts
```diff
@@ -0,0 +1,104 @@
...
+    "the format specified by --output.";
+
+export const handler = function (argv) {
+    let args = new IonCliCommonArgs(argv);
+    new Process(args);
+}
+
+/**
+ *  Read the input file(s) and re-write in the format
+ *  specified by --output-format
```
zslayton, line 20: Nit: The newline in the middle of this comment is probably unnecessary.

src/Cli.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    writeErrorReport() {
+        this.errorReportFile.write(this.errorType + ": " + this.location + " " + this.message + "\n");
+    }
+}
+
+const argv = yargs
+    .commandDir(__dirname,{
+        extensions: ['ts'],
+    })
```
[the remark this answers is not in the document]
desaikd: Yes. I think adding this would be a good idea tomato code readable.

src/Cli.ts
```diff
@@ -0,0 +1,89 @@
...
+    outputFile: any;
+    outputFormatName: OutputFormat;
+    errorReportFile: any;
+
+    constructor(argv: yargs.Arguments) {
+        this.outputFile = argv["output"] ? fs.createWriteStream(<string>argv["output"], {flags: 'w'}) : process.stdout;
+        // create error output stream (DEFAULT: stderr)
+        this.errorReportFile = argv["error-report"] ? fs.createWriteStream(<string>argv["error-report"], {flags: 'w'}) : process.stderr;
+        this.outputFormatName = <OutputFormat>argv["output-format"];
+        this.inputFiles = <Array<string>>argv["inputfile"];
```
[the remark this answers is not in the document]
desaikd: Yes. will change that in the next commit.

src/IonEventStream.ts
```diff
@@ -149,46 +157,67 @@ export class IonEventStream {
...
+                return;
+            }
+            if (tid === IonTypes.SEXP || IonTypes.LIST && this.reader.annotations()[0] === "embedded_documents") {
+                this.isEmbedded = true;
+            }
+            let currentContainer: IonEvent[] = [];
+            let currentContainerIndex: number[] = [];
+            while (true) {
+                if (this.reader.isNull()) {
+                    this.eventStream.push(this.eventFactory.makeEvent(IonEventType.SCALAR, tid!, this.reader.fieldName(), this.reader.depth(), this.reader.annotations(), true, this.reader.value()));
```
zslayton, line 174: Some of these lines are very wide; could you see if your IDE has an auto-formatter? 

Note: if you autoformat the code and it changes a huge number of lines, don't worry about it. We'll finish this PR and then we can open a separate PR to do sweeping, formatting-only changes. Once that's done we can figure out how to add formatting requirements to our build pipeline.

src/Cli.ts
```diff
@@ -1,27 +1,28 @@
...
-import fs from "fs";
+import fs, {WriteStream} from "fs";
 import {OutputFormat} from "./OutputFormat";
+import {IonTypes, makeTextWriter} from "./Ion";
 
 /** common CLI arguments structure */
 export class IonCliCommonArgs {
     inputFiles: Array<string>;
-    outputFile: any;
+    outputFile: WriteStream | NodeJS.WriteStream;
```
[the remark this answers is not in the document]
desaikd: https://nodejs.org/en/knowledge/advanced/streams/how-to-use-stream-pipe/. I am using process.stdout for printing to console if no output file specified. That is why I needed both writestreams.

src/IonEventStream.ts
```diff
@@ -105,6 +112,7 @@ export class IonEventStream {
 
             }
         }
+        writer.close();
```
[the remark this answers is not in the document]
desaikd: So for embedded streams it ends with stream_end for each stream embedded in that. I am not creating the writer every time but using the same for all of them.

src/Cli.ts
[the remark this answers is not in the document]
zslayton: Intuitively, I would expect there to be an interface for their common functionality that we could code to. The [documentation for writable streams](https://nodejs.org/api/stream.html#stream_writable_streams) seems to indicate that [`Writable`](https://nodejs.org/api/stream.html#stream_class_stream_writable) might fit that bill?

desaikd:
**Commit**: e200c9c

- Support compare command 
  -  `compare`  method for basic comparison 
  - `compareEquivs` for equivs, non-equips and equiv-timeline comparison.

**Commit** : fde92fc

- fixed suggested changes
- solved merge conflicts

**Commit**: 441a1c8

- write to file only once. (remove duplicate writes) 

**TO DO:**

- Investigate on the CI build fail.

desaikd:
**Commit**: ff1aa82

   

- `IonEventStream` 

   - added parsing for containers and eventstreams to `compareEquivs` method. (`parseContainer` and `parseEventStream`)
   - added `equivs` which return boolean answer from `compareEquivs`.
   - modified the `compareEquivs` signature to take `comparisonType` argument and made the `comparisonReport` optional.
- `equivsTest`
  - used `equivs` inside this method in` iontests.ts`
- `IonEvent`
     - small bug fixes for comparison. (`valueCompare`)

src/IonEventStream.ts
```diff
@@ -266,7 +302,7 @@ export class IonEventStream {
                 this.marshalStream();
                 return;
             }
-            if (tid === IonTypes.SEXP || IonTypes.LIST && this.reader.annotations()[0] === "embedded_documents") {
+            if (tid === IonTypes.SEXP && this.reader.annotations()[0] === "embedded_documents") {
```
tgregg, line 305: Shouldn't this be

```
if ((tid === IonTypes.SEXP || tid === IonTypes.LIST) && this.reader.annotations()[0] === "embedded_documents") {
```

src/IonEventStream.ts
```diff
@@ -189,29 +191,39 @@ export class IonEventStream {
         return new ComparisonResult(ComparisonResultType.EQUAL);
     }
 
-    compareEquivs(expected: IonEventStream, comparisonReport: IonComparisonReport): ComparisonResult {
+    equivs(expected: IonEventStream, comparisonType: ComparisonType): boolean {
```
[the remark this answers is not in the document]
desaikd: Sure.

src/IonEventStream.ts
desaikd: yes. Will change

src/IonEventStream.ts
```diff
@@ -138,6 +138,8 @@ export class IonEventStream {
     compare(expected: IonEventStream, comparisonReport?: IonComparisonReport): ComparisonResult {
         let actualIndex: number = 0;
         let expectedIndex: number = 0;
+        if(this.eventStream.length != expected.eventStream.length)
+            return new ComparisonResult(ComparisonResultType.ERROR, "Both event streams of different length");
```
zslayton, line 142: Nit: could you wrap the body of the `if` in braces?

src/IonEventStream.ts
```diff
@@ -189,29 +191,39 @@ export class IonEventStream {
...
             } else if(actualEvent.eventType == IonEventType.STREAM_END || expectedEvent.eventType == IonEventType.STREAM_END) {
-                comparisonReport.writeComparisonReport(ComparisonResultType.ERROR,
-                    "Different number of comparison sets.", actualIndex, expectedIndex);
-                return new ComparisonResult(ComparisonResultType.ERROR);
+                    throw new Error("Different number of comparison sets.");
             } else if (!(actualEvent.ionType == IonTypes.LIST || actualEvent.ionType == IonTypes.SEXP)
                 || !(expectedEvent.ionType == IonTypes.LIST || expectedEvent.ionType == IonTypes.SEXP)) {
                 throw new Error("Comparison sets must be lists or s-expressions.");
-            } else if(this.isEmbedded as any ^ expected.isEmbedded as any) {
+            } else if((actualEvent.annotations[0] === 'embedded_documents') as any ^ (expectedEvent.annotations[0] === 'embedded_documents') as any) {
```
zslayton, line 213: > ```ts
> actualEvent.annotations[0] === 'embedded_documents') as any
> ```

Why do the boolean expressions need to be cast to `any`?

Since we check for 
```ts
someEvent.annotations[0] === 'embedded_documents'
```
several times in this code block, it might be worth making an `isEmbedded(event): boolean` helper function so the intent is obvious.

src/IonEventStream.ts
```diff
@@ -240,23 +252,47 @@ export class IonEventStream {
...
                             : new ComparisonResult(ComparisonResultType.NOT_EQUAL, ionTimestampActual + " vs. " + ionTimestampExpected);
                     } else {
-                         eventResult = actualContainerEvent.compare(expectedContainerEvent);
+                            eventResult = actualContainerEvent.compare(expectedContainerEvent);
                     }
 
                     if ((comparisonType == ComparisonType.EQUIVS || comparisonType == ComparisonType.EQUIV_TIMELINE)
                         && eventResult.result == ComparisonResultType.NOT_EQUAL) {
-                        comparisonReport.writeComparisonReport(ComparisonResultType.NOT_EQUAL, eventResult.message, i + 1, j + 1);
+                        if(comparisonReport)
```
zslayton, line 260: Same thought throughout regarding `if` braces.

src/IonEventStream.ts
```diff
@@ -240,23 +252,47 @@ export class IonEventStream {
...
-        return new ComparisonResult(comparisonType == ComparisonType.EQUIVS ? ComparisonResultType.EQUAL : ComparisonResultType.NOT_EQUAL);
+        return new ComparisonResult(comparisonType == ComparisonType.NON_EQUIVS ? ComparisonResultType.NOT_EQUAL : ComparisonResultType.EQUAL);
+    }
+
+    private parseContainer(event: IonEvent): IonEvent[] {
+        let container: IonEvent[] = [];
+        for (let j = 0; j < event.ionValue.length - 1; j++) {
+            container.push(event.ionValue[j]);
+            if (event.ionValue[j].eventType === IonEventType.CONTAINER_START) {
+                j += event.ionValue[j].ionValue.length
```
zslayton, line 282: Could you add a semicolon to this line for consistency? (We really should add an automated style checker on this repo.)

desaikd: Yes. As you mentioned earlier should we do it in a separate PR?

zslayton: > As you mentioned earlier should we do it in a separate PR?

I suggest making a new PR any time you find yourself adding substantial new functionality. For example, I think this PR could have been broken up into one PR for the `yargs` framework and another PR for the `compare`/`equivs` implementation.

Incorporating feedback (style changes, refactoring, etc) can all be done as new commits in the same PR.

src/IonEventStream.ts
desaikd: For XOR operation in TS required both types to be any or number. it doesn't work with type boolean somehow.
And yes, will add isEmbedded as a helper function.

desaikd:
**Commit**: 9c1d0d2

- `IonEventStream` 

   - changed the way embedded streams were handled.
   - added doctype comment for `equivs` and `compareEquivs`.
  
- `Process`
  - changed the way embedded streams were handled to make it work properly for `List`. (`writeToEventStream`)

- Previously suggested changes

src/IonEventStream.ts
```diff
@@ -206,17 +230,19 @@ export class IonEventStream {
             if(actualEvent.eventType == IonEventType.STREAM_END && expectedEvent.eventType == IonEventType.STREAM_END) {
                 break;
             } else if(actualEvent.eventType == IonEventType.STREAM_END || expectedEvent.eventType == IonEventType.STREAM_END) {
-                    throw new Error("Different number of comparison sets.");
+                throw new Error("Different number of comparison sets.");
             } else if (!(actualEvent.ionType == IonTypes.LIST || actualEvent.ionType == IonTypes.SEXP)
                 || !(expectedEvent.ionType == IonTypes.LIST || expectedEvent.ionType == IonTypes.SEXP)) {
                 throw new Error("Comparison sets must be lists or s-expressions.");
-            } else if((actualEvent.annotations[0] === 'embedded_documents') as any ^ (expectedEvent.annotations[0] === 'embedded_documents') as any) {
+            } else if(this.isEmbedded(actualEvent) as any ^ this.isEmbedded(expectedEvent) as any) {
```
zslayton, line 237: Why does the boolean expression `this.isEmbedded(...)` need to be cast `as any`?

src/IonEventStream.ts
```diff
@@ -206,17 +230,19 @@ export class IonEventStream {
...
                 throw new Error("Both streams should be embedded streams.");
             }
 
+            // both containers has type any as it maybe IonEventStream[] or IonEvent[]
+            // depending on the type compare method will be called
```
zslayton, line 242: Since both of these are pretty limited in scope, could we use the `IonEventStream[] | IonEvent[]` type union here? If this adds a lot of complexity, feel free to say no.

src/Process.ts
```diff
@@ -75,10 +77,33 @@ export class Process {
         try {
             eventStream = new IonEventStream(ionReader, path, args);
 
-            // processes input stream into event stream
-            eventStream.writeEventStream(ionOutputWriter);
+            ionOutputWriter.writeSymbol(this.EVENT_STREAM);
+            let writeEvents: IonEvent[] = [];
+            for(let i =0; i < eventStream.getEvents().length; i++) {
```
zslayton, line 82: Nit:

```suggestion
            for(let i = 0; i < eventStream.getEvents().length; i++) {
```

src/Process.ts
```diff
@@ -75,10 +77,33 @@ export class Process {
...
+                let event = eventStream.getEvents()[i];
+
+                //process event stream into event stream
+                if(eventStream.isEventStream) {
+                    writeEvents = eventStream.getEvents();
+                    break;
+                }
```
zslayton, line 89: It looks like this case could be lifted out above the `for` loop? Consider making a helper function for collecting input stream events into an array.

```ts
if (eventStream.isEventStream) {
   writeEvents = eventStream.getEvents();
} else {
   writeEvents = collectInputStreamEvents(...); // <-- new helper function
}

for(let event of writeEvents) {
   event.write(ionOutputWriter);
}
```

That will simplify this function's responsibilities and make it easier to understand.

src/Process.ts
```diff
@@ -75,10 +77,33 @@ export class Process {
...
+
+                // processes input stream(text or binary) into event stream
+                writeEvents.push(event);
+
+                if(event.annotations[0] == 'embedded_documents') {
```
zslayton, line 94: Another use case for `isEmbedded`?

src/IonEventStream.ts
desaikd: Yah. Actually I tried to do the Union type but it would add complexity to the method. So I haven't added union type here.

src/IonEventStream.ts
desaikd: ^ operator can only be applied to the number type, according to the TypeScript specification.

zslayton (approved):

tgregg (approved):

PR #635, opened 2020-10-20
write ion type in uppercase for event

desaikd:
*Description of changes:*

For ion-test-driver an event needs to write ion type in uppercase to have consensus between implementations.


cheqianh (approved):
Nice!

PR #638, opened 2020-10-30
Changed the parsing of embedded streams for equivs comparison

desaikd:
*Description of changes:*
This PR is about changing the parsing method for embedded streams when comparing with equivs type. This makes ion-js align with other implementations.

e.g.
```
// Symbols that contain the same text
embedded_documents::[
    '''
    $ion_1_0
    $ion_symbol_table::{
        symbols:[ "foo", "bar", "baz", "qux" ]
    }
    $10
    $11
    $12
    $13
    ''',
    '''
    $ion_1_0
    foo
    bar
    baz
    qux
    ''',
]
```
This equivs comparison results into true after changing the parsing method. 

### DIFF
--- a/test-driver/src/Compare.ts
+++ b/test-driver/src/Compare.ts
@@ -150,8 +150,8 @@ export class Compare {
             let expectedContainer: any = [];
             if (actual.isEmbedded(actualEvent) && expected.isEmbedded(expectedEvent)) {
                 //we found a list of strings that we need to interpret as top level ion text streams.
-                actualContainer = this.parseEmbeddedStream(actualEvent);
-                expectedContainer = this.parseEmbeddedStream(expectedEvent);
+                actualContainer = this.parseEmbeddedStream(actualEvent)
+                expectedContainer = this.parseEmbeddedStream(expectedEvent)
             } else {//we're in an sexp/list
                 actualContainer = this.parseContainer(actualEvent);
                 expectedContainer = this.parseContainer(expectedEvent);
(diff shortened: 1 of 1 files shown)


test-driver/src/Compare.ts
```diff
@@ -150,8 +150,8 @@ export class Compare {
             let expectedContainer: any = [];
             if (actual.isEmbedded(actualEvent) && expected.isEmbedded(expectedEvent)) {
                 //we found a list of strings that we need to interpret as top level ion text streams.
-                actualContainer = this.parseEmbeddedStream(actualEvent);
-                expectedContainer = this.parseEmbeddedStream(expectedEvent);
+                actualContainer = this.parseEmbeddedStream(actualEvent)
+                expectedContainer = this.parseEmbeddedStream(expectedEvent)
```
zslayton, line 154: I'm surprised that `tslint` didn't complain about the missing semicolons and that `prettier` didn't automatically put them back.

test-driver/src/Compare.ts
```diff
@@ -218,11 +218,16 @@ export class Compare {
     // parse embedded stream into events
     private parseEmbeddedStream(event: IonEvent): any {
         let container: any = [];
+        let value = "";
         for (let j = 0; j < event.ionValue.length - 1; j++) {
-            if(event.ionValue[j].eventType == IonEventType.STREAM_END)
+            if(event.ionValue[j].eventType == IonEventType.STREAM_END) {
```
zslayton, line 223: I also expected `prettier` to add a space between the `if` and the `(`.

desaikd: I just checked the prettier & tslint isn't added to check for test-driver and test folders. Seems like I need to create another PR for that.

zslayton: That explains it, thanks for checking!

cheqianh (approved):
Nice, it passed the file that causes issue.

I'll use ion-test-driver to run it again to see if this issue is fixed.

PR #641, opened 2020-11-04
Changes for test-driver clob and float events

desaikd:
*Description of changes:*
 This PR works on modifying IonEvent valueCompare() method for clob and float. 

1. Clobs need to consider length of clobs for equality. (5926a79)
2. Floats need to differentiate between +0 and -0. (ce2bb26)

This PR solves the problem of non-equivs comparison in clob and float events and aligns it with other implementations of test-driver. 

e.g.  Clob
```
[
    {{"hello"}},
    {{"hello "}},
    {{" hello"}},
    {{" hello "}},
]
```
Here, `{{"hello"}}`  and `{{"hello "}}` were being stated as `EQUAL` before. Now it shows them as `NOT_EQUAL` as they both have different
lengths.

e.g. Float
```
[
    0e0,
    -0e0,
]
```
Here, 0e0 and -0e0 were being stated as `EQUAL` before. Now it shows them as `NOT_EQUAL` as -0 and +0 are not equal.

### DIFF
--- a/src/events/IonEvent.ts
+++ b/src/events/IonEvent.ts
@@ -516,7 +516,7 @@ class IonFloatEvent extends AbstractIonEvent {
     }
     if (
       expected instanceof IonFloatEvent &&
-      this.ionValue === expected.ionValue
+      Object.is(this.ionValue, expected.ionValue)
     ) {
       return new ComparisonResult(ComparisonResultType.EQUAL);
     }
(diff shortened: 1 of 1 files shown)


src/events/IonEvent.ts
```diff
@@ -516,7 +516,7 @@ class IonFloatEvent extends AbstractIonEvent {
     }
     if (
       expected instanceof IonFloatEvent &&
-      this.ionValue === expected.ionValue
+      Object.is(this.ionValue, expected.ionValue)
```
zslayton, line 519: Now that we're using `Object#is`, [`NaN` will be considered equal to another `NaN`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is#Description). This means we can remove the special case `if` statement just above this one.

zslayton (approved):

cheqianh (approved):
Nice it passed the specific files.

After merging it into main branch, ion-test-driver will test it again.

PR #642, opened 2020-11-05
Fix async main branch change issue.

cheqianh:
Before we get main branch sha too early which will cause an async issue. Now I move this step right before ion-test-driver run.


desaikd (approved):
Nice!

PR #643, opened 2020-11-19
Update semantic-release version hint

jobarr-amzn:
*Issue #, if available:*  https://github.com/advisories/GHSA-r2j6-p67h-q639

*Description of changes:*

Tested via `npm install`, changing this version hint, and then another `npm install`. In both cases all unit tests succeeded.
```
npm install
```
*many lines of output elided*
```
  3078 passing (11s)
  170 pending
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "mocha-typescript": "^1.1.17",
     "nyc": "^14.1.1",
     "prettier": "2.1.2",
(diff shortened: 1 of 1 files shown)


tgregg (approved):
No need to wait for the ion-test-driver run to finish since this should not affect correctness. It will keep going after the PR is merged and it will open an issue if it fails.

Issue #645, opened 2020-11-19
Shared Symbol Tables resolve text to SIDs incorrectly


zslayton:
The `SharedSymbolTable` class has [a member field called `_idsByText`](https://github.com/amzn/ion-js/blob/217feef0ca0cd462f6909ead17f386579fdded0a/src/IonSharedSymbolTable.ts#L22) that contains a mapping of symbol text back to their corresponding symbol IDs:

```ts
protected readonly _idsByText: Map<string, number>;
```

`SharedSymbolTable`'s implementation of [`getSymbolId`](https://github.com/amzn/ion-js/blob/217feef0ca0cd462f6909ead17f386579fdded0a/src/IonSharedSymbolTable.ts#L65) looks like this:

```ts
getSymbolId(text: string): number | undefined {
    return this._idsByText[text];
}
```

Notice that it uses `[]`-syntax to index into `_idsByText`. 

`_idsByText` is a [`Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map), not an `Object`, which means we should access its key value pairs using[ the `get` method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/get) to avoid accidentally referring to properties on the `Map`'s prototype.

Using `[]`-indexing causes `SharedSymbolTable#getSymbolId` to unintentionally access `Map` prototype properties in the event of naming collisions.

For example, running:

```ts
const sid = sst.getSymbolId("size");
```

when "size" is not in the symbol table should return `undefined`. However, `Map`'s prototype provides [a `size` property](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/size):

> The value of `size` is an integer representing how many entries the Map object has. A set accessor function for size is undefined; you can not change this property.

which means that `sst.getSymbolId("size")` _will return a plausible-looking number_.

This issue intersects with #644, but both problems must be addressed independently.

zslayton:
Fixed in #648.

PR #648, opened 2020-11-24
SharedSymbolTables no longer treat object properties as symbols.

zslayton:
*Issue #, if available:* #645 

*Description of changes:*

Modifies `SharedSymbolTable` (used to represent imported SSTs and the system symbol table) so it avoids co-mingling symbols and object properties.


desaikd (approved):
Looks good!

Issue #654, opened 2020-11-30
Add getPosition() function to Reader


bingkh:
Hi, 

**Description**
As a user of ion-js, I would love to be able to get position information from ionReader, so that when using ionReader to read a ion file, I can know which position I'm currently at.

**Application scenario**
Our team uses ion to define a format, e.g. paragraph `P` uses certain format style `S`, and the definition of style `S` is out side of the paragraph itself. see blow:
```
style::{
  name:S,
  font:Arial,
  .....
}
....
....
paragraph::{
  name:A,
  style:S,
  .....
}
```
I'm working on a vs code extension: ion-style-peek, so that in the above ion file, when users are viewing paragraph `A` and call `go to definition` on `S`, the editor will jump to style `S`.  Similar implementation can be found: https://github.com/pranaygp/vscode-css-peek

To implement the above, we need to get the position when parsing style container in the ion, hence the requesting issue.

Pull request:
https://github.com/amzn/ion-js/pull/655

Thanks!

Ethan

zslayton:
Added in #657.

PR #655, opened 2020-11-30
Adding `postion()` API to Reader (#654)

bingkh:
*Issue #, if available:*
#654

*Description of changes:*
(Copied from issue description)
Hi, 

**Description**
As a user of ion-js, I would love to be able to get position information from ionReader, so that when using ionReader to read a ion file, I can know which position I'm currently at.

**Application scenario**
Our team uses ion to define a format, e.g. paragraph `P` uses certain format style `S`, and the definition of style `S` is out side of the paragraph itself. see blow:
```
style::{
  name:S,
  font:Arial,
  .....
}
....
....
paragraph::{
  name:A,
  style:S,
  .....
}
```
I'm working on a vs code extension: ion-style-peek, so that in the above ion file, when users are viewing paragraph `A` and call `go to definition` on `S`, the editor will jump to style `S`.  Similar implementation can be found: https://github.com/pranaygp/vscode-css-peek

To implement the above, we need to get the position when parsing style in the ion, hence the requesting issue. 

We can get position information from the below two paths:
```
Reader -> BinaryReader -> ParserBinaryRaw -> StringSpan
Reader -> TextReader -> ParserTextRaw -> BinarySpan
```

Thanks!

Ethan

### DIFF
--- a/src/IonBinaryReader.ts
+++ b/src/IonBinaryReader.ts
@@ -103,6 +103,10 @@ export class BinaryReader implements Reader {
     this._raw_type = BOC;
   }
 
+  position(): number {
+    return this._parser.source().position();
+  }
+
   next(): IonType | null {
     this._annotations = null;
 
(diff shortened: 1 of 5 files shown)


bingkh:
Close due to pretty check failure.

PR #657, opened 2020-12-02
Adding `postion()` API to Reader (#654)

bingkh:
*Issue #, if available:*
#654

*Description of changes:*
(Copied from issue description)
Hi, 

**Description**
As a user of ion-js, I would love to be able to get position information from ionReader, so that when using ionReader to read a ion file, I can know which position I'm currently at.

**Application scenario**
Our team uses ion to define a format, e.g. paragraph `P` uses certain format style `S`, and the definition of style `S` is out side of the paragraph itself. see blow:
```
style::{
  name:S,
  font:Arial,
  .....
}
....
....
paragraph::{
  name:A,
  style:S,
  .....
}
```
I'm working on a vs code extension: ion-style-peek, so that in the above ion file, when users are viewing paragraph `A` and call `go to definition` on `S`, the editor will jump to style `S`.  Similar implementation can be found: https://github.com/pranaygp/vscode-css-peek

To implement the above, we need to get the position when parsing style in the ion, hence the requesting issue. 

We can get position information from the below two paths:
```
Reader -> BinaryReader -> ParserBinaryRaw -> StringSpan
Reader -> TextReader -> ParserTextRaw -> BinarySpan
```

Thanks!

Ethan


src/IonParserBinaryRaw.ts
```diff
@@ -317,6 +317,10 @@ export class ParserBinaryRaw {
     return Decimal._fromBigIntCoefficient(isNegative, coefficient, exponent);
   }
 
+  source(): BinarySpan {
```
zslayton, line 320: I think we should mark the `source()` methods (here and in `IonParserTextRaw`) as `private`. The `BinarySpan` and `StringSpan` classes need to be reviewed before we commit to them in the public API.

src/IonReader.ts
```diff
@@ -41,6 +41,17 @@ export type ReaderScalarValue =
...
+  /**
+   * Return the position of the current reader.
+   * The position refers to the character distance between the char where the reader
+   * stared (e.g. the first character of the file), and the current character the reader
+   * is reading.
+   *
+   * @returns a [[number]] type presenting the position of the character the reading is
+   * currently reading.
+   */
+  position(): number | null;
```
zslayton, line 53: A few thoughts here:
* What does it mean for a Reader's `position` to be `null`?
* The `number` returned by `position()` needs to make sense for both text and binary Readers. The comment describes only the meaning for a text reader.
* Unfortunately, the definition of a "character" [is ambiguous in Unicode terms](https://stackoverflow.com/a/27331885/109549). It could refer to a code point, code unit, glyph or grapheme cluster (among other possibilities). I think our best option here is to refer to code units:
  
  > A **code unit** is the unit of storage of a part of an encoded code point. In UTF-8 this means 8-bits, in UTF-16 this means 16-bits. A single code unit may represent a full code point, or part of a code point. For example, the snowman glyph (☃) is a single code point but 3 UTF-8 code units, and 1 UTF-16 code unit.

zslayton (changes requested):
Hi, thanks for the PR! Some minor questions/comments below.

src/IonParserBinaryRaw.ts
bingkh: Hi @zslayton , thanks so much for taking time to review the PR!

But question: if we mark `source()` private, how can we call this function in `BinaryReader`/`TextReader` and get position there?
Or do you suggest in `parserBinaryRaw`, we do something like:
```suggestion
  getPosition(): number {
    return this._in.postion();
  }
```
and call `this._parser.getPosition()` in `BinaryReader`/`TextReader` ?

src/IonReader.ts
bingkh: Thanks for the input! 
- for default value, I took a second look, and actually once the `TextReader`/`BinaryReader` is initialized, the `position` will be initialized as 0 in `StringSpan`/`BinarySpan`. So I guess the `null` actually doesn't make sense.
- Comment updated, how about the below:
```suggestion
  /**
   * Return the position of the current reader.
   * The position refers to the distance between the code units where the reader
   * stared (e.g. the first code unit of the file), and the current code unit the 
   * reader is reading.
   * 
   * A code unit is the unit of storage of a part of an encoded code point.
   * Ref: https://stackoverflow.com/a/27331885/109549
   *
   * @returns a [[number]] type presenting the position of the code unit the reader is
   * currently reading.
   */
  position(): number;
```

zslayton: The doc comment is an improvement, but we still need a couple of tweaks.

* Because the `Reader` interface is used for both text and binary, we need to explain what `position()` means when the `Reader` is a binary reader. Namely, that it will be the byte offset from the start of the input rather than the code unit offset.
* Because code units are different in different unicode encodings (UTF8 code units are 1-4 bytes, UTF16 are 2 bytes), we should specify that this method returns a number UTF-16 code units, the encoding used by JavaScript strings.
* Even though it's a good explanation, I'd like to avoid using a StackOverflow link in the public documentation. We can point to [JavaScript's String length documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/length#Description), however. We could also refer users to [chapter 2.5 of the Unicode Standard, "Encoding Forms"](https://www.unicode.org/versions/Unicode13.0.0/ch02.pdf), but that's a bit dense for casual reference.
* We should warn users that `Reader`s cannot safely skip to a given position in the stream and begin reading as they may be skipping over system values like symbol definitions.


What do you think of:
```typescript
/**
   * Returns the Reader's offset from the beginning of its input.
   * 
   * For binary Readers, the return value is the number of bytes that have
   * been processed.
   * 
   * For text Readers, the return value is the number of UTF-16 code units
   * that have been processed, regardless of the input's original encoding.
   * For more on JavaScript's in-memory representation of text, see:
   * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/length#Description
   * 
   * Note that a Reader cannot safely skip to a given position in input without
   * processing the stream leading up to that position. This is because there are
   * mid-stream system level values that must be processed to guarantee that the 
   * Reader is in a valid state. It is safe, however, to start at the beginning of a data
   * source and call next() until you reach the desired position, as the reader
   * will still have the opportunity to process system-level values along the way.)
   *
   * @returns the [[number]] of bytes or UTF-16 code units that the reader has processed.
  position(): number;
```
?

bingkh: Thanks for the detailed reply! This looks a lot better! 
I didn't think that much when I started out, and referring to StackOverflow was really a miss 🤦🏻‍♂️
I will update the commit with the latest comment.

src/IonParserBinaryRaw.ts
bingkh: Cool, I will keep the first revision

zslayton:
Ok, looking good! 

One other thing that I should've mentioned before (sorry! 😞): we need a couple of unit tests to prove that this works like we think it does.

Could you add some tests to [`test/BinaryReader.ts`](https://github.com/amzn/ion-js/blob/master/test/IonBinaryReader.ts#L28) and [`test/IonTextReader.ts`](https://github.com/amzn/ion-js/blob/master/test/IonTextReader.ts#L233) demonstrating the expected behavior? Some ideas:

* Calling `position()` on new text readers returns `0`
* Calling `position()` on new binary readers returns `0` (or `4` -- they might eagerly consume the Ion Version Marker?)
* Both formats: calling `position()` after calling `next()` returns the expected value for each of a few values in a stream
* A text stream with some multi-byte and multi-code-unit characters returns the expected position in UTF-16 code units.

bingkh:
Tests added!
No worry, I should've added the tests from the very beginning, tests are also a good demonstration of the reader APIs.
Please let me know if there needs to be any update!

test/IonTextReader.ts
```diff
@@ -249,4 +249,65 @@ class IonTextReaderTests {
             IonTextReaderTests.ivmTest(textReader, unsupportedIVM[i], false, 0, ["taco"]);
         }
     }
+
+    @test "get position from reader" () {
+        const ionToRead = "ann1::{ key: valX } ann2::{ keyX: ['v1', 'v2'], keyY: ['v3', 'v4'] }";
+        const ionTextReader = ion.makeReader(ionToRead);
+        
+        // vertical bar '|' indicate the position
```
zslayton, line 257:
```suggestion
        // In the comments below, the vertical bar '|' indicates the current position of the cursor at each step.
```

test/IonBinaryReader.ts
```diff
@@ -26,4 +26,51 @@ describe('Binary Reader', () => {
         reader.next();
         assert.deepEqual(reader.timestampValue(), timestamp);
     });
+    it('test position', () => {
+        const ionBinary: Uint8Array = ion.dumpBinary(null, 7, -17, "Hello", [1, 2, 3]);
```
zslayton, line 30:
```suggestion
    it('test position', () => {
        // In the comments below, the vertical bar '|' indicates the current position of the cursor at each step.
        const ionBinary: Uint8Array = ion.dumpBinary(null, 7, -17, "Hello", [1, 2, 3]);
```

test/IonTextReader.ts
```diff
@@ -249,4 +249,65 @@ class IonTextReaderTests {
...
+    }
+
+    @test "get position from multi-byte" () {
+
+        // const s0 = 's'; // \u0073
+        // const s1 = '好'; // \u597d
+        // const s2 = 'の'; // \u306e
+        // const s3 = 'ç'; // \u00e7
+        // const s4 = '👩'; // \ud83d \udc69
+        // const s5 = '👩🏽'; // \ud83d \udc69 \ud83c \udffd
```
zslayton, line 304:
```suggestion
        // 's': \u0073
        // '好': \u597d
        // 'の': \u306e
        // 'ç': \u00e7
        // '👩': \ud83d \udc69
        // '👩🏽': \ud83d \udc69 \ud83c \udffd
```

test/IonBinaryReader.ts
```diff
@@ -26,4 +26,51 @@ describe('Binary Reader', () => {
...
+        assert.equal(binaryReader.position(), 19);
+        binaryReader.next();
+        // at 3
+        // [0xE0, 0x1, 0x0, 0xEA, 0xF, 0x21, 0x7, 0x31, 0x11, 0x85, 0x48, 0x65, 0x6C, 0x6C, 0x6F, 0xB6, 0x21, 0x1, 0x21, 0x2, 0x21, 0x3|]
+        assert.equal(binaryReader.position(), 21);
+        binaryReader.stepOut();
+        // out of stream
+        // [0xE0, 0x1, 0x0, 0xEA, 0xF, 0x21, 0x7, 0x31, 0x11, 0x85, 0x48, 0x65, 0x6C, 0x6C, 0x6F, 0xB6, 0x21, 0x1, 0x21, 0x2, 0x21, 0x3]|
+        assert.equal(binaryReader.position(), 22);
+
```
zslayton, line 74:
```suggestion
```

zslayton (commented):
Looking good! One last round of small cleanups.

bingkh:
Picked all suggestions! 👍🏼

zslayton (approved):
Looks good, thanks for contributing!

bingkh:
Hi team, can we know when this change can be released?

Thanks!

Issue #618, opened 2020-06-15
Replace pure JS utf8 encoder/decoder with standard TextEncoder/Decoder


zslayton:
`decodeUtf8` and `encodeUtf8` are *much* slower than the native implementations offered by the now-ubiquitous [`TextEncoder`](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder) and [`TextDecoder`](https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder) classes. We should migrate if our supported runtimes list will allow it.

See [this thread](https://github.com/amzn/ion-js/issues/616#issuecomment-644198579) for details.

wesboyt:
I made the decision to use an unlicensed version of utf8decode because versions of node and browsers that we officially support/ed? did not include textdecoder.
For example on ubuntu 18.04 the latest version of node is 8x and does not include textdecoder. (Even though more recent distributions of 8 do so). The perf of the decode I yoinked is pretty terrible due to repeatedly concatenating strings.

Definetely needs to be replaced.

desaikd:
According to an investigation I did on the `TextDecoder.decode` vs `makeReader` performance when passing `Uint8Array` as input to these methods,

I found out that after 512 bytes(threshold) of `Uint8Array` data `TextDecoder.decode` works faster than `makeReader`. but for data less than 512 bytes `makeReader` works faster. 

```
For more than 512 bytes of Uint8Array of data,

makeReader:
0.000015406160113891825
Binary Reader Large x 64,909 ops/sec ±4.30% (72 runs sampled)

TextDecoder.decode:
0.00001150648431228379
Binary Reader Large x 86,908 ops/sec ±2.48% (80 runs sampled)
```
From above it's clear that `TextDecoder.decode` works faster for more than 512 bytes of data.

```
For 512 or less than 512 bytes of Uint8Array of data,

makeReader:
0.000010301126157451875
Binary Reader Large x 97,077 ops/sec ±2.06% (81 runs sampled)

TextDecoder.decode:
0.0000114280178005657
Binary Reader Large x 87,504 ops/sec ±2.76% (74 runs sampled)
```

Here, taking into account the threshold of 512 bytes . It seems more reasonable to use the current implementation of `makeReader` for 512 or less than 512 bytes and `TextDecoder` method for more than 512 bytes. This gives us best of both worlds and improves performance when there is more than 512 bytes of data as input.

[...]

PR #662, opened 2020-12-03
Performance improvement in unicode decode 

desaikd:
*Issue*: #618 

*Description of changes:*
This PR contains changes for improving performance of Unicode decode method.

*Changes:*

- Changed the `decodeUtf8` method to use native JS `TextDecoder` for `Uint8Array` of greater than 512 bytes. The threshold is set according to the findings from #618 using [Benchmark.js](https://benchmarkjs.com). 
- Also as not every platform would have runtime feature `TextDecoder`, added a check to verify if global variable `TextDecoder` exist or not.

### DIFF
--- a/src/IonUnicode.ts
+++ b/src/IonUnicode.ts
@@ -13,6 +13,11 @@
  * permissions and limitations under the License.
  */
 
+import {TextDecoder} from "util";
+
+// verify if TextDecoder exists globally or not
+let textDecoder = TextDecoder ? new TextDecoder("utf8", {fatal: true}) : null;
+
 /**
  * @file Constants and helper methods for Unicode.
  * @see https://amzn.github.io/ion-docs/stringclob.html
@@ -60,6 +65,10 @@ export function encodeUtf8(s: string): Uint8Array {
 }
 
 export function decodeUtf8(bytes: Uint8Array): string {
+  // for bytes > 512 use TextDecoder method - decode()
+  if(bytes.length > 512 && textDecoder != null) {
+    return textDecoder.decode(bytes);
+  }
   let i = 0,
     s = "",
     c;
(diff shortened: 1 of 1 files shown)


src/IonUnicode.ts
```diff
@@ -13,6 +13,11 @@
  * permissions and limitations under the License.
  */
 
+import { TextDecoder } from "util";
```
zslayton, line 16: Why is this being imported from `util`? The [current version of that file](https://github.com/amzn/ion-js/blob/be789e3f99f0c2ebd5365a634000eea62a827ddf/src/util.ts) doesn't seem to reference it. If the runtime supports it, won't it be available globally without an import?

src/IonUnicode.ts
```diff
@@ -13,6 +13,11 @@
...
+
+// verify if TextDecoder exists globally or not
```
zslayton, line 18: Nit: this comment is entirely accurate, but I think it would be more helpful to explain the higher-level purpose of the code it accompanies.

How about:
```suggestion
// Check whether this runtime supports the `TextDecoder` feature
```
?

src/IonUnicode.ts
```diff
@@ -60,6 +65,10 @@ export function encodeUtf8(s: string): Uint8Array {
 }
 
 export function decodeUtf8(bytes: Uint8Array): string {
+  // for bytes > 512 use TextDecoder method - decode()
+  if (bytes.length > 512 && textDecoder != null) {
+    return textDecoder.decode(bytes);
+  }
```
zslayton, line 71: Please create a named constant for `512`. `JS_DECODER_MAX_BYTES`?

src/IonUnicode.ts
desaikd: It is getting imported from @types/node/util.d.ts

zslayton (approved):

Issue #621, opened 2020-06-26
[Binary] Writing decimal values can produce invalid Ion


zslayton:
Using this data:
```js
'com.example.organization.model.data.ClassName2@1.0'::{
  values: [
    'com.example.organization.model.data.ClassName1@1.0'::{
      field_name_1: 'com.example.organization.model.types.ClassName1@1.0'::{
        field_name_2: 'com.example.organization.model.types.ClassName2@1.0'::{
          field_name_3: 9999999999.00,
          field_name_4: ABC
        }
      },
      field_name_5: AB
    },
    'com.example.organization.model.data.ClassName1@1.0'::{
      field_name_1: 'com.example.organization.model.types.ClassName1@1.0'::{
        field_name_2: 'com.example.organization.model.types.ClassName2@1.0'::{
          field_name_3: 9999999999.00,
          field_name_4: DEF
        }
      },
      field_name_5: CD
    }
  ]
}
```
the following code produces an unreadable binary Ion file:
```javascript
let ionText = fs.readFileSync("exampleDataFromAbove.ion");
let ionValue = ion.load(ionText);
let binaryIonBytes = ion.dumpBinary(ionValue);
fs.writeFileSync("outputFile.10n", binaryIonBytes);
````

Changing both instances of `field_name_3: 9999999999.00` to `field_name_3: 9.99` fixes it, so the issue may be tied to the size of the value.

desaikd:
The issue was with the generation of invalid ion while using `IonBinaryWriter` to dump binary data. It seems like it was not considering the annotation length(according to spec) along with the contained value to write length field for a container node. The fix has been merged with #665.

PR #665, opened 2020-12-21
Regarding issue with calculation of annotation length for IonBinaryWriter

desaikd:
*Issue:* #621 

*Description of changes:*
This PR shows the issue with the generation of invalid ion while using `IonBinaryWriter` to dump binary data. It seems like it is not considering the annotation length(according to [spec](http://amzn.github.io/ion-docs/docs/binary.html#annotations)) along with the contained value to write length field for a [container node](https://github.com/amzn/ion-js/blob/3c7a67c3b04655b33df32227ca554ea874d2fcc6/src/IonBinaryWriter.ts#L706).

*Changes:*
1. added a unittest to demonstrate the issue with invalid ion due to wrong annotation length calculation. (ae37bc5)
2. add the changes to solve the issue considering the [spec](http://amzn.github.io/ion-docs/docs/binary.html#annotations) for calculating annotation length.


test/dump.ts
```diff
@@ -74,4 +74,10 @@ describe('dump*()', () => {
         writer.close();
         assert.deepEqual(ionBinary, writer.getBytes());
     });
+    it('Struct with annotated field large enough to require a VarUInt length', () => {
```
zslayton, line 77: Could you also add the example input from issue #621 as a separate unit test? Add a comment with the GitHub issue link too, please.

test/dump.ts
```diff
@@ -74,4 +74,40 @@ describe('dump*()', () => {
...
+            "        }\n" +
+            "      },\n" +
+            "      field_name_5: CD\n" +
+            "    }\n" +
+            "  ]\n" +
+            "}";
+        let ionValue = ion.load(ionText);
+        let binaryIonBytes = ion.dumpBinary(ionValue);
+        let ionBinaryValue = ion.load(binaryIonBytes);
+        assert.deepEqual(ionValue, ionBinaryValue);
```
zslayton, line 111: A word of caution: `assert.deepEqual` works for these unit tests, but it's going to be overly precise in other situations where we're comparing Ion values. There are lots of different Javascript in-memory representations for the same Ion data.

zslayton (approved):

Issue #664, opened 2020-12-11
BinaryReader throws Skipped over end of source error when traversing using next()


desaikd:
When trying to read a large binary file with series of top-level structs, binaryReader throws `Skipped over end of source` error while traversing using only `next()`. This error disappears when using `stepIn()` along with `next()` while traversing through the file. It is unclear as to why this happens although it's probably the `limit` value specified in `IonSpan` that causes this error probably due to an overflow, but it needs more investigation.

desaikd:
As specified in PR #666 I have investigated on this issue and found out that the issue was with the way encode and decode was working for storing type information of container when stepping in and out. 

The issue was occurring for `len` values that were greater than 27 bits.  The encode method uses 5 bit left shift and the decode uses 5 bit right shift, to store both `len` and `type` information in the same value. As a result of this the information was lost for the first 5 bits and leads to an incorrect value when decoding the stack value.  

Hence in the PR, I have suggested a change for storing the stack value using an Object with variables `len` and `type`. This will help storing `len` and `type information separately and will give correct values.

PR #666, opened 2020-12-22
Change the way container type information is stored in stack

desaikd:
*Issue #664 :*

*Description of changes:*
This PR is to resolve the error `Skipped over end of source` when traversing input Ion data with only doing `next()` operation.

*Cause of Issue:*
The reason that this error occurred was because `IonParserBinaryRaw` was using a value(`ts`) created by storing both `len` and `type` field of container type into one. This value is then pushed inside a stack(`_ts`) which is used to store information of a container when steping in and the same value would be popped on the step out side to get back the container information. 

 Here, the `len` value that is encoded by `encode_type_stack` if exceeds 27 bits, it won't be able to get back to its correct `len` value when decoded using `decode_type_stack_len`. This gives an incorrect `len` value and eventually it sets the `limit` value of `IonSpan` to be incorrect and hence throws the error `Skipped over end of source`.

*Solution to the Issue:*
As a solution to this issue I have used a `stackPointer` Object which has `len` and `type` variables. This stackPointer would be added to the stack `_ts` and can be popped out to get back the correct `len` and `type` information.

*Test:*
Tested this using a large enough file that could give a value greater than 27 bits for `len`.


src/IonParserBinaryRaw.ts
```diff
@@ -131,6 +115,17 @@ const ivm_image_1 = IVM.binary[1];
 const ivm_image_2 = IVM.binary[2];
 const ivm_image_3 = IVM.binary[3];
 
+// stack pointer for _ts to store length and type information of container
+class stackPointer {
```
zslayton, line 119: To me, the phrase "stack pointer" means the address of the top of a stack. Since we'll have an instance of this type for each container we step into, I suggest `EncodingContainer`, `EncodingLevel` or similar. Note that in TypeScript's style conventions, [class names always begin with a capital letter](https://basarat.gitbook.io/typescript/styleguide#class).

src/IonParserBinaryRaw.ts
```diff
@@ -131,6 +115,17 @@ const ivm_image_1 = IVM.binary[1];
...
+  len: number;
+  type: number;
+
+  constructor(len: number, type: number) {
```
zslayton, line 123: Subjective: when ordering parameters, I prefer to put the details last. To me, the `type` of a container is more interesting than the number of encoding bytes it occupies, so I'd reverse these. Feel free to ignore this if you disagree.

src/IonParserBinaryRaw.ts
```diff
@@ -131,6 +115,17 @@ const ivm_image_1 = IVM.binary[1];
 const ivm_image_2 = IVM.binary[2];
 const ivm_image_3 = IVM.binary[3];
 
+// stack pointer for _ts to store length and type information of container
+class stackPointer {
+  len: number;
```
zslayton, line 120: I know the original variable names were `len` and `type`, but since this is new code I'd like to spell `len` out in full.

zslayton (commented):
Some superficial comments below.

zslayton (approved):

Issue #663, opened 2020-12-08
Store smaller integers as numbers instead of BigInts


desaikd:
*Description:*
Using `number` for smaller integers instead of `JSBI.BigInt` could be  a helpful useful performance change (which can potentially be providing 20% speed up)

*Performance change:*
Instead of using [`JSBI.BigInt`](https://github.com/GoogleChromeLabs/jsbi#usage) for all the Signed Integers we can use native `number` datatype for values less than [5 bytes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isSafeInteger). This will speed up the performance because it saves the time of converting `JSBI.BIgInt` into `number`.

Here's the result of improvement in performance using [benchmarkjs](https://benchmarkjs.com) for measurement.
```
Using JSBI.BigInt for smaller no of bytes:
10.172569152600001
Binary Reader Large x 0.10 ops/sec ±1.86% (5 runs sampled)

Using number for smaller no of bytes:
8.447150749799999
Binary Reader Large x 0.12 ops/sec ±1.63% (5 runs sampled)
```
The threshold used here is 5 bytes. The results show that this change could improve the speed up to 20%. but this change would have some requirements to be addressed as described below.

*Requirements for the performance change to be implemented:*

- Add a warning to the already-deprecated `.value()` method explaining that two values with the same Ion type can have two different JS type representations. (e.g. `number` and `jsbi` for Ion’s `int`). This will require people to use `numberValue()`  and `bigIntValue()` instead of `value()`.

- Provide an `equals` method on `dom.Value` API and let people use that instead. (This can also be used for test suites in ion-js).

*Some related issues addressing this concern:*

- #576: regarding the having an `equals` method on `dom.Value`.
- #218: regarding the performance improvement using `number`.

zslayton:
This optimization was merged in PR #673, which includes some nice benchmark results. 👌

PR #669, opened 2021-01-08
Implementation of equals() API for dom.Value

desaikd:
**Issue #663**

**Description of changes:**
This PR implements the `equals` API for `dom.Value`. 

**Terminology:**
- _Strict equivalence_: 
It refers to Ion data model equivalence as defined above and by [Ion Specification](http://amzn.github.io/iondocs/docs/spec.html)

- _Structural or non strict equivalence_: 
It follows the same rules as strict equivalence, except that
   - Annotations are not considered, and
   - Timestamps that represent the same instant in time are always considered equivalent.

**equals():**
This method can be used on each `dom.Value` for checking it's equivalence with other dom.Value. It return a boolean value representing whether both values are equal or not.

*Options:*
This method has additional options for equivalence as below:
`epsilon` is used by Float for an equality with given epsilon precision. (Default: null)
`strict` specifies whether equality should be strict or non strict. (Default: true)

**Tasks:**

1. Created an `equals()` method in dom.Value for implementation.
2. Created an `ionEquals()` method that cam be used as a helper method inside equals() for each dom.Value specific equivalence logic.
3. Implemented `ionEquals()` for each `dom.Value` type.

**Test:**
Created equivalence unit tests for each `dom.Value` Ion data types.


src/dom/Value.ts
```diff
@@ -309,6 +343,99 @@ export function Value<Clazz extends Constructor>(
       this._unsupportedOperation("deleteField");
     }
 
+    /**
+     * Helper method to be implemented by each dom.Value with individual equivalence logic.
+     *
```
zslayton, line 348: I think this text is borrowed directly from the Ion spec. If so, I suggest adding a link to that section of the spec instead of including the text inline here. That way if the spec is updated to clarify something, we won't have to update it here as well.

src/dom/Value.ts
```diff
@@ -309,6 +343,99 @@ export function Value<Clazz extends Constructor>(
...
+        ignoreAnnotations: false,
+        ignoreTimestampPrecision: false,
+        onlyCompareIon: true,
+      }
+    ): boolean {
+      this._unsupportedOperation("ionEquals");
+    }
+
+    equals(
+      expectedValue: any,
```
zslayton, line 404: I associate the name `expectedValue` with unit testing, where we'd throw an exception if `this` wasn't equal to it. In an equality check, we don't have any expectations. How about `otherValue` or `other`?

src/dom/Value.ts
```diff
@@ -309,6 +343,99 @@ export function Value<Clazz extends Constructor>(
...
+        ignoreTimestampPrecision?: boolean;
+        onlyCompareIon?: boolean;
+      } = {
+        epsilon: null,
+        ignoreAnnotations: false,
+        ignoreTimestampPrecision: false,
+        onlyCompareIon: true,
+      }
+    ): boolean {
+      this._unsupportedOperation("ionEquals");
```
zslayton, line 400: Could you add the new `_` prefix to the operation name?

src/dom/Value.ts
```diff
@@ -309,6 +343,99 @@ export function Value<Clazz extends Constructor>(
       this._unsupportedOperation("deleteField");
     }
 
+    /**
+     * Helper method to be implemented by each dom.Value with individual equivalence logic.
```
zslayton, line 347: Nit: to be implemented by each subclass of `dom.Value`

src/dom/Value.ts
```diff
@@ -309,6 +343,99 @@ export function Value<Clazz extends Constructor>(
...
+        epsilon: null,
+        ignoreAnnotations: false,
+        ignoreTimestampPrecision: false,
+        onlyCompareIon: true,
+      }
+    ): boolean {
+      this._unsupportedOperation("ionEquals");
+    }
+
+    equals(
```
zslayton, line 403: This method could use a doc comment.

src/dom/Value.ts
```diff
@@ -309,6 +343,99 @@ export function Value<Clazz extends Constructor>(
...
+        if (!(expectedValue instanceof Value)) {
+          return false;
+        }
+        let actualAnnotations = this.getAnnotations();
+        let expectedAnnotations = (expectedValue as Value).getAnnotations();
+        if (actualAnnotations.length !== expectedAnnotations.length) {
+          return false;
+        }
+        for (let i = 0; i < actualAnnotations.length; i++) {
+          if (actualAnnotations[i].localeCompare(expectedAnnotations[i])) {
```
zslayton, line 431: * What's the motivation to use `localeCompare` here instead of `===`?
* While this logic is correct, I had to look up the return value description for `localeCompare` to understand why. I suggest adding a ` !== 0` at the end of this to make your expectations clear.

src/dom/Decimal.ts
```diff
@@ -56,4 +57,44 @@ export class Decimal extends Value(
...
+      !(
+        (options.onlyCompareIon && expectedValue instanceof Decimal) ||
+        (!options.onlyCompareIon &&
+          (expectedValue instanceof ion.Decimal ||
+            expectedValue instanceof Number ||
+            typeof expectedValue === "number"))
+      )
+    ) {
+      return false;
+    }
```
zslayton, line 86: These code paths are (necessarily) pretty complicated. I think it would be nice to break these conditionals into separate cases that readers can digest independently and to which you can add separate comments.

```typescript
let isSupportedType = false;
let valueToCompare = null;
if (options.compareOnlyIon) {
  // `compareOnlyIon` requires that the provided value be an ion.dom.Decimal instance.
  if (expectedValue instanceof Decimal) {
    isSupportedType = true;
    valueToCompare = expectedValue.decimalValue();
  }
} else {
  // We will consider other Decimal-ish types
  if (expectedValue instanceof ion.Decimal) {
    // expectedValue is a non-DOM Decimal
    isSupportedType = true;
    valueToCompare = expectedValue;
  } else if (expectedValue instanceof Number || typeof expectedValue === "number") {
    isSupportedType = true;
    // calling numberValue() on ion.Decimal is lossy and could result in imprecise comparisons
    // hence converting number to ion.Decimal for comparison even though it maybe expensive
    valueToCompare = new ion.Decimal(expectedValue.toString());
  }
}

if (!isSupportedType) {
  return false;
}

return this.decimalValue().equals(valueToCompare);
```

zslayton (approved):
👍

Issue #559, opened 2020-02-18
Deprecate Reader.byteValue() in favor of a new Reader.uInt8ArrayValue() method


pbcornell:
src/dom/Value.ts provides a clearer `uInt8ArrayValue(): Uint8Array | null` method; consider updating the `Reader` interface to follow suit.

desaikd:
Remove the deprecated method for `v5.0.0` release.

desaikd:
Fixed with #750

PR #672, opened 2021-01-27
Create  uInt8ArrayValue method for reader and deprecate Reader.byteValue

desaikd:
*Issue #559*

*Description of changes:*
This Pr works on the above mentioned issue for deprecating use of `Reader.byteValue()` and instead use a new method `Reader.uInt8ArrayValue()`.

*Tasks:*
- create a new method called `uInt8ArrayValue.
- deprecate use of `byteValue` from tests

### DIFF
--- a/src/IonReader.ts
+++ b/src/IonReader.ts
@@ -133,6 +133,7 @@ export interface Reader {
    * @return `null` if the current Ion value [[isNull]].
    *
    * @throw Error when the reader is not positioned on a `clob` or `blob` typed value.
+   * @deprecated Since version 4.2.  Use the `uInt8ArrayValue` method instead
    */
   byteValue(): Uint8Array | null;
 
--- a/src/IonReader.ts
+++ b/src/IonReader.ts
@@ -137,6 +137,16 @@ export interface Reader {
   byteValue(): Uint8Array | null;
 
   /**
+   * Returns the current value as a `Uint8Array`.  This is only valid if `type() == IonTypes.CLOB`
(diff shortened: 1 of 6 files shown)


src/IonReader.ts
```diff
@@ -136,6 +136,16 @@ export interface Reader {
    */
```
zslayton, line 136: Please add a `@deprecated` warning to the docs pointing users to the new `uInt8ArrayValue` method.

src/IonTextReader.ts
```diff
@@ -318,6 +318,25 @@ export class TextReader implements Reader {
     throw new Error("Current value is not a blob or clob.");
   }
 
+  uInt8ArrayValue(): Uint8Array | null {
+    this.load_raw();
```
zslayton, line 322: Rather than maintaining two copies of this function body, could you modify `byteValue()` to call `uInt8ArrayValue()` as its implementation?

src/IonReader.ts
```diff
@@ -133,6 +133,7 @@ export interface Reader {
    * @return `null` if the current Ion value [[isNull]].
    *
    * @throw Error when the reader is not positioned on a `clob` or `blob` typed value.
+   * @deprecated Since version 4.2.  Use the `uInt8ArrayValue` method instead
```
zslayton, line 136:
```suggestion
   * @deprecated since version 4.2. Use the `uInt8ArrayValue` method instead.
```

zslayton (approved):
One nit below.

PR #673, opened 2021-01-27
Use number instead of JSBI.BigInt for smaller integers

desaikd:
*Issue #663*

*Description of changes:*
This PR works on a performance change using `number` instead of `JSBI.BigInt` for smaller integers.

*Changes:*
1. Change IonBinaryParser logic to use `number` for integers smaller than 6 bytes. Otherwise, use `JSBI.BigInt`. 
2. Modified dom.Value#Integers to use `JSBI.equal` for comparison. Hence converting both comparison values to `JSBI.BigInt` first for precision.
3. Modified the `IonEvent#valueCompare` for `Integer` Event to convert both comparison values to `JSBI.BigInt` first and then compare.

*Test:*
Verified performance using [benchmarkjs](https://benchmarkjs.com) with large binary ion file as input.

*Test Results:*
| Input file type  | Time(before enhancement) | Time (after enhancement) | % decrease in Time
|----------|:-------------:|------:|-----:|
| Binary - big integers | 12.666 | 12.820 | -1.215 
| Binary - small integers | 10.356 | 1.000 | 90.342

### DIFF
--- a/src/dom/Integer.ts
+++ b/src/dom/Integer.ts
@@ -96,18 +96,10 @@ export class Integer extends Value(Number, IonTypes.INT, _fromJsConstructor) {
   ): boolean {
     let isSupportedType: boolean = false;
     let valueToCompare: any = null;
-    if (other instanceof Integer) {
-      isSupportedType = true;
-      if (this._bigIntValue == null && other._bigIntValue == null) {
-        valueToCompare = other.numberValue();
-      } else {
-        // One of them is a JSBI
-        valueToCompare = other.bigIntValue();
-      }
-    }
 
     if (options.onlyCompareIon) {
       // `compareOnlyIon` requires that the provided value be an ion.dom.Integer instance.
+      isSupportedType = true;
       if (this._bigIntValue == null && other._bigIntValue == null) {
         valueToCompare = other.numberValue();
       } else {
--- a/src/dom/Integer.ts
+++ b/src/dom/Integer.ts
@@ -100,13 +100,13 @@ export class Integer extends Value(Number, IonTypes.INT, _fromJsConstructor) {
       // `compareOnlyIon` requires that the provided value be an ion.dom.Integer instance.
       if (other instanceof Integer) {
         isSupportedType = true;
-        valueToCompare = other.numberValue();
+        valueToCompare = other.bigIntValue();
       }
     } else {
       // We will consider other Integer-ish types
       if (other instanceof global.Number || typeof other === "number") {
         isSupportedType = true;
-        valueToCompare = other.valueOf();
+        valueToCompare = JSBI.BigInt(other.valueOf());
       } else if (other instanceof JSBI) {
         isSupportedType = true;
         valueToCompare = other;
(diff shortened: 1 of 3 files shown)


src/dom/Integer.ts
```diff
@@ -100,13 +100,13 @@ export class Integer extends Value(Number, IonTypes.INT, _fromJsConstructor) {
       // `compareOnlyIon` requires that the provided value be an ion.dom.Integer instance.
       if (other instanceof Integer) {
         isSupportedType = true;
-        valueToCompare = other.numberValue();
+        valueToCompare = other.bigIntValue();
```
zslayton, line 103: I think this means that two `ion.dom.Integer` instances representing small numbers won't be able to compare their `number` values without first converting them to `JSBI`. Is that right?

I think you could side step that extra construction by testing whether `valueToCompare` is a `JSBI` or not right before the final `return` and doing the appropriate comparison.

```typescript
if (other instanceof Integer) {
  isSupportedType = true;
  if (this.bigIntValue == null && other.bigIntValue == null) {
    valueToCompare = other.numberValue();
  } else {
    // One of them is a JSBI
    valueToCompare = other.bigIntValue();
  }
}
```

```typescript
if (other instanceof global.Number || typeof other === "number") {
  isSupportedType = true;
  if (this.bigIntValue == null) {
    valueToCompare = other.valueOf();
  } else {
    valueToCompare = JSBI.BigInt(other.valueOf());
  }
}
```

```typescript
if (valueToCompare instanceof JSBI) {
  return  JSBI.equal(this.bigIntValue(), valueToCompare);
}
return this.numberValue() == valueToCompare;
```

src/dom/Integer.ts
```diff
@@ -96,17 +96,33 @@ export class Integer extends Value(Number, IonTypes.INT, _fromJsConstructor) {
...
+    }
+
     if (options.onlyCompareIon) {
       // `compareOnlyIon` requires that the provided value be an ion.dom.Integer instance.
-      if (other instanceof Integer) {
-        isSupportedType = true;
+      if (this._bigIntValue == null && other._bigIntValue == null) {
+        valueToCompare = other.numberValue();
+      } else {
+        // One of them is a JSBI
```
zslayton: Sorry, I meant for my example code to be an update to the existing code and not an additional block. Some of the logic is now being repeated (unless I'm missing something?).

src/dom/Integer.ts
```diff
@@ -96,17 +96,25 @@ export class Integer extends Value(Number, IonTypes.INT, _fromJsConstructor) {
   ): boolean {
     let isSupportedType: boolean = false;
     let valueToCompare: any = null;
+
     if (options.onlyCompareIon) {
       // `compareOnlyIon` requires that the provided value be an ion.dom.Integer instance.
-      if (other instanceof Integer) {
```
zslayton, line 101: I think this `if` check was still needed to make sure that `isSupportedType = true` is right.

zslayton (approved):
This is a great speedup, nice work!

Issue #675, opened 2021-02-02
Timestamp enhacements 


chafey:
Timestamp currently requires that you specify each component (offset, year, month, day, hour, minute, second).  A common use case is to create a Timestamp for now or based on an existing Date.  It would be nice if there were convenience functions for both.  Here is some code I wrote for now():

https://github.com/chafey/dicom2ion-js/blob/main/src/timestampNow.js

I would be happy to submit a PR if this is this functionality is desired

desaikd:
This looks like a good change to have for `ion-js`. Your change looks good although we would like to use a multi-constructor approach here, something like we have used in [`ion.Decimal`](https://github.com/amzn/ion-js/blob/master/src/IonDecimal.ts#L85) class. Currently the class that takes all the (offset, year, month, day, hour, minute, second) components to create a Timestamp is [`ion.Timestamp`](https://github.com/amzn/ion-js/blob/master/src/IonTimestamp.ts#L73-L80), where you can add your change and submit a PR.

chafey:
Thank you for the quick response, I'll start working on this now.

chafey:
I decided to create a static factory method as the constructor was already getting pretty long and unwieldy - hope thats OK!  If not, I can resubmit with overloaded constructor.  Thanks

desaikd:
PR #686 contains changes for creating `Timestamp` using `Date` object.

PR #677, opened 2021-02-03
Added support for creating Timestamps from Date objects

chafey:
#675 675

Added static factory method to Timestamp that takes a JavaScript Date object as input

### DIFF
--- a/src/IonTimestamp.ts
+++ b/src/IonTimestamp.ts
@@ -191,7 +191,10 @@ export class Timestamp {
    * @param date - valid JavaScript date object
    */
   static fromDate(date: Date): Timestamp {
-    const seconds = new Decimal(date.getSeconds(), (date.getMilliseconds() / 1000))
+    const seconds = new Decimal(
+      date.getSeconds(),
+      date.getMilliseconds() / 1000
+    );
     return new Timestamp(
       date.getTimezoneOffset() / 60,
       date.getFullYear(),
@@ -199,7 +202,8 @@ export class Timestamp {
       date.getDate(),
       date.getHours(),
       date.getMinutes(),
-      seconds)  
+      seconds
+    );
   }
 
   /**
--- a/src/IonTimestamp.ts
+++ b/src/IonTimestamp.ts
@@ -186,6 +186,23 @@ export class Timestamp {
   }
 
   /**
+   * Static factory method for creating a Timestamp from a JavaScript Date object
+   *
+   * @param date - valid JavaScript date object
+   */
+  static fromDate(date: Date): Timestamp {
+    const seconds = new Decimal(date.getSeconds(), (date.getMilliseconds() / 1000))
+    return new Timestamp(
+      date.getTimezoneOffset() / 60,
+      date.getFullYear(),
+      date.getMonth() + 1,
+      date.getDate(),
+      date.getHours(),
+      date.getMinutes(),
+      seconds)  
+  }
+
+  /**
    * Parses a string and returns a corresponding Timestamp object.
    * The provided string must be a text-encoded Timestamp as specified
    * in the Ion specification.
--- a/test/IonTimestamp.ts
+++ b/test/IonTimestamp.ts
@@ -242,4 +242,11 @@ describe("Timestamp", () => {
         assert.deepEqual(ts.getSecondsDecimal(), ion.Decimal.parse('45'));
         assert.deepEqual(ts._getFractionalSeconds(), ion.Decimal.ZERO);
     });
+    it('constructor with Date', () => {
+        const date = new Date(2000, 0, 1, 12, 30, 45);
+        let timestamp1 = new ion.Timestamp(-7 * 60, 2000, 1, 1, 12, 30, 45);
+        let timestamp2 = ion.Timestamp.fromDate(date);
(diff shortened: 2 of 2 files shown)


test/IonTimestamp.ts
```diff
@@ -242,4 +242,11 @@ describe("Timestamp", () => {
         assert.deepEqual(ts.getSecondsDecimal(), ion.Decimal.parse('45'));
         assert.deepEqual(ts._getFractionalSeconds(), ion.Decimal.ZERO);
     });
+    it('constructor with Date', () => {
+        const date = new Date(2000, 0, 1, 12, 30, 45);
+        let timestamp1 = new ion.Timestamp(-7 * 60, 2000, 1, 1, 12, 30, 45);
+        let timestamp2 = ion.Timestamp.fromDate(date);
+        assert.equal(timestamp1!.equals(timestamp2!), timestamp1 === timestamp2);  // data model equivalence
```
zslayton, line 249: This assertion is a bit off, which is causing this test to pass despite the offset discrepancy described above.

If `timestamp1` and `timestamp2` are not equal to each other, then both
```typescript
timestamp1!.equals(timestamp2!)
```
and
```typescript
timestamp1 === timestamp2!
```
will be `false`. That means that `assert.equal` will be satisfied and the test will pass, which I don't believe is the intent.

I think this can be simplified to:
```typescript
assert.isTrue(timestamp1!.equals(timestamp2!));
```

test/IonTimestamp.ts
```diff
@@ -242,4 +242,11 @@ describe("Timestamp", () => {
         assert.deepEqual(ts.getSecondsDecimal(), ion.Decimal.parse('45'));
         assert.deepEqual(ts._getFractionalSeconds(), ion.Decimal.ZERO);
     });
+    it('constructor with Date', () => {
+        const date = new Date(2000, 0, 1, 12, 30, 45);
+        let timestamp1 = new ion.Timestamp(-7 * 60, 2000, 1, 1, 12, 30, 45);
+        let timestamp2 = ion.Timestamp.fromDate(date);
```
zslayton, line 248: The [Ion spec specifies](https://amzn.github.io/ion-docs/docs/spec.html#timestamp) that two timestamps are only equal if (among other things) their offsets are the same:

> Two timestamps are only equivalent if they represent the same instant with **the same offset** and precision.

(Emphasis mine.)

`timestamp1` specifies an offset of UTC-7 while `timestamp2` is built from a source `Date` that doesn't specify an offset. The `Date`'s offset will be that of the host system, which will vary between developer machines and build servers. This means that the two timestamps will only be considered equal when the test is run on a machine with an offset of UTC-7.

`Date`'s constructor doesn't allow you to specify a timezone offset directly (weird!), and one might reasonably be tempted to provide one by passing an ISO8601 timestamp instead of individual parameters. Unfortunately, JavaScript will only adjust the resulting `Date` back to the host system's local time:

image

This means that there's not going to be a reliable way to construct a `Date` with the desired offset and configure `timestamp2` with the same offset independently. I think the best we can do is something like this:

```suggestion
        const date = new Date(2000, 0, 1, 12, 30, 45);
        let timestamp1 = new ion.Timestamp(date.getTimezoneOffset() * -1, 2000, 1, 1, 12, 30, 45);
        let timestamp2 = ion.Timestamp.fromDate(date);
```

I was pleased to learn this morning that TC39, the standards body that governs ECMAScript/JavaScript, [is working on a new datetime standard](https://tc39.es/proposal-temporal/docs/) that looks to be much more sane.  🙂

src/IonTimestamp.ts
```diff
@@ -185,6 +185,27 @@ export class Timestamp {
...
+   *
+   * @param date - valid JavaScript date object
+   */
+  static fromDate(date: Date): Timestamp {
+    const seconds = new Decimal(
+      date.getSeconds(),
+      date.getMilliseconds() / 1000
+    );
+    return new Timestamp(
+      date.getTimezoneOffset() / 60,
```
zslayton, line 199: I'm not sure this needs to be divided by 60. Could you double check the units here?

src/IonTimestamp.ts
```diff
@@ -185,6 +185,27 @@ export class Timestamp {
     );
   }
 
+  /**
+   * Static factory method for creating a Timestamp from a JavaScript Date object
+   *
+   * @param date - valid JavaScript date object
+   */
+  static fromDate(date: Date): Timestamp {
+    const seconds = new Decimal(
```
zslayton, line 194: Capturing the `Date`'s seconds as a decimal to preserve millisecond precision is the right idea here, but `Decimal`'s constructor is expecting an coefficient/exponent pair (scientific notation components) rather than an integer/fraction pair. You can find the documentation [here](https://github.com/amzn/ion-js/blob/3554f29b438fbc7021cf3760a46775257514be1d/src/IonDecimal.ts#L21-L37).

zslayton (changes requested):

zslayton:
Hi @chafey, are you still interested in working on this? It'd be a really helpful change.

chafey:
I am but super busy - my feelings won't be hurt if someone else does it

desaikd:
With #686 (which contains `Timestamp` changes for using `Date` object) being merged this PR can be closed.

PR #679, opened 2021-02-10
Add support for duplicate fieldnames in dom.Value#Struct

desaikd:
*Description of changes:*
This PR is to support duplicate fields in a Struct according to the [Ion Spec](https://amzn.github.io/ion-docs/docs/spec.html#struct).

*Tasks:*

1.  Added changes in the `dom.Value#Struct` for supporting duplicate fields. (All the `Struct` field will now be stored in `(fieldname, values)` -> `[string, Value[]]` form)
2.  `dom.Value#Struct` constructor would accept both `Iterable<string, Value>` as well as `Iterable<string, Value[]>` for creating `Struct`.
2.  Added `getAll` and `allFields` methods for getting all values of Struct fields.
3.  `toString` method returns with all the values of Struct fields.
2.  Changes in test suite for duplicate fields.
3.  add tests for duplicate fields.

### DIFF
--- a/src/dom/Null.ts
+++ b/src/dom/Null.ts
@@ -120,7 +120,7 @@ export class Null extends Value(Object, IonTypes.NULL, FromJsConstructor.NONE) {
     this._unsupportedOperationOrNullDereference("fieldNames");
   }
 
-  fields(): [string, Value][] {
+  fields(): [string, Value[]][] {
     this._unsupportedOperationOrNullDereference("fields");
   }
 
--- a/src/dom/Null.ts
+++ b/src/dom/Null.ts
@@ -120,7 +120,7 @@ export class Null extends Value(Object, IonTypes.NULL, FromJsConstructor.NONE) {
     this._unsupportedOperationOrNullDereference("fieldNames");
   }
 
-  fields(): [string, Value[]][] {
+  fields(): [string, Value][] {
     this._unsupportedOperationOrNullDereference("fields");
   }
 
--- a/src/dom/Struct.ts
+++ b/src/dom/Struct.ts
@@ -48,9 +48,8 @@ export class Struct extends Value(
     super();
     // Struct fields will always be stored as (fieldName, fieldValues) => (string, Value[])
     for (const [fieldName, fieldValue] of fields) {
-      this._fields[fieldName] = Array.isArray(fieldValue)
-        ? fieldValue
-        : [fieldValue];
+      this._fields[fieldName] =
+        fieldValue instanceof Value ? [fieldValue] : fieldValue;
     }
     this._setAnnotations(annotations);
 
--- a/src/dom/Struct.ts
+++ b/src/dom/Struct.ts
@@ -41,7 +41,7 @@ export class Struct extends Value(
    * @param fields        An iterator of field name/value pairs to represent as a struct.
    * @param annotations   An optional array of strings to associate with this null value.
    */
-  constructor(fields: Iterable<[string, Value[]]>, annotations: string[] = []) {
+  constructor(fields: Iterable<[string, Value]> | Iterable<[string, Value[]]>, annotations: string[] = []) {
     super();
     for (const [fieldName, fieldValue] of fields) {
       this._fields[fieldName] = fieldValue;
@@ -98,11 +98,35 @@ export class Struct extends Value(
     return child[child.length - 1]!.get(...pathTail);
   }
 
+  getAll(...pathElements: PathElement[]): Value[] | null {
+    if (pathElements.length === 0) {
+      throw new Error("Value#get requires at least one parameter.");
+    }
+    const [pathHead, ...pathTail] = pathElements;
+    if (typeof pathHead !== "string") {
+      throw new Error(`Cannot index into a struct with a ${typeof pathHead}.`);
+    }
+    const child: Value[] | undefined = this._fields[pathHead];
+    if (child === undefined) {
+      return null;
+    }
+    if (pathTail.length === 0) {
+      return child!;
+    }
+    let values: Value[] | undefined = [];
+    child.forEach(value => values!.push(...value.getAll(...pathTail)!));
+    return values!;
+  }
+
   fieldNames(): string[] {
     return Object.keys(this._fields);
   }
 
-  fields(): [string, Value[]][] {
+  allFields(): [string, Value[]][] {
+    return Object.entries(this._fields);
+  }
+
+  fields(): [string, Value][] {
     return Object.entries(this._fields);
   }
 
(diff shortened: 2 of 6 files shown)


src/dom/Null.ts
```diff
@@ -120,7 +120,7 @@ export class Null extends Value(Object, IonTypes.NULL, FromJsConstructor.NONE) {
     this._unsupportedOperationOrNullDereference("fieldNames");
   }
 
-  fields(): [string, Value][] {
+  fields(): [string, Value[]][] {
```
zslayton, line 123: Unfortunately, we can't change this method signature, the `Struct` constructor's signature, or how iteration works as those would be breaking changes requiring a new major version.

I think we should leave those as-is and add new functions that allow users to opt-in to the array format. For example:

```ts
.get(name)
// vs
.getAll(name)

.fields()
// vs
.allFields()

for (let [fieldName, fieldValue] of s) {/*...*/}
// vs
for (let [fieldName, fieldValueList] of s.allFields()) {/*...*/}
```

While I think `get()`/`getAll()` is a clear single-value/list-of-values name pair, I'm less enthusiastic about `fields()`/`allFields()`. Open to suggestions there. `fields()`/`fieldsWithAllValues` is more accurate but pretty verbose.

Iteration is interesting, because it's _possible_ to support multiple fields given the existing signatures. For example, the constructor takes an iterator:

```ts
constructor(fields: Iterable<[string, Value]>, annotations: string[] = []) { /*...*/ }
```

The `fields` iterable could yield `["foo", 4]` and then `["foo", 6]`. I think the constructor can/should support that, with a later call to `.getAll("foo")` returning `[4, 6]`.

On the other hand, I _don't_ think `Struct`'s iterator should produce the same behavior because that would be a breaking change. Some number of folks may already be relying on deduplication in their iteration. I think `Struct`'s iterator should stay the same; users interested in all values can iterate over `myStruct.allFields()` (or whatever we name it) instead.

test/dom/equivalence.ts
```diff
@@ -347,4 +347,13 @@ describe("Equivalence", () => {
       )
     );
   });
+  it("equals() for Struct with duplicate fields", () => {
+    let value1: Value = load("{ foo: 'bar', foo: 'baz' }")!;
+    let value2: Value = load("{ foo: 'bar', foo: 'baz' }")!;
+    let value3: Value = load("{ foo: 'bar', foo: 'qux' }")!;
+
+    assert.isTrue(value1.equals(value2));
+    assert.isTrue(value1.equals(value2));
```
zslayton, line 356: This line is a dup of the one above it.

test/dom/equivalence.ts
```diff
@@ -347,4 +347,13 @@ describe("Equivalence", () => {
       )
     );
   });
+  it("equals() for Struct with duplicate fields", () => {
+    let value1: Value = load("{ foo: 'bar', foo: 'baz' }")!;
+    let value2: Value = load("{ foo: 'bar', foo: 'baz' }")!;
+    let value3: Value = load("{ foo: 'bar', foo: 'qux' }")!;
```
zslayton, line 353: Can you add a test that verifies that field order is not considered for equality? e.g.

```ts
{foo: 1, baz: true, foo: 2}
// is equal to
{foo: 2, foo: 1, baz: true}
```

test/dom/dom.ts
```diff
@@ -484,12 +484,37 @@ describe('DOM', () => {
...
+          'age: 55, ' +
+          'age: 41' +
+          '}'
+      )!;
+
+      assert.equal(IonTypes.STRUCT, s.getType());
+      assert.deepEqual(['foo', 'bar'], s.getAnnotations());
+
+      // Field access return last value for a fieldname
+      assert.equal(41, s.age);
```
zslayton, line 508: We'll need a test of the `getAll` behavior when it's available.

src/dom/Null.ts
zslayton: I suppose the constructor could also accept a `Iterable<[string, Value]> | Iterable<[string, Value[]]>` to be more flexible.

desaikd:
*Completed Tasks:*

1. added changes for `getAll()` and `allFields()` methods.
2. changed `ionEquals()` method for checking unordered duplicate fields.
2. Tests suite changes for all the above tasks.

*Questions:*
Could you explain with an example on how the `fields` return value would look like? Also for the iterator?
e.g.
```
{foo: 1, baz: true, foo: 2} 
```
For above Struct value what should be the return value for `fields()` as well as iterator ? 
Currently it will return something like this:
```
[['foo',[1,2]] , ['baz' , true]]
```

desaikd:
*Changes:*
```
{foo: 1, baz: true, foo: 2} 
```
`fields()` would return Struct fields as below with only the last value:
```
[['foo', 2] , ['baz' , true]]
```

- To get all the field values `allFields()` can be used.
- `toString` method returns with all the values of Struct fields.

src/dom/Struct.ts
```diff
@@ -46,8 +46,11 @@ export class Struct extends Value(
     annotations: string[] = []
   ) {
     super();
+    // Struct fields will always be stored as (fieldName, fieldValues) => (string, Value[])
     for (const [fieldName, fieldValue] of fields) {
-      this._fields[fieldName] = fieldValue;
+      this._fields[fieldName] = Array.isArray(fieldValue)
```
zslayton, line 51: `dom.List` and `dom.SExpression` inherit from `Array` -- is this able to distinguish between an `Array` and a `Value` that happens to inherit from `Array`?

desaikd: No. Its doesn't distinguish. Maybe I can reverse the check and see if it's an `instanceof Value` first and otherwise it will always be `Value[]` array because of the constructor arguments.

src/dom/Value.ts
```diff
@@ -106,6 +106,14 @@ export interface Value {
    */
   get(...pathElements: PathElement[]): Value | null;
 
+  /**
+   * For the Struct type, returns an array containing all the values for given field name or field name path
```
zslayton, line 110: We should emphasize that this is meant to be used when working with input that contains duplicate fields. Maybe:

```suggestion
   * Like the `get(PathElement[])` method, but returns *all* of the values associated 
   * with a given field name instead of just the last one. For example, for the
   * following struct:
   * 
   *     {
   *         foo: 1,
   *         foo: null,
   *         foo: true,
   *     }
   * 
   * a call to `get("foo")` would return a `dom.Boolean` representing `true`
   * while a call to `getAll("foo")` would return an `Array` of
   * `dom.Value`s: `[1, null, true]`.
```

src/dom/Value.ts
```diff
@@ -118,6 +126,12 @@ export interface Value {
    */
   fields(): [string, Value][];
 
+  /**
+   * For the Struct type, returns an array containing the field name/values pairs in the Struct;
```
zslayton, line 130: Like `get()`/`getAll()`, I would describe this function in its relation to `fields()`.

zslayton (approved):

Issue #681, opened 2021-02-25
.equals() should be lenient by default


zslayton:
Currently, the (not yet released) `.equals()` method for `dom.Value` subclasses is strict by default. It checks for Ion equivalence between the two values being compared. JS values cannot be compared and annotations must be equal. This functionality is necessary for our test suite, but is of limited utility to end users.

I think we should make `.equals()` a loose definition of equality, comparing across Ion and JS values and ignoring annotations. Then we can provide a separate method (`ionEquals()`?) that performs the strict test.

desaikd:
Fixed with #683

PR #683, opened 2021-03-04
.equals() and .ionEquals() for dom.Value equivalence

desaikd:
*Issue #681*

*Description of changes:*
This PR works on creating two methods for `dom.Value` equivalence:

* `ionEquals`: For checking equivalence of two `dom.Value` instances
* `equals`: For checking non-strict equivalence of a `dom.Value` and another value. The other value may be a `dom.Value` or a native JS type.

*Completed Tasks:*
- Created renamed equals method as `ionEquals` to compare with `dom.Value` only. It contains options for strict/non-strict comparisons. 
- Remove `onlyCompareIon` option entirely from `ionEquals`.
- Created `equals` for JS Object equivalence with Ion Value. Allows only non-strict comparison. 
- Renamed `_ionEquals` helper method as `_valueEquals`
- Tests suite changes for using `equals` / `ionEquals` accordingly for tests.
- Removed `onlyCompareIon` option from namespace `Value.Equality`.
- Both `Sequence` and `Lob` equivalence changed to use `ionEquals` or `equals` based on `onlyCompareIon` for child value comparisons.

*Questions:*
1.  Any more tests that should be added here? 

*Equivalence reference:*

**Strict** equivalence refers to Ion data model equivalence as defined in Ion Equivalence[1] and by Ion Specification[2]

**Structural** or **non-strict** equivalence follows the same rules as strict equivalence, except that
1. Annotations are not considered, and
2. Timestamps that represent the same instant in time are always
     considered equivalent.
     
[1] https://www.javadoc.io/doc/com.amazon.ion/ion-java/latest/com/amazon/ion/util/Equivalence.html
[2]  http://amzn.github.io/ion-docs/docs/spec.html

### DIFF
--- a/src/dom/Boolean.ts
+++ b/src/dom/Boolean.ts
@@ -59,7 +59,7 @@ export class Boolean extends Value(
     writer.writeBoolean(this.booleanValue());
   }
 
-  _ionEquals(
+  _valueEquals(
     other: any,
     options: {
       epsilon?: number | null;
--- a/src/dom/Boolean.ts
+++ b/src/dom/Boolean.ts
@@ -75,13 +75,12 @@ export class Boolean extends Value(
   ): boolean {
     let isSupportedType: boolean = false;
     let valueToCompare: any = null;
-    if (options.onlyCompareIon) {
-      // `compareOnlyIon` requires that the provided value be an ion.dom.Boolean instance.
-      if (other instanceof Boolean) {
-        isSupportedType = true;
-        valueToCompare = other.booleanValue();
-      }
-    } else {
+    // `compareOnlyIon` requires that the provided value be an ion.dom.Boolean instance.
+    if (other instanceof Boolean) {
+      isSupportedType = true;
+      valueToCompare = other.booleanValue();
+    }
+    if (!options.onlyCompareIon) {
       // We will consider other Boolean-ish types
       if (typeof other === "boolean" || other instanceof global.Boolean) {
         isSupportedType = true;
--- a/src/dom/Boolean.ts
+++ b/src/dom/Boolean.ts
@@ -75,12 +75,11 @@ export class Boolean extends Value(
   ): boolean {
     let isSupportedType: boolean = false;
     let valueToCompare: any = null;
-    // `compareOnlyIon` requires that the provided value be an ion.dom.Boolean instance.
+    // if the provided value is an ion.dom.Boolean instance.
     if (other instanceof Boolean) {
       isSupportedType = true;
       valueToCompare = other.booleanValue();
-    }
-    if (!options.onlyCompareIon) {
+    } else if (!options.onlyCompareIon) {
       // We will consider other Boolean-ish types
       if (typeof other === "boolean" || other instanceof global.Boolean) {
         isSupportedType = true;
--- a/src/dom/Decimal.ts
+++ b/src/dom/Decimal.ts
@@ -65,28 +65,27 @@ export class Decimal extends Value(
       ignoreAnnotations?: boolean;
       ignoreTimestampPrecision?: boolean;
       onlyCompareIon?: boolean;
-      equals: boolean;
+      coerceNumericType: boolean;
     } = {
       epsilon: null,
       ignoreAnnotations: false,
       ignoreTimestampPrecision: false,
       onlyCompareIon: true,
-      equals: false,
+      coerceNumericType: false,
     }
   ): boolean {
     let isSupportedType: boolean = false;
     let valueToCompare: any = null;
-    if (options.onlyCompareIon) {
-      // `compareOnlyIon` requires that the provided value be an ion.dom.Decimal instance.
-      if (other instanceof Decimal) {
-        isSupportedType = true;
-        valueToCompare = other.decimalValue();
-      }
-      if (options.equals === true && other instanceof Float) {
-        isSupportedType = true;
-        valueToCompare = new ion.Decimal(other.toString());
-      }
-    } else {
+    // `compareOnlyIon` requires that the provided value be an ion.dom.Decimal instance.
+    if (other instanceof Decimal) {
+      isSupportedType = true;
+      valueToCompare = other.decimalValue();
+    }
+    if (options.coerceNumericType === true && other instanceof Float) {
+      isSupportedType = true;
+      valueToCompare = new ion.Decimal(other.toString());
+    }
+    if(!options.onlyCompareIon) {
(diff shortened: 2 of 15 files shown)


src/dom/Value.ts
```diff
@@ -206,18 +206,21 @@ export interface Value {
...
     options?: {
       epsilon?: number | null;
       ignoreAnnotations?: boolean;
       ignoreTimestampPrecision?: boolean;
-      onlyCompareIon?: boolean;
     }
   ): boolean;
+
+  /**
+   * For checking non-strict equivalence of an Ion Values represented by dom.Value with the other value represented
```
zslayton, line 220:
```suggestion
   * Compares this instance of dom.Value to the provided value and returns true
   * if they are equal. If `other` is a dom.Value, this comparison checks for structural
   * (or "non-strict") equivalence. If `other` is a native JS value, this comparison will
   * convert it to an Ion value and then check for structural equivalence.
```

test/dom/equivalence.ts
```diff
@@ -74,41 +74,41 @@ describe("Equivalence", () => {
...
-    );
+    assert.isTrue(float1.equals(1.5));
+    assert.isFalse(float1.ionEquals(1.5));
+    assert.isFalse(float1.equals(1.2));
+    // assert.isTrue(float1.equals(1.2, {epsilon: 0.5}));
+    assert.isTrue(float1.equals(new Number(1.5)));
+    assert.isFalse(float1.equals(new Number(1.2)));
+    // assert.isTrue(
+    //     float1.equals(new Number(1.2), {epsilon: 0.5})
+    // );
```
zslayton, line 95: Some leftover comments here.

src/dom/Value.ts
zslayton: Sorry, I just re-read the implementation for `_valueEquals` and remembered it doesn't convert JS types to Ion types during the comparison. That part of my suggested edit wouldn't be accurate.

desaikd:
*Changes:*

1. Made `ionEquals` accept only `dom.Value` type for comparison where comparison type can be `non strict` or `strict`.
2. changed `equals` to accept either `dom.Value` or JS object for comparison. comparison type can be only `non strict`.
3. added epsilon option into `equals`. 
4. Changed `Decimal` equivalence method `_valueEquals` to accept `Float` and compare by converting it to `Decimal`.
5. Changes in test suite for `epsilon` option demonstration in `equals`.

src/dom/Value.ts
```diff
@@ -217,10 +217,21 @@ export interface Value {
...
-   * For checking non-strict equivalence of an Ion Values represented by dom.Value with the other value represented
-   * by an JS object.
+   * Compares this instance of dom.Value to the provided value and returns true
+   * if they are equal. If `other` is a dom.Value, this comparison checks for structural
+   * (or "non-strict") equivalence. If `other` is a native JS value, this comparison will
+   * convert it to an Ion value and then check for structural equivalence.
+   *
+   * @param other                       other Ion Value or Js Object to be compared with this Ion Value.
+   * @param options                     options provided for equivalence as below
+   *        epsilon                     used by Float for an equality with given epsilon precision. (Default: null)
```
zslayton, line 227: Now that `equals` allows for comparisons across numeric types (e.g. `someFloat.equals(someDecimal)`), should `epsilon` has to be considered across types too?

I'm inclined to say yes for consistency, but `ion.Decimal` doesn't currently make it very easy to do arithmetic so it might take non-trivial effort to implement.

src/dom/Value.ts
```diff
@@ -217,10 +217,21 @@ export interface Value {
   ): boolean;
 
   /**
-   * For checking non-strict equivalence of an Ion Values represented by dom.Value with the other value represented
-   * by an JS object.
+   * Compares this instance of dom.Value to the provided value and returns true
+   * if they are equal. If `other` is a dom.Value, this comparison checks for structural
+   * (or "non-strict") equivalence. If `other` is a native JS value, this comparison will
+   * convert it to an Ion value and then check for structural equivalence.
```
zslayton, line 223: > If `other` is a native JS value, this comparison will convert it to an Ion value and then check for structural equivalence.

I know this was originally my phrasing suggestion, but I think we determined that it's not accurate. We do the comparisons without converting the JS value to Ion first.

desaikd:
*Changes:*
1. Added support `Float` and `Decimal` comparisons when using `equals`.
2. Remove `Value.Equality` namespace.
3. Tests suite changes for removing `Value.Equality`.
4. added `equals` option to `_valueEquals` private helper method.

src/dom/Boolean.ts
```diff
@@ -75,13 +75,12 @@ export class Boolean extends Value(
...
     let isSupportedType: boolean = false;
     let valueToCompare: any = null;
-    if (options.onlyCompareIon) {
-      // `compareOnlyIon` requires that the provided value be an ion.dom.Boolean instance.
-      if (other instanceof Boolean) {
-        isSupportedType = true;
-        valueToCompare = other.booleanValue();
-      }
-    } else {
+    // `compareOnlyIon` requires that the provided value be an ion.dom.Boolean instance.
```
zslayton, line 78: Now that we're not checking `if (options.onlyCompareIon)` here, I think this comment may be confusing. (Same thought in the other similarly refactored areas.)

src/dom/Boolean.ts
```diff
@@ -75,13 +75,12 @@ export class Boolean extends Value(
...
+    if (other instanceof Boolean) {
+      isSupportedType = true;
+      valueToCompare = other.booleanValue();
+    }
+    if (!options.onlyCompareIon) {
```
zslayton, line 83: This could be an `else if`, I think?

src/dom/Decimal.ts
```diff
@@ -66,26 +65,27 @@ export class Decimal extends Value(
...
+    // `compareOnlyIon` requires that the provided value be an ion.dom.Decimal instance.
+    if (other instanceof Decimal) {
+      isSupportedType = true;
+      valueToCompare = other.decimalValue();
+    }
+    if (options.coerceNumericType === true && other instanceof Float) {
+      isSupportedType = true;
+      valueToCompare = new ion.Decimal(other.toString());
+    }
+    if (!options.onlyCompareIon) {
```
zslayton, line 88: I think these can be `else if`s too. (I'll stop adding this comment from here on.)

zslayton (approved):
Looks good! Some minor cleanup suggestions below.

PR #695, opened 2021-06-18
added README for dom equals API

desaikd:
*Description of changes:*
This PR adds description and usage for equals API of `dom.Value`.

*Changes:*
- description of `equals` and `ionEquals` methods for `dom.Value`.
- added examples for usage of both methods.

### DIFF
--- a/src/dom/README.md
+++ b/src/dom/README.md
@@ -264,3 +264,140 @@ assert.isTrue(string === "" + ionString);
 assert.isTrue(string === ionString.stringValue());
 assert.isFalse(ionString === ionString2);
 ```
+
+### `dom.Value` equivalence
+
+dom.Value object equivalence is implemented by two methods ionEquals(compares two dom.Value objects based on strict or non-strict comparison type) and equals(compares a dom.Value object with a corresponding JS object).
+
+#### `ionEquals`:
+
+This method compares a dom.Value object with another dom.Value object with either strict or non-strict comparison type. 
+
+*Strict equivalence* refers to Ion data model equivalence as defined in Ion Equivalence (https://www.javadoc.io/doc/com.amazon.ion/ion-java/latest/com/amazon/ion/util/Equivalence.html) and by Ion Specification (http://amzn.github.io/ion-docs/docs/spec.html)
+*Structural or non-strict equivalence* follows the same rules as strict equivalence,except that
+
+* 1. Annotations are not considered, and
+* 2. Timestamps that represent the same instant in time are always considered equivalent.
+
+#### `equals`:
+
+This method compares a dom.Value object with a corresponding JS object. 
+
+#### Integer Example:
+```javascript
+let int1: Value = load("foo::bar::7");
+let int2: Value = load("foo::bar::7");
+let int3: Value = load("7");
+
+// Equivalence between two Ion DOM Values
+int1.ionEquals(int2); // returns true
+int1.ionEquals(int3); // returns false as annotations didn't match in a strict equivalence
+int1.ionEquals(int3, {ignoreAnnotations: true}); // returns true for non-strict equivalence
+
+// Equivalence between JS Value and Ion DOM Value
+int1.equals(7); // returns true
+int1.ionEquals(7); // returns false
+```
+
+#### Float Example:
+```javascript
+let float1: Value = load("baz::qux::15e-1");
+let float2: Value = load("baz::qux::15e-1");
+let float3: Value = load("15e-1");
+let float4: Value = load("1.5");
+let float5: Value = load("12e-1");
+
+// Equivalence between two Ion DOM Values
+float1.ionEquals(float2); // returns true
+float1.ionEquals(float3); // return false as annotations didn't match in a strict equivalence
+
+float1.ionEquals(float4); // returns false as Decimal and Float values will not be equivalent
+float3.ionEquals(float5, {epsilon: 0.5}); // returns true as both values are equivalent by epsilon value
+float3.ionEquals(float5, {epsilon: 0.2}); // returns false as both values are not equivalent by epsilon value
+
+// Equivalence between JS Value and Ion DOM Value
+float1.equals(1.5); // returns true
+float1.ionEquals(1.5); // returns false
+```
+
+#### Timestamp example:
+```javascript
+let timestamp1: Value = load("DOB::2020-01-16T20:15:54.066Z");
+let timestamp2: Value = load("DOB::2020-01-16T20:15:54.066Z");
+let timestamp3: Value = load("DOB::2001T");
+let timestamp4: Value = load("DOB::2001-01-01T");
+
+// Equivalence between two Ion DOM Values
+timestamp1.ionEquals(timestamp2); // returns true 
+timestamp1.ionEquals(timestamp3); // returns false as in strict mode the precision and local offsets are also compared
+timestamp3.ionEquals(timestamp4, {ignoreTimestampPrecision: true}); // retruns true as in non strict comparison precision and local offset are ignored along with annotations
+
+// Equivalence between JS Value and Ion DOM Value
+timestamp1.equals(new Date("2020-01-16T20:15:54.066Z")); // returns true
+timestamp1.equals(new Date("2020-02-16T20:15:54.066Z")); // returns false 
+```
+
+#### List Example:
+```javascript
+let list1: Value = load('planets::["Mercury", "Venus", "Earth", "Mars"]');
+let list2: Value = load('planets::["Mercury", "Venus", "Earth", "Mars"]');
+let list3: Value = load('planets::["Mercury", "Venus"]');
+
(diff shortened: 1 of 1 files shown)


src/dom/README.md
```diff
@@ -264,3 +264,140 @@ assert.isTrue(string === "" + ionString);
 assert.isTrue(string === ionString.stringValue());
 assert.isFalse(ionString === ionString2);
 ```
+
+### `dom.Value` equivalence
+
+dom.Value object equivalence is implemented by two methods ionEquals(compares two dom.Value objects based on strict or non-strict comparison type) and equals(compares a dom.Value object with a corresponding JS object).
```
jobarr-amzn, line 270: We ought to be using ``` `dom.Value` ```, ``` `ionEquals(compares two dom.Value objects based on strict or non-stric comparison type)` ```, I think. Code-formatted.

src/dom/README.md
```diff
@@ -264,3 +264,140 @@ assert.isTrue(string === "" + ionString);
...
+
+#### `ionEquals`:
+
+This method compares a dom.Value object with another dom.Value object with either strict or non-strict comparison type. 
+
+*Strict equivalence* refers to Ion data model equivalence as defined in Ion Equivalence (https://www.javadoc.io/doc/com.amazon.ion/ion-java/latest/com/amazon/ion/util/Equivalence.html) and by Ion Specification (http://amzn.github.io/ion-docs/docs/spec.html)
+*Structural or non-strict equivalence* follows the same rules as strict equivalence,except that
```
jobarr-amzn, line 277: Should these be links? E.g. [Ion Equivalence](https://www.javadoc.io/doc/com.amazon.ion/ion-java/latest/com/amazon/ion/util/Equivalence.html)?

src/dom/README.md
```diff
@@ -264,3 +264,140 @@ assert.isTrue(string === "" + ionString);
...
+float1.ionEquals(float2); // returns true
+float1.ionEquals(float3); // return false as annotations didn't match in a strict equivalence
+
+float1.ionEquals(float4); // returns false as Decimal and Float values will not be equivalent
+float3.ionEquals(float5, {epsilon: 0.5}); // returns true as both values are equivalent by epsilon value
+float3.ionEquals(float5, {epsilon: 0.2}); // returns false as both values are not equivalent by epsilon value
+
+// Equivalence between JS Value and Ion DOM Value
+float1.equals(1.5); // returns true
+float1.ionEquals(1.5); // returns false
```
jobarr-amzn, line 320: I think this particularly this example (and also the preceding one) would be clearer without variables. Reading this I'm bouncing back and forth constantly from definitions for `floatN` and their uses.

If we must have variables names, let the names be descriptive and helpful. `same1`, `same2`, `unannotated`, `ion_float`, `ion_decimal`, `nearby_float`... something like that?

src/dom/README.md
```diff
@@ -264,3 +264,140 @@ assert.isTrue(string === "" + ionString);
...
+let list2: Value = load('planets::["Mercury", "Venus", "Earth", "Mars"]');
+let list3: Value = load('planets::["Mercury", "Venus"]');
+
+// Equivalence between two Ion DOM Values
+list1.ionEquals(list2) // returns true
+list1.ionEquals(list3) // returns false because the lists have different length
+
+// Equivalence between JS Value and Ion DOM Value
+list1.equals(["Mercury", "Venus", "Earth", "Mars"]); // returns true
+list1.equals(["Mercury", "Venus", "Earth"]); // returns false because the lists differ with one lement "Mars"
```
jobarr-amzn, line 352: typo: element

src/dom/README.md
```diff
@@ -264,3 +264,140 @@ assert.isTrue(string === "" + ionString);
...
+```
+
+####Struct Example:
+```javascript
+let struct1: Value = load(
+  "foo::bar::{" +
+    "name: {" +
+    'first: "John", ' +
+    'middle: "Jacob", ' +
+    'last: "Jingleheimer-Schmidt",' +
```
jobarr-amzn, line 362: 😀

src/dom/README.md
```diff
@@ -264,3 +264,140 @@ assert.isTrue(string === "" + ionString);
...
+struct1.equals(
+    {
+      name: {
+        first: "John",
+        middle: "Jacob",
+        last: "Jingleheimer-Schmidt",
+      },
+      age: 41,
+    }
+); // retrurns true
```
jobarr-amzn, line 402: typo: returns

src/dom/README.md
desaikd: Yes, I think we can have variable names descriptive for making it more readable. I will add this change along with other typos and minor changes.

src/dom/README.md
```diff
@@ -264,3 +264,140 @@ assert.isTrue(string === "" + ionString);
...
+
+*Strict equivalence* refers to Ion data model equivalence as defined in Ion Equivalence (https://www.javadoc.io/doc/com.amazon.ion/ion-java/latest/com/amazon/ion/util/Equivalence.html) and by Ion Specification (http://amzn.github.io/ion-docs/docs/spec.html)
+*Structural or non-strict equivalence* follows the same rules as strict equivalence,except that
+
+* 1. Annotations are not considered, and
+* 2. Timestamps that represent the same instant in time are always considered equivalent.
+
+#### `equals`:
+
+This method compares a dom.Value object with a corresponding JS object.
```
zslayton, line 284: I think you should go into a bit more detail here.
* How are annotations handled? (I think they're ignored, right?) 
* For Ion types that don't have an obvious mapping to JS, what types can be considered equal? 
   * Is an `ion.dom.Blob` equal to an `Array`? A `Utf8Array`? A `Buffer`?)
   * Can I test an `ion.dom.Decimal` for equality with a `Number`?
   * etc

src/dom/README.md
```diff
@@ -267,94 +267,111 @@ assert.isFalse(ionString === ionString2);
...
-#### Integer Example:
+This method compares a `dom.Value` object with a corresponding JS object as defined in [Ion Data Types section](#iondom-data-types). 
+If the other object is a `dom.Value` instance then this comparison checks for structural (or "non-strict") equivalence.
+
+_NOTE:_ 
+
+* This method also provides an optional parameter `epsilon` for `Float` type comparison.
+* This method allows coerce numeric types comparison for `Decimal` and `Float` values:
+    * Comparison of `Float` with `Decimal` or `number` type is allowed.
+    * Comparison of `Decimal` with `Float` or `number` type is allowed.
```
zslayton, line 301: Can this be simplified to:
```suggestion
* This method allows `Decimal`, `Float`, and `number` values to be compared to one another.
```
?

Does it _not_ allow `Integer` values to be compared to the others?

src/dom/README.md
```diff
@@ -267,94 +267,111 @@ assert.isFalse(ionString === ionString2);
...
-* 1. Annotations are not considered, and
-* 2. Timestamps that represent the same instant in time are always considered equivalent.
+Optional parameters that can be used with this method:
 
-#### `equals`:
+| option                      | description                                                                    |
+| --------------------------- | :------------------------------------------------------------------------------|
+|`epsilon`                    | used by `Float` for an equality with given epsilon precision. (Default: `null`)|
+|`ignoreAnnotations`          | specifies whether to ignore annotations or not for equality. (Default: `false`)|
+|`ignoreTimestampPrecision`   | specifies whether to ignore timestamp local offset and precision or not for equality. (Default: `false`)|
```
zslayton, line 289: Consider making `default` its own column.

src/dom/README.md
```diff
@@ -267,94 +267,111 @@ assert.isFalse(ionString === ionString2);
...
 
 *Strict equivalence* refers to Ion data model equivalence as defined in Ion Equivalence (https://www.javadoc.io/doc/com.amazon.ion/ion-java/latest/com/amazon/ion/util/Equivalence.html) and by Ion Specification (http://amzn.github.io/ion-docs/docs/spec.html)
-*Structural or non-strict equivalence* follows the same rules as strict equivalence,except that
+*Structural* (or *non-strict*) *equivalence* follows the same rules as strict equivalence,except that:
+* Annotations are not considered, and
+* Timestamps that represent the same instant in time are always considered equivalent.
 
-* 1. Annotations are not considered, and
-* 2. Timestamps that represent the same instant in time are always considered equivalent.
+Optional parameters that can be used with this method:
```
zslayton, line 283:
```suggestion
This method accepts an optional configuration object that allows you to modify how equivalence is determined. It recognizes the following properties:
```

src/dom/README.md
desaikd: No it only allows `number` and `JSBI` with `Integer`.

zslayton: That's surprising to me--I expected value-based `equals` to allow `Integer`-to-`Float`/`Decimal` comparisons. I've opened #696 to track this.

zslayton (approved):

PR #699, opened 2021-12-08
adds README changes for including shared symbol table usage with an example

desaikd:
*Description of changes:*
This PR works on changing the README to add an example usage of shared symbol table with `ion-js`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,11 +155,34 @@ at `dist/browser/js/ion-bundle.js`.
 | `sexp`                    | yes           | yes           | none                                          |
 | annotations               | yes           | yes           | none                                          |
 | local symbol tables       | yes           | yes           | none                                          |
-| shared symbol tables      | no            | no            | none                                          |
+| shared symbol tables      | no            | yes           | none                                          |
 
 Notes:
 * [test/iontests.ts](https://github.com/amzn/ion-js/blob/master/test/iontests.ts) defines multiple skipList variables
   referencing test vectors that are not expected to work at this time.
+
+* ion-js supports shared symbol table for Ion Binary. Below is an example of how shared symbol table can be used here: 
+```javascript
+// create a shared symbol table and use it with binary writer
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -155,12 +155,35 @@ at `dist/browser/js/ion-bundle.js`.
...
 
 Notes:
 * [test/iontests.ts](https://github.com/amzn/ion-js/blob/master/test/iontests.ts) defines multiple skipList variables
   referencing test vectors that are not expected to work at this time.
 
+* ion-js supports shared symbol table for Ion Binary. Below is an example of how shared symbol table can be used here: 
+```javascript
+// create a shared symbol table and use it with binary writer
+let sst = new SharedSymbolTable('foo', 1, ['id', 'name']);
+let writer = new BinaryWriter(new LocalSymbolTable(new Import(new Import(null, getSystemSymbolTable()), sst)), new Writeable());
```
zslayton, line 168: Maybe we could break this out for teaching purposes?

```suggestion

// Create a SharedSymbolTable with the desired strings
let sharedSymbolTable = new SharedSymbolTable('foo', 1, ['id', 'name']);

// Define an import chain
// The system symbol table does not import any other tables (`null` below)
let systemSymbolTableImport = new Import(null, getSystemSymbolTable());
// The shared symbol table imports the system symbol table
let sharedSymbolTableImport = new Import(systemSymbolTableImport, sharedSymbolTable);

// Create a local symbol table that imports the shared symbol table
let localSymbolTable = new LocalSymbolTable(sharedSymbolTableImport);

// Create a writer that uses our new local symbol table
let writer = new BinaryWriter(localSymbolTable, new Writeable());
```

zslayton: (I haven't tested this code, please let me know if it's incorrect or somehow misleading!)

desaikd: Yes, I tested this code change and it works fine by adding the shared symbol table correctly.

zslayton (approved):

PR #702, opened 2022-02-15
Updates ion-js to be compatible with jsbi@^4.0.0

### LINKED ISSUE #700: DOM Decimal: super constructor call needs update (and looks buggy)
See [here](https://github.com/amzn/ion-js/blob/ada4439df48b0467c26ce6487bed78f050716360/src/dom/Decimal.ts#L34):
```
  constructor(value: ion.Decimal, annotations: string[] = []) {
    super(...[value.getCoefficient(), value.getExponent(), value.isNegative()]);
    this._decimalValue = value;
    this._numberValue = value.numberValue();
    this._setAnnotations(annotations);
  }
```
Since `class Decimal extends Value(Number, ...)`, `super` is effectively the plain JavaScript `Number` constructor. There are two issues here:

(1) `value.getCoefficient()` returns a JSBI instance. Newer versions of JSBI disallow `new Number(some_jsbi)` constructor calls. The supported alternative is to use `JSBI.toNumber(some_jsbi)` instead. This came up at https://github.com/GoogleChromeLabs/jsbi/issues/85.
(Background: this change was made in an effort to disallow invalid conversions that wouldn't work any more once JSBI is compiled away to native BigInts. `Number(42n)` is fine, but `+42n` is not; so JSBI now disallows `+some_jsbi`, which as "collateral damage" also disallows `Number(some_jsbi)`. FWIW, JSBI's [documentation](https://github.com/GoogleChromeLabs/jsbi#how) has always recommended to use `JSBI.toNumber`, the fact that `Number(some_jsbi)` also worked was accidental.)

(2) The `Number` constructor takes only one argument. Calling it with three arguments looks like a bug, as `value.getExponent()` and `value.isNegative()` will just be ignored. (Presumably the value passed to the super ctor is never used for anything, which would explain why this hasn't caused issues so far?)

I would suggest to fix both issues at once by changing the super call to `super(...[value.numberValue()])`. (Or, more simply, `super(value.numberValue())`. Or, if my suspicion is correct that the value is indeed never used, just drop the super call entirely.)

Apologies in advance if I've misunderstood anything.

popematt:
**Issue #, if available:**

Fixes #700 

**Description of changes:**

Updates usage of JSBI so that consumers of `ion-js` can use JSBI 3 or 4. JSBI 4 no longer allows certain built-in operators to be used with JSBI, and disallows new Number(some_jsbi) constructor calls. (See https://github.com/GoogleChromeLabs/jsbi/issues/85 and #700 for more details.) 

Anywhere that a JSBI was being coerced to a `Number` or a `string` was replaced with an explicit conversion (using `toString()`, `toNumber()`, or a template string). In the `Decimal` constructor, there was an indirect invocation of `new Number(some_jsbi)` that was updated to use `JSBI.toNumber()` instead.

*By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.*

### DIFF
--- a/src/events/IonEvent.ts
+++ b/src/events/IonEvent.ts
@@ -870,7 +870,7 @@ class IonListEvent extends AbsIonContainerEvent {
       ) {
         return new ComparisonResult(
           ComparisonResultType.NOT_EQUAL,
-          `${child.ionValue} vs. ${expectedContainer[1].ionValue}`,
+          `${child.ionValue} vs. ${expectedContainer[i].ionValue}`,
           i + 1,
           i + 1
         );
--- a/src/events/IonEvent.ts
+++ b/src/events/IonEvent.ts
@@ -468,7 +468,7 @@ class IonIntEvent extends AbstractIonEvent {
     }
     return new ComparisonResult(
       ComparisonResultType.NOT_EQUAL,
-      this.ionValue + " vs. " + expected.ionValue
+      `${this.ionValue} vs. ${expected.ionValue}`
     );
   }
 
@@ -870,7 +870,7 @@ class IonListEvent extends AbsIonContainerEvent {
       ) {
         return new ComparisonResult(
           ComparisonResultType.NOT_EQUAL,
-          child.ionValue + " vs. " + expectedContainer[i].ionValue,
+          `${child.ionValue} vs. ${expectedContainer[1].ionValue}`,
           i + 1,
           i + 1
         );
--- a/package.json
+++ b/package.json
@@ -7,7 +7,9 @@
   "scripts": {
     "commit": "git-cz",
     "prepare": "grunt release",
-    "test": "nyc mocha",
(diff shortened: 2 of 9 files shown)


src/events/IonEvent.ts
```diff
@@ -870,7 +870,7 @@ class IonListEvent extends AbsIonContainerEvent {
       ) {
         return new ComparisonResult(
           ComparisonResultType.NOT_EQUAL,
-          child.ionValue + " vs. " + expectedContainer[i].ionValue,
+          `${child.ionValue} vs. ${expectedContainer[1].ionValue}`,
```
zslayton, line 873: Was this change intentional?
```suggestion
          `${child.ionValue} vs. ${expectedContainer[i].ionValue}`,
```

zslayton (approved):
👍

PR #715, opened 2022-05-09
Remove dependency on `global` global variable

YourFin:
*Issue #, if available:*

N/A

*Description of changes:*

The `global` global variable [1] is a node-ism, and is not available on
browsers and other js runtime environments. Browser bundlers typically overcome this with a polyfill that defines `global` as a reference to the `window` variable [2], but not all runtimes have an equivalent option. As a result, use of `global` tends to cause headaches:

https://github.com/aws/aws-sdk-js/issues/2141

https://github.com/angular/angular-cli/issues/8160

https://stackoverflow.com/questions/57361546/referenceerror-global-is-not-defined-with-web3

https://dev.to/richardbray/how-to-fix-the-referenceerror-global-is-not-defined-error-in-sveltekitvite-2i49

Therefore, this change replaces all references to `global` with behavior that is guaranteed by the ECMA spec, which should be available in any runtime.

[1]: https://nodejs.org/docs/latest/api/all.html#all_globals_global
[2]: https://github.com/browserify/browserify-handbook#global

### DIFF
--- a/src/dom/JsValueConversion.ts
+++ b/src/dom/JsValueConversion.ts
@@ -5,6 +5,13 @@ import { IonTypes } from "../IonTypes";
 import { _hasValue } from "../util";
 import { Value } from "./Value";
 
+// In ./Boolean.ts and ./String.ts, the native Boolean and String types
+// are shadowed by the export class defined in the file, but said class
+// definitions still need to reference the original class definitions.
+// This provides a mechanism for doing so.
+export const _NativeJsBoolean = Boolean;
+export const _NativeJsString = String;
+
 // Typescript interfaces can be used to describe classes' static methods; this can
 // be a bit surprising as the methods in the interface are not marked 'static' and
 // the class definition does not need to indicate that it implements this interface.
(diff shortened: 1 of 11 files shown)


src/dom/JsValueConversion.ts
```diff
@@ -5,6 +5,13 @@ import { IonTypes } from "../IonTypes";
 import { _hasValue } from "../util";
 import { Value } from "./Value";
 
+// In ./Boolean.ts and ./String.ts, the native Boolean and String types
+// are shadowed by the export class defined in the file, but said class
+// definitions still need to reference the original class definitions.
+// This provides a mechanism for doing so.
+export const _NativeJsBoolean = Boolean;
+export const _NativeJsString = String;
+
```
YourFin, line 14: I initially tried tackling this with

(String.ts)
```
export class String extends Value(
  global.String,
  IonTypes.STRING,
  _fromJsConstructor
) {
 ...
}
```
->
```
class StringImpl extends Value(
  String,
  IonTypes.STRING,
  _fromJsConstructor
) {
  ...
}
export { StringImpl as String };
```

But swapping the directly exported variable from the class itself to a constant screwed up the generated documentation.

jobarr-amzn (approved):

PR #718, opened 2022-05-11
adds documentation for v4.3.0

popematt:


api/classes/_comparisonresult_.comparisonresult.html
```diff
@@ -116,7 +116,7 @@ <h3>constructor</h3>
 						<li class="tsd-description">
 							<aside class="tsd-sources">
 								<ul>
-									<li>Defined in <a href="https://github.com/amzn/ion-js/blob/2cf812f/src/ComparisonResult.ts#L34">src/ComparisonResult.ts:34</a></li>
+									<li>Defined in <a href="https://github.com/popematt/ion-js/blob/9264b31/src/ComparisonResult.ts#L34">src/ComparisonResult.ts:34</a></li>
```
jobarr-amzn, line 119: I'd rather generate from amzn/ion-js, I think

jobarr-amzn (changes requested):

api/classes/_comparisonresult_.comparisonresult.html
popematt: Good catch. This happened because I generated these docs in a repo that has my fork as "origin" and `amzn/ion-js` as "upstream"

popematt:
Discarded in favor of #718

PR #720, opened 2022-05-12
Refactors the release workflow

popematt:
**Issue #, if available:**

N/A

**Description of changes:**

Refactors the release workflow to have separate jobs that can be independently retried. Also, the workflow now opens a pull request instead of pushing directly to the `gh-pages` branch.

You can view a test of this workflow that I ran on a separate branch in my fork.
* [release workflow run](https://github.com/popematt/ion-js/actions/runs/2310376548) for `v125.1.0` of `popematt/ion-js`.
* [v125.1.0 release](https://github.com/popematt/ion-js/releases/tag/v125.1.0) with the zip file that was added by the workflow
* https://github.com/popematt/ion-js/pull/1

I am creating this PR from a different branch than I ran the test to ensure that the v125.1.0 tag does not get merged into `amzn/ion-js` somehow by mistake.

The `.npmignore` change looks like a lot, but it is just adding `.github`, `.nyc_output/`, `.prettierignore`, `test`, `test-driver`, and `tslint.json`, and then I alphabetized the list.

*By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.*

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,16 +5,13 @@ on:
     types: [created]
 
 jobs:
-  release:
+  build:
     if: startsWith(github.ref, 'refs/tags/v')
     runs-on: ubuntu-latest
     steps:
-      - name: set env
-        run: echo "RELEASE_TAG=${GITHUB_REF#refs/tags/}" >> $GITHUB_ENV
       - uses: actions/checkout@v3
         with:
           submodules: recursive
-      # Setup .npmrc file to publish to npm
       - uses: actions/setup-node@v3
         with:
           node-version: '16.x'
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -5,45 +5,83 @@ on:
...
         run: ./node_modules/.bin/grunt release
+      - uses: actions/upload-artifact@v3
+        with:
+          name: dist
+          path: dist/
+      - uses: actions/upload-artifact@v3
+        with:
+          name: docs
+          path: docs/
+  publish-to-github:
```
cheqianh, line 30: Ah nice we don't have to manually upload artifacts to GitHub release, I should apply it to ion-hive-serde release too. I didn't use gh-cli in GHA before, do we need authentication somewhere first or the corresponding workflow will recognize it?

cheqianh (approved):
Looks good to me!

.github/workflows/release.yml
popematt: You authenticate by putting `GITHUB_TOKEN` in the `env` of the `run` step where you are using the `gh` CLI.

cheqianh: Thanks, found it.

PR #723, opened 2022-09-08
Drops support for Node 10.x, 12.x and adds 18.x

popematt:
**Issue #, if available:**

None

**Description of changes:**

This updates the supported versions of Node.js and bumps the version to `5.0.0-SNAPSHOT` (since dropping support for old node is theoretically a backwards incompatible change).

Note that this PR is against the `v5.0.0-development` branch, not against master.

_By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice._

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,9 +2,9 @@ name: Node.js CI
 
 on:
   push:
-    branches: [ master ]
+    branches: [ master, v5.0.0-development ]
   pull_request:
-    branches: [ master ]
+    branches: [ master, v5.0.0-development ]
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


popematt:
`ion-test-driver` failure is independent of my changes. Either way, my changes would not affect the outcome of those tests.

.github/workflows/build.yml
```diff
@@ -2,25 +2,26 @@ name: Node.js CI
...
 jobs:
   build:
 
     runs-on: ubuntu-latest
 
     strategy:
+      fail-fast: false
       matrix:
-          node-version: [10.x, 12.x, 14.x, 16.x]
+          node-version: [14.x, 16.x, 18.x]
```
linlin-s, line 17: Just a quick question here. According to the README.md, should this node-version also include 20?

linlin-s (approved):

.github/workflows/build.yml
popematt: My mistake. README is wrong—Node 20 hasn't been released yet.

linlin-s: > My mistake. README is wrong—Node 20 hasn't been released yet.

Other than that, looks good to me.

linlin-s (approved):

Issue #729, opened 2022-11-11
Migrate from TSLint to ESLint


amazlite:
TSLint is not recommended for linting anymore, switch to ESLint + TypeScript plugins instead.

jobarr-amzn:
We'd be happy to take a pull request here @version0chiro  :)

PR #730, opened 2022-11-11
Migrate to eslint

amazlite:
*Issue #, if available:*

#729 

*Description of changes:*

TSLint is not evolving anymore, it is recommended to migrate to ESLint and that's what this PR is doing. I have also disabled all recommended rules that fail for current code base - can be fixed in further PRs. Notes:
* TSLint was not connected to builds at all, and when I tried to run it manually multiple checks failed. I am enabling linting checks as a part of build process.
* We cannot use the latest versions of ESLint and plugins as this library still supports Node v10.


zslayton:
Thanks for putting this together! I notice it's still in `draft` mode, were you planning to wait until the dependency on node v10 was removed?

zslayton:
Could you have this target the `5.0.0-development` branch so we can include it in the next release? The dependency on Node v10 has already been removed there.

Issue #693, opened 2021-05-07
Add an allElements() method alongside elements()


zslayton:
v4.2.0 added support for structs with repeated field names, e.g.
```js
{
  foo: 1,
  foo: true,
  foo: bar
}
```
To preserve backwards compatibility, existing struct accessor methods (`get`, `fields`) continued to return deduplicated results, using the value associated with the last appearance of each field name. New flavors of those methods (`getAll`, `allFields`) were added that would instead return an array with every value associated each field name. 

We have not yet added an `allElements` method for the `elements` method. A call to `elements` with the sample struct above would return `[bar]` while the `allElements` method would return `[1, true, bar]`.

desaikd:
Express `allElements()` behavior in `elements()` with `v5.0.0` release.

desaikd:
fixed with #754

PR #735, opened 2022-11-28
adds `allElements()` method in Value API

desaikd:
*Issue #693*

*Description of changes:*
This PR works on adding `allElements()` along with `elements()` for Value API.

*Test:*
Adds unit tests for `allElements()`.

### DIFF
--- a/src/dom/Struct.ts
+++ b/src/dom/Struct.ts
@@ -141,10 +141,14 @@ export class Struct extends Value(
 
   elements(): Value[] {
     let singleValueFields = Object.create(null);
-    for (const [fieldName, values] of this.allFields()) {
-      singleValueFields[fieldName] = values[values.length - 1];
+    for (const [fieldName, value] of this.fields()) {
+      singleValueFields[fieldName] = value;
     }
     return Object.values(singleValueFields);
+  }
+
+  allElements(): Value[][] {
+    return Object.values(this._fields);
   }
 
   [Symbol.iterator](): IterableIterator<[string, Value]> {
--- a/src/dom/Struct.ts
+++ b/src/dom/Struct.ts
@@ -147,8 +147,9 @@ export class Struct extends Value(
     return Object.values(singleValueFields);
   }
 
-  allElements(): Value[][] {
-    return Object.values(this._fields);
+  allElements(): Value[] {
+    // TODO: this can use flat()  method once we have esnext
+    return [].concat.apply([], Object.values(this._fields));
   }
 
   [Symbol.iterator](): IterableIterator<[string, Value]> {
--- a/src/dom/Value.ts
+++ b/src/dom/Value.ts
@@ -164,7 +164,7 @@ export interface Value {
   elements(): Value[];
 
   /**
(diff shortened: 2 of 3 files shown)


src/dom/Struct.ts
```diff
@@ -141,12 +141,16 @@ export class Struct extends Value(
...
     let singleValueFields = Object.create(null);
-    for (const [fieldName, values] of this.allFields()) {
-      singleValueFields[fieldName] = values[values.length - 1];
+    for (const [fieldName, value] of this.fields()) {
+      singleValueFields[fieldName] = value;
     }
     return Object.values(singleValueFields);
   }
 
+  allElements(): Value[][] {
```
zslayton, line 150: This is returning an array-of-arrays, but I think we want it to have the same method signature as `elements` on line 142 above.

You can use the [`Array` method `flat()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat) to achieve this.
```ts
return Object.values(this._fields).flat();
```

src/dom/Value.ts
```diff
@@ -159,9 +159,32 @@ export interface Value {
...
 
+  /**
+   * For the Struct, List, and SExpression types, returns an array containing the container's
+   * nested values; otherwise throws an Error.
+   * If there are duplicate fields in a Struct, then it only returns all the values for that field.
+   *
+   * Like the `elements()` method, but returns an array containing the field name/value pairs with *all* of
+   * the associated values to a field name, instead of field name/value pairs with just the last associated value
+   * to a field name. For example, for the
+   * following struct:
```
zslayton, line 174: Ah, now I understand why `Struct#allElements` was returning `Value[][]`.

The `Value#elements` method makes the most sense when you're working with a `list` or `sexp`. In that use case, it's straightforward: it gives you all of the child elements of the sequence.

For `struct`s, it gets weird. It only returns the field values. If a user wanted the name/value(s) pairs, they could use `fields()` or `allFields()` to get them. Arguably, we could've made `elements()` just throw an exception if you called it on a `struct`, but we decided to make it act more or less like the other containers.

For consistency with its behavior for `list` and `sexp`, I think this should return a flat `Value[]` with a list of the struct's field values.

src/dom/Value.ts
```diff
@@ -159,9 +159,32 @@ export interface Value {
   /**
    * For the Struct, List, and SExpression types, returns an array containing the container's
    * nested values; otherwise throws an Error.
+   * If there are duplicate fields in a Struct, then it only returns the last value for that field.
    */
   elements(): Value[];
 
+  /**
+   * For the Struct, List, and SExpression types, returns an array containing the container's
```
zslayton, line 167: This PR doesn't seem to implement `allElements()` for `Struct` and `List`. Am I missing it?

desaikd: That should be just `Struct` not `SExpression` and `List`. I will change it.

desaikd:
Closing this with the intention to express `allElements()` behavior in `elements()` with `v5.0.0` release.

PR #746, opened 2023-03-02
adds changes to allow `dom.Decimal` construction from `Number` and `String`

desaikd:
### Description of changes:
This PR works on allowing dom `Decimal` construction from `Number` or `String`.

## Tests:
adds unit tests to verify construction from `Number` or `String`.

----

### DIFF
--- a/src/dom/Decimal.ts
+++ b/src/dom/Decimal.ts
@@ -60,13 +60,15 @@ export class Decimal extends Value(
       super(value.numberValue());
       this._decimalValue = value;
       this._numberValue = value.numberValue();
-    } else if (typeof value === "number")  {
+    } else if (typeof value === "number") {
       // if value is a number type
       super(value);
       this._decimalValue = new ion.Decimal("" + value);
       this._numberValue = value;
     } else {
-      throw new Error("Decimal value can only be created from number, ion.Decimal or string");
+      throw new Error(
+        "Decimal value can only be created from number, ion.Decimal or string"
+      );
     }
     this._setAnnotations(annotations);
   }
--- a/src/dom/Decimal.ts
+++ b/src/dom/Decimal.ts
@@ -34,7 +34,7 @@ export class Decimal extends Value(
 
   /**
    * Constructor.
-   * @param value         The string value to represent as a decimal.
+   * @param value         The text Ion value to be parsed as a decimal.
    * @param annotations   An optional array of strings to associate with `value`.
    */
   constructor(value: string, annotations?: string[]);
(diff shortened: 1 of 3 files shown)


src/dom/Decimal.ts
```diff
@@ -27,13 +27,44 @@ export class Decimal extends Value(
...
    */
-  constructor(value: ion.Decimal, annotations: string[] = []) {
-    super(value.numberValue());
-    this._decimalValue = value;
-    this._numberValue = value.numberValue();
+  constructor(value: ion.Decimal, annotations?: string[]);
+
+  /**
+   * Constructor.
+   * @param value         The string value to represent as a decimal.
```
zslayton, line 37:
```suggestion
   * @param value         The text Ion value to be parsed as a decimal.
```

src/dom/Decimal.ts
```diff
@@ -27,13 +27,44 @@ export class Decimal extends Value(
...
+
+  // This is the unified implementation of the above signatures and is not visible to users.
+  constructor(
+    value: ion.Decimal | string | number,
+    annotations: string[] = []
+  ) {
+    if (typeof value === "string") {
+      super(Number(value));
+      this._decimalValue = new ion.Decimal(value);
+      this._numberValue = Number(value);
```
zslayton, line 57: We can avoid doing the same conversion twice:
```suggestion
      let numberValue = Number(value);
      super(numberValue);
      this._decimalValue = new ion.Decimal(value);
      this._numberValue = numberValue;
```
I'm unsure if there's a benefit/drawback to doing `Number(value)` (`string` -> `Number`) over `this._decimalValue.numberValue()` (`string` -> `decimal` -> `Number`). The latter is more steps, but may preserve precision better? I'm not terribly concerned about it because it only affects the `Number` approximation of the `Decimal`, which will definitely preserve precision.

src/dom/Decimal.ts
```diff
@@ -27,13 +27,44 @@ export class Decimal extends Value(
...
+    } else if (value instanceof ion.Decimal) {
+      super(value.numberValue());
+      this._decimalValue = value;
+      this._numberValue = value.numberValue();
+    } else {
+      // if value is a number type
+      super(value);
```
zslayton, line 64: In a TypeScript program, this is correct. If this is run in JavaScript, I believe `value` could be another non-`number` type.
```suggestion
    } else if (typeof value === "number") {
      // if value is a number type
      super(value);
```
with a corresponding `else` that throws for any other type would be safer.

desaikd: Yes, that makes sense. I will change it.

zslayton (approved):

PR #754, opened 2023-04-20
modifies `Struct#elements()` to return all field values

desaikd:
### Description of changes:
This PR modifies `Struct#elements()` to return all field values instead of returning the last value for a duplicate field name.

----

### DIFF
--- a/src/dom/Struct.ts
+++ b/src/dom/Struct.ts
@@ -149,11 +149,7 @@ export class Struct extends Value(
   }
 
   elements(): Value[] {
-    let singleValueFields = Object.create(null);
-    for (const [fieldName, values] of this.allFields()) {
-      singleValueFields[fieldName] = values[values.length - 1];
-    }
-    return Object.values(singleValueFields);
+    return Object.values(this._fields).flat() as Value[];
   }
 
   [Symbol.iterator](): IterableIterator<[string, Value]> {
--- a/test/dom/dom.ts
+++ b/test/dom/dom.ts
@@ -546,9 +546,9 @@ describe("DOM", () => {
 
     // elements returns values for given Struct where for a fieldname it only contains the last value
     // instead of an array of all values
-    assert.equal(41, s.elements()[0]);
+    assert.equal(55, s.elements()[0]);
     assert.isFalse(Array.isArray(s.elements()[0]));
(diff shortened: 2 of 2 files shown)


src/dom/Struct.ts
```diff
@@ -149,11 +149,7 @@ export class Struct extends Value(
   }
 
   elements(): Value[] {
-    let singleValueFields = Object.create(null);
-    for (const [fieldName, values] of this.allFields()) {
-      singleValueFields[fieldName] = values[values.length - 1];
-    }
-    return Object.values(singleValueFields);
+    return Object.values(this._fields).flat() as Value[];
```
zslayton, line 152: nice!👌

src/dom/Struct.ts
```diff
@@ -149,11 +149,7 @@ export class Struct extends Value(
   }
 
   elements(): Value[] {
```
zslayton, line 151: Interestingly, our doc comment for `Value#elements` never said anything about de-duping values:

https://github.com/amazon-ion/ion-js/blob/f83fcf0e2c359b52c13ac62db149b0f2ee3ef58b/src/dom/Value.ts#L159-L163

...so now the behavior aligns with its documentation.

test/dom/dom.ts
```diff
@@ -546,9 +546,9 @@ describe("DOM", () => {
 
     // elements returns values for given Struct where for a fieldname it only contains the last value
     // instead of an array of all values
```
zslayton, line 548: This comment is now out of date.

test/dom/dom.ts
```diff
@@ -546,9 +546,9 @@ describe("DOM", () => {
...
-    assert.equal(41, s.elements()[0]);
+    assert.equal(55, s.elements()[0]);
     assert.isFalse(Array.isArray(s.elements()[0]));
-    assert.equal("Jessie", s1.elements()[0]["first"]);
+    assert.equal("John", s1.elements()[0]["first"]);
```
zslayton, line 551: This does exercise the new behavior, but it does so in a pretty subtle way; it returns the first value, which previously was skipped in favor of the second field with the same name.

Could you add some assertions for the other values that we expect `elements()` to return?

zslayton (approved):

PR #760, opened 2023-07-11
adds support for local symbol table append read

desaikd:
### Description of changes:
This PR adds local symbol table append read support for binary and text reader.

### List of changes: 
- modifies `makeSymbolTable` signature to take current LST value which will be sued to append symbols to current LST.
- If the reader finds `$ion_symbol_table` in the imports then it will perform LST append adding all the symbols to current LST.
- adds unit tests for LST append read support
----


tgregg (approved):
Looks good

PR #761, opened 2023-07-20
Updates release script

desaikd:
### Description of changes:
This PR updates the release script which was broken as per the [last release](https://github.com/amazon-ion/ion-js/actions/runs/4800737685).

----

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -31,6 +31,8 @@ jobs:
     if: startsWith(github.ref, 'refs/tags/v')
     runs-on: ubuntu-latest
     needs: build
+    permissions:
+      checks: write
     steps:
       - name: set env
         run: echo "RELEASE_TAG=${GITHUB_REF#refs/tags/}" >> $GITHUB_ENV
@@ -49,6 +51,8 @@ jobs:
     if: startsWith(github.ref, 'refs/tags/v')
     runs-on: ubuntu-latest
     needs: build
+    permissions:
+      checks: write
     steps:
       - name: set env
         run: echo "RELEASE_TAG=${GITHUB_REF#refs/tags/}" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -31,6 +31,8 @@ jobs:
     if: startsWith(github.ref, 'refs/tags/v')
     runs-on: ubuntu-latest
     needs: build
+    permissions:
+      checks: write
```
popematt, line 35: Since this job is uploading the zipped artifacts to the GH release, then it will also need `contents: write`.

The `update-docs` job will likely also need some more write permissions.

See https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

.github/workflows/release.yml
```diff
@@ -69,6 +74,10 @@ jobs:
...
+      contents: write
+      pages: write
```
popematt, line 80: You probably don't need `pages: write` since it's not directly triggering a pages build.

I don't think it needs any other permissions in order to create the branch and push the updated docs, and I suspect that it does _not_ need `pull-requests: write` because you generally don't need any special permissions to create a new PR.

popematt: I don't think you need `checks: write` for this one either.

I'm willing to try it with just `contents: write`, and see what we learn.

.github/workflows/release.yml
```diff
@@ -49,6 +52,8 @@ jobs:
     if: startsWith(github.ref, 'refs/tags/v')
     runs-on: ubuntu-latest
     needs: build
+    permissions:
+      checks: write
```
popematt, line 56: Definitely don't need `checks: write` here. Looks like this job failed last time because `grunt` wasn't installed correctly. https://github.com/amazon-ion/ion-js/actions/runs/4800737685/job/13015905967#step:7:15

desaikd: I am not sure about why the `grunt` command shows not found. In the build script it seems to work fine.

desaikd: I think we should probably add `run: npm ci` before doing `run: npm publish` to solve installing grunt or any other dependencies which might be needed for publish.  
From the [github documents](https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages) on publishing node.js package:
```
 - run: npm ci
 - run: npm publish
        env:
          NODE_AUTH_TOKEN: ${{ GITHUB_TOKEN }}
```

popematt (approved):

PR #768, opened 2023-08-26
Add `catalog` parameter to makeReader

### LINKED ISSUE #765: shared symbol tables is advertised in readme but in private implementation.
shared symbol table support is advertised in [README.md](https://github.com/amazon-ion/ion-js/blob/master/README.md?plain=1#L157) but the feature is not available because ion-js does not expose any API's that accept a `Catalog` or `SharedSymbolTable`.

The [example](https://github.com/amazon-ion/ion-js/blob/master/README.md?plain=1#L164-L193) uses `new BinaryReader(new BinarySpan(bytes), catalog);` but BinaryReader is not exported from Ion.js. `makeReader` and `loadAll` do not take a catalog argument.

I cannot import IonBinaryReader from the file the exports it because ion-js uses a `exports` map, which makes any file not listed illegal to import and ion-js only exports `Ion.ts`
```
require('ion-js/dist/commonjs/es6/IonBinaryReader.js')
Uncaught:
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './dist/commonjs/es6/IonBinaryReader.js' is not defined by "exports" in /home/ANT.AMAZON.COM/calebev/tmp/node_modules/ion-js/package.json
```

---

My recommendation is to add arguments to `makeReader` function.

```
import { makeReader, Catalog } from 'ion-js'
const catalog = new Catalog()
makeReader(buffer, catalog)
```

everett1992:
This allows users to create a reader with a catalog

fixes #765

----

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,7 +189,7 @@ catalog.add(sharedSymbolTable);
 
 // Create a reader with catalog
 let bytes = writer.getBytes();
-let reader = new BinaryReader(new BinarySpan(bytes), catalog);
+let reader = makeReader(bytes, catalog);
 ```
 
 ## Contributing
--- a/src/Ion.ts
+++ b/src/Ion.ts
@@ -58,15 +58,15 @@ export type ReaderBuffer = ReaderOctetBuffer | string;
  * @param buf       The Ion data to be used by the reader. Typically a string, UTF-8 encoded buffer (text), or raw
  *                  binary buffer.
  */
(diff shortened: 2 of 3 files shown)


desaikd (commented):
Thank you for the contribution @everett1992!
The change looks good, I have just added a minor comment.

desaikd (approved):

PR #774, opened 2024-03-28
Adds `browserslist` integration for browserify grunt task

desaikd:
### Description of changes:
This PR fixes the issue of `TypeError: Cannot convert a BigInt value to a number` when importing browser bundle of `ion-js`.

**Issue:** This is a known issue with babel: https://github.com/babel/babel/issues/13109. (this issues also has a suggested solution)

The fix lies in defining `browserslist` in `package.json` for `@babel/preset-env`. From the babel documentation:
> By default @babel/preset-env will use browserslist config sources unless either the `targets` or `ignoreBrowserslistConfig` options are set.

**Fix:** Following the solution specified in https://github.com/babel/babel/issues/13109, added `browserslist` in `package.json`. The list of compatible browsers is gathered from AWS SDK [supported browsers](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/setting-up.html#browsers-supported).

----

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -107,7 +107,7 @@ module.exports = function(grunt) {
          },
         transform: [["babelify", 
                      { 
-                       "presets": ["@babel/preset-env"], // TODO this targets ES2015, we should be more specific
+                       "presets": ["@babel/preset-env"], // Follows the browserslist from package.json
                        "plugins" : [["@babel/transform-runtime"],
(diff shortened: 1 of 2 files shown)


desaikd:
`ion-test-driver` seems to fail for its first task of `pip install` and doesn't seem to be related to this change.

zslayton (approved):

